PR #3, opened 2015-07-24
Creating inistial travis.phplint.sh

Reception123:


.travis.phplint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+#Author Reception123
```
Southparkfan: You're not the author.

.travis.phplint.sh
```diff
@@ -0,0 +1,12 @@
...
+
+find ../ -type f -regex '.*\.php\|.*\.php\.j2' -exec php -l {} \; | grep "Errors parsing ";
```
Southparkfan: Replace: ../ -type f -regex '.*\.php\|.*\.php\.j2'
With: ./ -type f -name '*.php'

PR #6, opened 2015-07-25
creating initial LocalExtensions

Reception123:
presuming we will have one.

### DIFF
--- a/LocalExtensions.php
+++ b/LocalExtensions.php
@@ -1,3 +1,6 @@
 <?php
 // Switch case
-switch( $wgDBname ) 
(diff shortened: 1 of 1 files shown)


LocalExtensions.php
```diff
@@ -0,0 +1,3 @@
+<?php
+// Switch case
+switch( $wgDBname )
```
NDKilla: Needs an opening { and } after the switch? Possibly add case meta wiki; just so you know how it's set up. (Look at the switch on Orain)

Reception123:
per SPF

PR #9, opened 2015-07-25
Revoke some perms from bureaucrat

NDKilla:


Reception123:
Thank you for fixing my mistakes.

LocalSettings.php
```diff
@@ -157,6 +159,10 @@
                 'suppressionlog' => true,
                 'suppressrevision' => true,
             ),
+            'bureaucrat' => array(
```
Southparkfan: In alphabetic order please :) so above oversight

PR #10, opened 2015-07-26
add wikicreator right

Reception123:


LocalSettings.php
```diff
@@ -187,6 +187,10 @@
                 'noratelimit' => true,
                 'userrights' => true,
                 'userrights-interwiki' => true,
+             'wikicreator' => array(
```
Southparkfan: The wikicreator array is not closed

LocalSettings.php
```diff
@@ -187,6 +187,10 @@
...
+               	'createwiki' => true,
+               	'autopatroll' => true,
```
Southparkfan: Should be autopatrol

JohnFLewis: Should just be createwiki.

Reception123:
Duplicated by #9

PR #31, opened 2015-08-05
add cratprotect for testwiki&admin rights add for default

Reception123:


Reception123:
@Southparkfan is this good now?

LocalSettings.php
```diff
@@ -257,6 +257,12 @@
 				'bot',
 				'sysop',
 				'bureaucrat',
+			'sysop' => array(
+				'autopatrolled',
+				'confirmed',
+				'rollbacker',
+			
+		),
```
NDKilla: This needs to be before the sysop, the two arrays look jumbled

LocalSettings.php
```diff
@@ -307,6 +313,11 @@
 			),
 			'wikicreator' => array(
 				'createwiki' => true,
+		'+testwiki' => array(
+			'bureaucrat' => array(
+				'bureaucratprotect' => true,
```
NDKilla: This is set up right I think but this right isn't definged or used anywhere

Southparkfan:
What do you want? Restricting edit/move/upload/create to bureaucrats?

NDKilla:
Probably all of the above like I have on my orain wiki and like reception tried to do before. Delete needs to be added as a restriction type and then bureaucrat needs to be given several rights (probbly 'bureaucrat' and 'protect' at the least)

Reception123:
Okay. To make things clear. I want to have bureaucrat protected pages. And the other one is so that administrators can grant rollback,patroll and confirmed

Reception123:
@NDKilla @Southparkfan I have fixed some stuff. Is it fine now?

Reception123:
I will leave it for later.

PR #39, opened 2015-08-08
add sysop groups rollback, confirmed and autopatroll for default

Reception123:


Reception123:
@Southparkfan

Southparkfan:
Can't merge this because a) PHP syntax error b) groups like autopatrolled/confirmed don't exist (yet).

PR #77, opened 2015-08-28
add skipcaptcha to autopatrolled and confirmed

Wynter:


Reception123:
I am not sure this is need as we do have the "skicaptcha" right

Reception123:
I am not sure this is needed as we do have the "skicaptcha" right
@Southparkfan What do you think?

NDKilla:
I don't see a reason not to give it to confirmed (at the least) and just remove the skipcaptcha right.

Reception123:
I perfectly agree with what @NDKilla said. But we can't have both. Either add the skipcaptcha right to confirmed and delete it, or not add it at all.

PR #78, opened 2015-08-28
enable GlobalCssJs globally

NDKilla:


NDKilla:
Close for now bc Idk how this works :D

PR #106, opened 2015-09-16
Add diavwiki license

revi:
Per IRC request.


Reception123:
actually was a on-wiki request

PR #107, opened 2015-09-16
RightsPage diavwiki+quantixwiki

NDKilla:
I didn't remove diavwiki RightsUrl as it might be nice to have the full path. RightsPage overrides RightsUrl


revi:
Note: this is followup of #106.

PR #134, opened 2015-09-22
cineama wiki changes

NDKilla:
requested on irc 

(might have to revoke some permissions from user, question wasn't answered)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -528,6 +528,14 @@ $wgConf->settings = array(
 				'rollbacker',
 			),
 		),
+		'+cinemawiki' => array(
+			'bureaucrat' => array(
+				'editor',
+			),
+			'sysop' => array(
+				'editor',
+			),
+		),
 		'+dpwiki' => array(
 			'bureaucrat' => array(
 				'respected',
@@ -620,6 +628,13 @@ $wgConf->settings = array(
 			),
 			'user' => array(
 				'user' => true, // for "Allow logged in users" protection level
+			),
+		),
+		'+cinemawiki' => array(
+			'editor' => array(
+				'edit' => true,
+				'move' => true,
+				'protect' => true,
 			),
 		),
 		'+dpwiki' => array(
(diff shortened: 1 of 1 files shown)


Reception123:
Don't forget to add MU and VE

revi:
>This branch has conflicts that must be resolved

You want to resolve conflicts by me or you will resolve them?

TheTVFan:
I would like someone to resolve the conflicts...

NDKilla:
@revi would love it if you could fix :)
btw added you to https://github.com/NDKilla/mw-config/ collab

TheTVFan:
$wgGroupPermissions['sysop']['bigdelete'] = false  - it should be true ;) can you fix?

NDKilla:
@thetvfan that should definitely not be true. (In this PR). That is an already in effect global setting not specific to your wiki. Sysops may delete pages but big delete is for deleting pages with over 250 revisions. Big delete limits vary wiki-farm to farm but most places leave bigdelete for sysadmins because of limitations of resources. The db servers don't have much ram which is why this limit is set

NDKilla:
@thetvfan https://www.mediawiki.org/wiki/Manual:$wgDeleteRevisionsLimit

TheTVFan:
Ok @NDKilla, I thought it was an option to mass delete pages on the wiki. :)

JohnFLewis:
The limit is honestly a plucked number out of no where. it'll get adjusted as we see how loads cope with large deletions. 250 is probably way below the limit of the database server. It's just the case that allowing everyone to delete pages with large revisions can lag databases and worst case crash them.

revi:
Will do, when I have time.

NDKilla:
@revi group permissions false doesn't work inside of wg conf (something special was done on Orain and not sure how that even works). 

This can be tested by ctrl+f'ing the file to see the one wiki with a group permission set to false, then going to "Special:ListGroupRights" on that wiki.

PR #146, opened 2015-09-24
enable VE and MU for cinemawiki per req

Reception123:


Reception123:
parsoid config was already added previously

PR #164, opened 2015-09-28
Add SyntaxHighlight to ATT

Arcane21:


NDKilla:
I will do this after I enable the extension and rebuild the LC in ~an hour. 

you should know that this can't be merged as is because the extension is disabled on extloadwiki

NDKilla:
Done in [commit 8f6ae37](https://github.com/miraheze/mw-config/commit/8f6ae371bb2c0d9337a8533d7c2044dec8a41bd4)

PR #168, opened 2015-09-30
Trope Workshop namespace config info

Arcane21:
Need this added for proper function of these namespaces.


JohnFLewis:
Please integrate this with existing config.

Arcane21:
We kinda need this, otherwise our Trope Workshop is broken, how do fix this?

revi:
You just need to see other wikis' example and follow them.

Specifically: define NS as is, name/alias at the respective config place below them. See NS_SERVER for reviwiki as example (Though no alias)

PR #184, opened 2015-10-16
add Variables (for LocalSettings.php) per req

Reception123:


Southparkfan:
Note: https://github.com/miraheze/mw-config/pull/185 depends on this.

PR #185, opened 2015-10-16
add Variables to szk wiki per req

Reception123:
Waiting for #184 to be merged.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -393,8 +393,10 @@ $wgConf->settings = array(
 		'torejorgwiki' => true,
 		'unikumwiki' => true,
 	),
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -393,6 +393,10 @@
 		'torejorgwiki' => true,
 		'unikumwiki' => true,
 	),
+	'wmgVariables' => array(
```
Southparkfan: Should be wmgUseVariables. Also, please add extloadwiki and allthetropeswiki too (and don't forget to remove https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L1259 :)).

PR #186, opened 2015-10-16
allow external images to szk wiki per req

Reception123:


revi:
I'd not merge it, since it might allow some malicious user putting some problematic images and let the victims get bad effects. Leaving final decision to SPF.

revi:
The expected downsides:
 - remote server owner will be able to gather private informations of visitors.
 - if remote file is deleted afterwards, it becomes deadlink.

I cannot think of any other downside for now.

Reception123:
Okay. I did not know that it could cause security issues. Let's wait for SPF's decision.

Southparkfan:
No, I don't like this.
Technically, it was possible to load external images (even without this setting): https://www.mediawiki.org/wiki/Manual:$wgEnableImageWhitelist - I disabled this behaviour in 27e2a065ee1323012e1dad309dcc4eb8333fe383.

If a founder can assure us that they will take full responsibility for everything if a website owner complains about web server load/bandwidth usage, then I'm fine with allowing external images (but, I prefer the image whitelist/$wgAllowExternalImagesFrom in that case. Don't just allow all sites!).

PR #197, opened 2015-10-21
change logo on hshsinfoportalwiki per req

Reception123:


revi:
-1
Logo file is not 135px*135px.

Past experience tells me that if it is not exactly 135px*135px it ALWAYS breaks.

PR #203, opened 2015-10-27
Correct permissions for walthamstowlabourwiki.

martinjos:
Removed redundant items, added 'createaccount' => false, and moved to wgGroupPermissions.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -876,6 +876,12 @@ $wgConf->settings = array(
 				'testgroup' => true,
 			),
 		),
+		'+walthamstowlabourwiki' => array(
+			'*' => array(
+				'edit' => false,
+				'createaccount' => false,
+			),
+		),
 	),
 	'wgGroupsRemoveFromSelf' => array(
 		'default' => array(),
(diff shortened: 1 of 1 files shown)


revi:
I do not think `createaccount => false` is appropriate. Accounts will be created via CA anyway, AFAIK.

martinjos:
This is valid (unless there is something Miraheze-specific at work here). See https://www.mediawiki.org/wiki/Manual:Preventing_access#Restrict_account_creation.
We would prefer to have control over account creation. See https://meta.miraheze.org/wiki/User_talk:Martinjs, under "...broad requirements".
Thanks.

JohnFLewis:
The general patch is sane and correct but revoking CreateAccount would cause issues with the central account login (and as in the past elsewhere, does create some annoying situations when automating things server side). As such, I'm going to block the removal of createaccount for technical reasons since I see no gain from preventing new accounts if they can't edit anyway.

Obviously my understanding is the original patch was incorrect so if you remove the CreateAccount part, we'd merge it. Thanks.

martinjos:
You misunderstand my intentions. New accounts _will_ be able to edit, as the __user__ role still has __edit__ (removing __edit__ from __*__ does not remove it from __user__ - see https://www.mediawiki.org/wiki/Manual:$wgGroupPermissions#Details).

In light of what you are saying, however, perhaps it would be better if I create a new group called __editor__ with __edit__ permissions, and remove __edit__ from __user__.

JohnFLewis:
Yes, that would be better in the long run.

martinjos:
I've made the changes - please review.

revi:
Merged, cherry-picked 117c619988c3c6c7fd9fb4577d03752e1577f8a7 and 43fca0c140b0ac671da15df55402b6f1268a3c41 with the first redundant commit omitted.

PR #204, opened 2015-10-29
add logo and scribunto to wdbwiki per req

Reception123:


Reception123:
@Southparkfan @revi @NDKilla  Still not merged.

PR #206, opened 2015-10-30
Extensions for walthamstowlabourwiki

martinjos:
Hi, I'd like to add these extensions to walthamstowlabourwiki:
AdminLinks
CharInsert
DynamicPageList
Flow (all talk namespaces)
Popups
Scribunto
VisualEditor
Variables

And also this config option:
wmgDisableAnonEditing
(The wiki still seems to be world-editable at the moment, and it looks like setting this might help.)

By the way, I think you should just be able to cherry-pick c63cfc1. Thanks!


revi:
Please submit a patch for parsoid (required for flow/ve) referencing miraheze/puppet#29 to the puppet repository.

revi:
Ah, ok. It has legacy stuffs from old PRs.

martinjos:
Sorry about that, I keep forgetting to use branches.

PR #222, opened 2015-12-21
add 'exampleuser' permission on testwiki

Reception123:
I want to test it out to see if it works, if yes it will be good so that the Example user's password cannot be changed on testwiki.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -777,6 +777,7 @@ $wgConf->settings = array(
 				'bot',
 				'bureaucrat',
 				'consul',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -947,7 +948,10 @@
 				'read' => true,
 				'testgroup' => true,
 			),
-		),
+			'exampleuser' => array(
+				'editmyoptions' => false,
```
Southparkfan: I'm not sure if editmyoptions also disables Special:ChangePassword?

Issue #224, opened 2015-12-22
Hook for not being able to change password for specific user


Reception123:
People at mediawiki told me that the hook below should work, only I'm not sure where it should be insterted : 
````php
$wgHooks['PrefsPasswordAudit'][] = 'test_onPrefsPasswordAudit';
function test_onPrefsPasswordAudit( $user, $newPass, $error ) {
    if ( $user->getName() == 'TestUser' ) {
        return "User not allowed to change password";
    }
    return true;
}
````

**Comment by Southparkfan** [comment]
Wrap it in an "if ( $wgDBname === 'testwiki' )" block at the bottom of LocalSettings.php, so:

```php
if ( $wgDBname === 'testwiki' ) {
	$wgHooks['PrefsPasswordAudit'][] = 'test_onPrefsPasswordAudit';
	function test_onPrefsPasswordAudit( $user, $newPass, $error ) {
		if ( $user->getName() == 'TestUser' ) {
			return "User not allowed to change password";
		}
		return true;
	}
}
```

Reception123:
Done in #225

Reception123:
Waiting for PR to be merged.

Reception123:
Has been reverted as hook did not work properly.

Southparkfan:
It did work.

PR #225, opened 2015-12-23
testwiki hook blocking password change for Example

Reception123:
To block changing password for Example


Reception123:
closes #224

PR #236, opened 2015-12-25
Update LocalSettings.php

D66Ha:
*added mirahezestaff Group
*added wikifounder Group
*Fixed and Adjusted The Supervisor Group
*Added 4 Protection Levels


Reception123:
Miraheze staff is not needed as the Steward permissions is enough.

PR #245, opened 2015-12-27
Revert wgReadOnly and sitenotice (when 1.26 upgrade is done)

Reception123:
DO NOT MERGE EXCEPT IF YOU ARE SURE THAT THE 1.26 UPGRADE IS DONE!!


Reception123:
Not needed anymore

PR #257, opened 2015-12-29
enable VectorBeta per req

Reception123:
Blocked #255


Reception123:
Conflicts. Will redo PR

PR #261, opened 2015-12-30
add foreground DO NOT MERGE!

Reception123:
I am not sure if this is supposed to be a global skin. DO NOT MERGE WITHOUT THE CONSENT OF SOUTHPARKFAN!


Reception123:
Per IRC moving to LS.php

PR #268, opened 2016-01-02
add TemplateData and add adnovumwiki to it

Reception123:


Southparkfan:
If wmgUseVisualEditor is set to true, TemplateData is automatically loaded.

PR #271, opened 2016-01-02
disable mobilefrontend for izangiwiki

Reception123:
Someone gave me this, i'm not 100% sure it works. @Southparkfan Do you think it would work?


Southparkfan:
It won't work. Please use a wmgUseMobileFrontend var.

Reception123:
I tried as you can see above but I don't think it's right.

NDKilla:
done in #275

PR #284, opened 2016-01-13
Enable visualeditor on boulderwikiwiki

robacarp:


robacarp:
Companion to https://github.com/miraheze/puppet/pull/53

PR #323, opened 2016-01-27
add notitle variable

Reception123:


LocalExtensions.php
```diff
@@ -85,6 +85,10 @@
 	require_once( "$IP/skins/Monaco/monaco.php" );
 }
 
+if ( $wmgUseNoTitle ) {
+	require_once( "$IP/skins/NoTitle/NoTitle.php" );
```
Southparkfan: skins?

LocalExtensions.php
```diff
@@ -48,7 +48,7 @@
 }
 
 if ( $wmgUseForeground) {
-	require_once( "$IP/skins/foreground/foreground.php" );
+	require_once( "$IP/extensions/foreground/foreground.php" );
```
Southparkfan: foreground is not an extension!

PR #335, opened 2016-01-28
add scribunto and disable anon edits for izangiwiki

Reception123:


NDKilla:
It's worth noting that this is Izanagiwiki not izangiwiki (2 A's)

PR #345, opened 2016-02-01
enable deletedhistory for sysop on metawiki

Reception123:
to fix translate problems


JohnFLewis:
right is already assigned.

PR #352, opened 2016-02-05
whitelist quantixgaming custom main page

NDKilla:


Southparkfan:
```
if ( $wgDBname === 'quantixwiki' ) {
    $wgWhitelistRead[] = "Quantix Gaming";
}
```
makes more sense.

NDKilla:
@Southparkfan I was kind of wondering how to do that without overwriting the whole array :) Thanks. will fix this later

JohnFLewis:
Seems like you'll make a new PR really with a single commit so closing. @NDKilla do as you wish!

PR #360, opened 2016-02-10
add VE to hobbieswiki per req

Reception123:
BLOCKED BY miraheze/puppet/issues/82


Reception123:
@NDKilla @Southparkfan

PR #361, opened 2016-02-10
add logo to christipediawiki per req

Reception123:


Reception123:
@NDKilla @Southparkfan

PR #402, opened 2016-03-02
lclwiki settings and order

Reception123:


JohnFLewis:
You've just redone the wrong commit. The url is https://lclwiki.miraheze.org/wiki/ therefore the database name is lclwikiwiki

Reception123:
Re-done in #403

PR #407, opened 2016-03-07
newcolumbia wiki update

samuelnajdowski:
Done.


labster:
Squished these commits into fa12d0e.  Thanks!

PR #412, opened 2016-03-10
Disable anonymous editing for fieldresearchwiki

D66Ha:
Per request at "Request features": Disable anonymous editing for "fieldresearch.miraheze.org"(fieldresearchwiki)


JohnFLewis:
Not done in favour of 14347ef63eee8571752cb8111b6bcf330d24e366 which is the (nicer) best way of doing it.

PR #446, opened 2016-04-04
add scribunto to whufcyouthwiki per req

Reception123:


D66Ha:
Medium Urgency - breaking templates

Reception123:
Techincally it's not breaking them as they never worked...

PR #476, opened 2016-05-05
add RandomSelection

Reception123:
add RandomSelection here


Southparkfan:
Do not merge before 'wmgUseRandomSelection' has been set in wgConf.

PR #478, opened 2016-05-06
add RandomSelection to humorpediawiki per req

Reception123:


Reception123:
Not 100% sure I did Subpages properly.." Feature 1: Enable subpages on all namespaces"

PR #485, opened 2016-05-16
ontariobrasswiki settings

Reception123:


LocalSettings.php
```diff
@@ -2099,6 +2104,7 @@
 		'carvingwiki' => 'America/Denver',
 		'catboxwiki' => 'America/Detroit',
 		'doraemonpediawiki' => 'Asia/Taipei',
+		'ontariobrasswiki' => 'New York/Toronto',
```
revi: It should be `America/Toronto` imo but you should double check it.

LocalSettings.php
```diff
@@ -1959,6 +1961,7 @@
 	'wgDefaultSkin' => array(
 		'default' => 'vector',
 		'cybercrimewiki' => 'modern',
+		'ontariobrasswiki' = 'monobook',
```
revi: Missing `>`

PR #492, opened 2016-05-20
Add ExtImg & TitleCap settings to jubeat Wiki @ TH

jiravatt:
- Allow jubeatwikithwiki to have whitelist for external image
- jubeatwikithwiki doesn't want their page's titles to have capitalized first letter so I've set $wgCapitalLinks = false only for this wiki and = true for all other wiki

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -968,6 +968,7 @@ $wgConf->settings = array(
 	),
 	'wgEnableImageWhitelist' => array(
 		'default' => false,
+		'jubeatwikithwiki' => true,
 	),
 
 	// Flow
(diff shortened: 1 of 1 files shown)


NDKilla-public:
@jiravatt
Syntax is right. Capital links is fine, but I think there's reasons we don't allow images (non HTTPS content etc) is there a reason you can't upload the images you want to your wiki?
Leaving this for more discussion

jiravatt:
Thank you for your replies, @NDKilla-public ! As a number of images I want to make a display on my wiki is about 600 and they're just a small .gif files. They're cover jacket for music that appeared in a game (FYI: Every pictures in this page http://p.eagate.573.jp/game/jubeat/qubell/p/information/music_list1.html). What I want to do in my wiki is to display some details about these musics with their jacket cover for the convenient of readers (like a picture that I've attached with this comment.) So if it violates the site's regulations, it's okay for you to ignore this request. Thank you again for your response in a short time :)
music_table

revi:
It's unfortunate that we are enforcing HTTPS always and everywhere, we can't allow plain http request. Also, it could mean that the visitor's IP may be collected by KONAMI (well, I'm a long player of jubeat, since the knit.) which is also unacceptable for us. Thank you for your contribution, though!

PR #493, opened 2016-05-22
Fix broken tmewiki aliases

stranger195:
By converting them to multi-byte


labster:
Sorry, that's not going to work for us.  Don't assume the encoding of the file you're using -- use escape sequences in strings.

PR #497, opened 2016-05-24
studynotekrwiki group permission changes per req

Reception123:
https://phabricator.miraheze.org/T290


Reception123:
@Southparkfan @revi @JohnFLewis

LocalSettings.php
```diff
@@ -2024,7 +2031,11 @@
 			'supervisor',
 			'wikifounder',
 		),
+		'+studynotekrwiki' => array(
+			'voter',
```
JohnFLewis: Is user right right not exist.

Reception123:
What do you mean? (The user right is created in $WgGroupPermissions)

JohnFLewis:
I can't see it. I see a voter user *group* but I don't see a user *right*. See examples for the user protection level, the user group has the right 'user' which is used to define who has access.

Reception123:
Add 'voter'

PR #498, opened 2016-05-24
add SocialProfile to microsoftwiki per req

Reception123:
// Requires copying of two directories: https://www.mediawiki.org/wiki/Extension:SocialProfile#Directories
	// Requires run of update.php!!


Reception123:
@Southparkfan @revi @JohnFLewis

JohnFLewis:
Needs copying of directory but not uodate.php.

JohnFLewis:
Still not handled? bah.

JohnFLewis:
Handled by @labster already it seems.

PR #504, opened 2016-05-31
Chage right in wiki idtest

AYST201:
Add right


labster:
This looks good to me.  Does someone else want to double-check?

JohnFLewis:
Syntax is good. No issues with any restricted right.

PR #507, opened 2016-06-01
add idwiki test

AYST201:
Add

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2186,7 +2186,7 @@ $wgConf->settings = array(
 		),
 		'+idtestwiki' => array(
 			'trusted',
-			'Bureaucrat',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2184,6 +2184,11 @@
 			'supervisor',
 			'wikifounder',
 		),
+		'+idtestwiki' => array(
+			'trusted',
+			'Bureaucrat',
```
labster: I don't think this will work with 'Bureaucrat'.  This should be 'bureaucrat'.

revi: Indeed.

AYST201:
Bah why@revi you comment in the diff? How if can't to merge :-)

PR #509, opened 2016-06-03
some edit

AYST201:
Some

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1853,6 +1853,7 @@ $wgConf->settings = array(
 				'nuke' => false,
 				'movefile' => false,
 				'blockemail' => false,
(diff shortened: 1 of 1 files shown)


NDKilla:
For the record, setting group permissions false inside of this array does not work. You must use [wgRevokePermissions](https://www.mediawiki.org/wiki/Manual:$wgRevokePermissions) or specify this outside of the wgConf (there are some examples near the end of the file.

For proof, look [here (Special:ListGroupRights)](https://idtest.miraheze.org/wiki/Istimewa:Daftar_hak_kelompok) where it says Administrators can still use blockemail even though you have the right set to false.

AYST201:
Yes I look, I will try .I'm sorry about my comment

PR #512, opened 2016-06-05
enable VE for cisowiki

AYST201:
Enable Visual Editor for Cisowiki


labster:
Added parsoid config in miraheze/puppet@70128058

PR #516, opened 2016-06-07
add Ilearnthingswiki for ve

AYST201:


JohnFLewis:
Database is missing an s from things.

PR #519, opened 2016-06-08
Add visual editor to Wikihoyo

atfornes:
This pull request has been done directly without opening a feature request. Should we open a feature request every time we want an extension to be enabled or is it OK to do a pull request like this one?

Thanks a lot.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -951,6 +951,7 @@ $wgConf->settings = array(
 		'webflowwiki' => true,
 		'wikicervanteswiki' => true,
 		'wisdomwikiwiki' => true,
+		'wikihoyo' => true,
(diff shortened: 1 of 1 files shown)


JohnFLewis:
It's fine to just open a pull request here.

VisualEditor requires parsoid however. If you'd like to do that change as well assuming no one beats you, https://github.com/miraheze/puppet/blob/master/modules/parsoid/manifests/init.pp#L26 is where to find the configuration! :)

If you or someone doesn't handle this by tomorrow afternoon I'll get around to it. Thanks.

atfornes:
Thanks @JohnFLewis, I did the pull request: https://github.com/miraheze/puppet/pull/154

PR #534, opened 2016-06-30
SeldirWiki disable anon edits + createaccount per req

Reception123:


JohnFLewis:
We do not to disable the createaccount right on wikis.

JohnFLewis:
Looking at the requests, you want to set user and editor to true to achieve this.

Reception123:
Ok. Will do.

PR #538, opened 2016-06-30
add InputBox to musiclibrarywiki per req

Reception123:


JohnFLewis:
First line is 'default' => true,

PR #548, opened 2016-07-04
add bitcoindebateswiki to DP exception per request on IRC

Reception123:
Requested on IRC by founder: 

9:22 AM <go1111111> the wiki 'bitcoindebates' which I maintain is being flagged for inactivity. the wiki is a roughly finished reference which continues to be useful in its current form. How can I keep it from being closed without having to make a bunch of no-op edits?

9:51 AM <•labster> go1111111: We can mark that wiki with an exception to the dormancy policy
10:03 AM <go1111111> labster: that would be awesome

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1367,6 +1367,7 @@ $wgConf->settings = array(
 		'default' => array(
 			'metawiki',
 			'allthetropeswiki',
+			'bitcoindebateswiki',
 			'spiralwiki',
 			'extloadwiki',
 			'loginwiki',


Southparkfan:
https://meta.miraheze.org/wiki/Dormancy_Policy#Exemptions_from_the_Dormancy_Policy not yet.

Reception123:
User confirms reading on IRC:
1:26 AM <go1111111> I was told to read https://meta.miraheze.org/wiki/Dormancy_Policy#Exemptions_from_the_Dormancy_Policy about getting an exception for the bitcoindebates wiki. I've read it and I think I qualify (it's a site with lots of info, meant to be read). does posting here count as 'showing some sign of activity globally'? 
2:19 AM <•PuppyKun> go1111111: You can talk on SN, or on IRC, basically anything that lets us know you didn't forget Miraheze existed
2:56 AM <go1111111> OK, thanks. so looks like that step is fulfilled. in that case could someone add an exception for the bitcoindebates site?

PR #549, opened 2016-07-04
remove ndtestwiki as wiki is deleted

Reception123:


JohnFLewis:
A lot of deleted wikis potentially fall into this as well.

PR #556, opened 2016-07-07
enable BetaFeatures for inazumaelevenwiki per req

Reception123:
https://phabricator.miraheze.org/T459


JohnFLewis:
BetaFeatures, MultimediaViewer and Popuos aren't enabled on the wiki.

PR #562, opened 2016-07-09
add more wisdomwikiwiki namespaces per req on IRC

Reception123:


AiraJun:
pls READ before pulling:

the NS 1630+31 collide with ATT NS 1630/31.

since id appreciate 4 consecutive numbers, i suggest we push them all up starting with 1654.
if you could add one more NS additionally, (call it NS blank for now) we should be covered with a series of consecutive numbers (in one row from 1654 upwards). 

this would leave the 1632/33 empty and to be reused in another wiki.

dont bother for possible data loss, i sign here with my blood that i wont hold you responsible, i have them all copied here, and its less than 5 mins to redo them in case they get lost.

thanks guys!

PR #564, opened 2016-07-10
activate extensions and subpages for NS

AiraJun:
activating 8 extenstions, 
enabling subpages for new namespaces


Reception123:
looks good :)

PR #570, opened 2016-07-12
configuration of wisdomsandboxwiki

AiraJun:
~ enabled FLOW and FORUM (needs parsoid enabled and run update.php manually)
~ more or less cloning most of the wisdomwiki-settings
~ all wisdomsandbox stuff is underneath wisdomwiki, for its a sub-wiki, and only appears when WW entry is present (seemed sensemaking). hope thats an acceptable exception of the alphanum-sortrule.
~ doublechecked the 'NS_TEST',1632 and TALK 1633 they are leftovers from an earlier shift and not in use as of now.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1572,10 +1572,11 @@ $wgConf->settings = array(
 			NS_TEACHING_TALK => 'Teaching_talk',
 			NS_BLANK => 'Blank',
 			NS_BLANK_TALK => 'Blank_talk',
-		)
+		),
 		'wisdomsandboxwiki' => array(
 			NS_TEST	=> 'TEST',
 			NS_TEST_TALK => 'TEST_talk',
+		),	
 	),
 	'wgContentNamespaces' => array(
 		'default' => array( NS_MAIN ),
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -477,6 +481,8 @@
 			'editor' => true,
 			'sysop' => true,
 		),
+		'+wisdomsandboxwiki' => array(
+			'anon' => true,)
```
NDKilla: the ) needs to be on a new line lined up with the start of line 484, and it needs a , after it

LocalSettings.php
```diff
@@ -85,6 +85,10 @@
 define('NS_BLANK', 1662);
 define('NS_BLANK_TALK', 1663);
 
+// wisdomsandboxwiki
+define( 'NS_TEST',1632);
+define( 'NS_TEST_TALK',1633);
```
NDKilla: Needs to be assigned/used in wgExtraNamespaces (near 1462)
https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L1462

PR #578, opened 2016-07-14
add protect restriction on testwiki

Reception123:
As most people have administrator we need an oversight which is only visible by bureaucrats/consuls. I'm not sure this is done correctly.


JohnFLewis:
Not looking actively but it is wrong. This allows you to set protection on the ability to delrev a page (which is impossible). It works as a single tool, you have it or you don't for a certain group unfortunately.

PR #580, opened 2016-07-15
change settings for thelonsdalebattalionwiki per req

Reception123:
Change Metrolook settings and read-only. T479


JohnFLewis:
You don't included th$.

PR #581, opened 2016-07-15
set scribunto active per default

AiraJun:
we talked about this in IRC, 
so far no objections. 

i sent SPF a pm in irc, no reply as of yet.

maybe someone wants to make a list of the wikis explicitly activated for preservation...

[list_of_wikis_that_had_scribunto_active.txt](https://github.com/miraheze/mw-config/files/366042/list_of_wikis_that_had_scribunto_active.txt)


JohnFLewis:
This is the wrong way to do it.

PR #592, opened 2016-07-16
Prettify wiki not found page

penguinstyles:


JohnFLewis:
it's ack'd and looks good to me but @Southparkfan made comments on IRC.

Handing to him.

Southparkfan:
Here are my comments:
- I do not like the importing of external resources (googleapis)
- The webdesign does not seem responsive
  - I cannot view the text without scrolling https://gyazo.com/1b21f3f2ecc3158cc4cefd54c208bea7
- I want a 'Miraheze theme'. This theme doesn't have any Vector, MonoBook or Miraheze colors styling. That doesn't mean this can't be a good theme, I'm just curious if we can re-use this (needs the opinion from other people).

labster:
I'm pretty sure this PR would violate our (not-quite-existing) privacy policy.  Public CDN resources are a public good, but they're also a public evil. *shrug*

If you really think that we need to change typography -- and I know this is a big thing with designers -- then I think I'd rather go about it by uploading the resource somewhere more static.  @Southparkfan do we have a good place to put static resources like that with long (⪆1 year expires) for CDN replacements?

penguinstyles:
I'll work on creating a responsive version of the page. What colours would you suggest to make it more "Miraheze-y"?

revi:
>I'm pretty sure this PR would violate our (not-quite-existing) privacy policy. Public CDN resources are a public good, but they're also a public evil. _shrug_

Actually that would mean Google ReCAPTCHA would be same in that regards. IIRC Google ReCAPTCHA NoCAPTCHA would send request to Google servers, too. 

But yeah, Google ReCAPTCHA is almost impossible to replace (there is almost no better CAPTCHA software) and we have a solution in this place (just store the googleapis.com materials on our side and use our server to display it.

Reception123:
This is probably the first and last PR that lastes/lasts 30 days :)

PR #631, opened 2016-07-26
fix alphabetical order in wgFavicon

Reception123:


penguinstyles:
You removed spacegamewiki but it does not look like you added it back.

PR #654, opened 2016-08-03
Add AccessControl for mindgearwiki

samtarling:
Add AccessControl for mindgearwiki

https://phabricator.miraheze.org/T589#8580 - confirmation of responsibility

Founder may also need to post a note on their own wiki: https://mindgear.miraheze.org - please check before merging


Reception123:
Sysadmins need to check if the confirmation is on the wiki as the wiki is private.

Reception123:
Confirmed: https://mindgear.miraheze.org/w/index.php?title=Main_Page&diff=541&oldid=522
Can be merged now

PR #684, opened 2016-08-20
add function for sitenotices on closed wikis

Reception123:


JohnFLewis:
Bad tabbing, you want a newline between the group rights and the hook calling and youve closed a comment that doesn't exists it's (*/)

PR #694, opened 2016-08-29
add wgFlowEditorList to espiralwiki and spiralwiki per req T693

Reception123:
Not sure if this is done correctly.


Reception123:
already defaulted. i got confused.

PR #705, opened 2016-09-11
enable VisualEditor for cnvwiki for T739

ImBoPhil:


Reception123:
VisualEditor also requires parsoid configuration so please also add cnv to https://github.com/miraheze/puppet/blob/master/modules/parsoid/manifests/init.pp (no wiki suffix just 'cnv')

PR #758, opened 2016-09-30
Enable CSS on applewikiwiki for T791

LegoFan4000:


LegoFan4000:
It's because there is no tab key on iOS.

Reception123:
@LegoFan4000 Fixed it.

PR #759, opened 2016-09-30
Add SocialProfile and WikiForum to allthetropeswiki for T761

LegoFan4000:


Reception123:
SocialProfile needs copying of directories, and since last time I didn't do it properly I am leaving this to someone else.

LegoFan4000:
As I have said before there is no tab key in iOS

NDKilla-public:
@LegoFan4000 Tabbing doesn't mean literally use the tab key, it means fix the indentation. In this case, if you compare the added lines to the rest of the lines, you actually have one to many space (on each line, before the first ' )

NDKilla-public:
@LegoFan4000 also it's not as important since someone else will probably eventually fix it, but these should be alphebatized

PR #761, opened 2016-10-01
Update trexwiki

TriX1:
Trexwiki


LocalSettings.php
```diff
@@ -3488,7 +3488,7 @@ function onPrefsPasswordAuditTestWiki( $user, $newPass, $error ) {
 }
 
 // WikiId Test overrides
```
NDKilla: please update this line as well

PR #762, opened 2016-10-02
update trex wiki

TriX1:


NDKilla:
Yay, conflicts! this is why you should put all of your changes into one PR :)

TriX1:
@NDKilla I'm sorry, because I'm not use desktop, I'm use mobile phone :)

PR #766, opened 2016-10-03
enable flow on moziwiki for T829

LegoFan4000:


ImBoPhil:
For Flow to you need to enable parsoid by editing the init.pp file which can be found here https://github.com/miraheze/puppet/blob/master/modules/parsoid/manifests/init.pp

ImBoPhil:
I've made a PR for parsoid too

PR #773, opened 2016-10-04
add WikiTextLoggedInOut for T761

ImBoPhil:


ImBoPhil:
this is what i'm doing now

PR #776, opened 2016-10-05
Enable VE on mylogicwiki for T842

LegoFan4000:


ImBoPhil:
Redone in #778 per tabbing issues

PR #779, opened 2016-10-05
add Refreshed skin for T835 and T590

ImBoPhil:


ImBoPhil:
doing LocalSettings.php now

PR #780, opened 2016-10-05
add Refreshed skin extension for T835 and T590

ImBoPhil:


ImBoPhil:
Conflicts fixed in #781

PR #781, opened 2016-10-05
add Refreshed skins for T835 and T590

ImBoPhil:


LegoFan4000:
Redone in #782 to compleat some requests for this skin.

PR #782, opened 2016-10-05
Enable refreshed on extloadwiki, applewikiwiki, and inazumaelevenwiki for T791 T590 and T835

LegoFan4000:


Reception123:
Conflicts (I have merged the other PR as it's always better to enable extload first to test it out before enabling it on all wikis)

PR #787, opened 2016-10-07
Enable Flow on marcoschriekwiki for T851

LegoFan4000:


Reception123:
parsoid has to be enabled as well

PR #788, opened 2016-10-08
add SQL files for MediaWikiChat for T849

ImBoPhil:
Sorry, I forgot to add the SQL files.


JohnFLewis:
The chat_away ones are patches and do not need to be added.

JohnFLewis:
Looks good.

For whoever merges it - these need to be ran on every wiki through sql.php

Reception123:
@JohnFLewis I'm not sure how exactly it is done..

PR #793, opened 2016-10-09
Enable VE and foreground on 3dicwiki for T859

LegoFan4000:


LegoFan4000:
I already made a pull request in the other repo and as I have said before there is no tab key in iOS.

ImBoPhil:
Re-done with the correct tabs in #794

Reception123:
Per comment above.

PR #795, opened 2016-10-09
disable anon and user editing (justwikiwiki) T858

ImBoPhil:


LegoFan4000:
@ImBoPhill this is done on meta by privlaged users. In this case @JohnFLewis did it.

Reception123:
@LegoFan4000 No, this is **not** done on Meta. This is the correct way to do it. JohnLewis just made the wiki public, he didn't actually make the wiki non-editable by users.

PR #796, opened 2016-10-09
Allow sysops to delete big pages on applewikiwiki for T791

LegoFan4000:


NDKilla:
The bigdelete limit is set because of technical limitations. I'm actually not sure that non-technical staff should ever have the ability to delete big pages. Instead, maybe (if the servers can handle it) that maybe the bigdelete limit should be raised on that wiki instead.

JohnFLewis:
If a page exceed this limit, then there is a chance it can cause unecessary load, memory exhaustion and so on. It's reserved an split off for a reason.

PR #797, opened 2016-10-09
Enable Google Analytics on amaninfowiki for T860

LegoFan4000:


Reception123:
google Analytics does not respect privacy therefore it has been decided that it will not be used on Miraheze (at least for the time being).

PR #799, opened 2016-10-09
Reenable MediaWikiChat on extloadwiki

LegoFan4000:


Reception123:
Please remove my message next to it in that case and also I will wait for @NDKilla as I'm not sure if the directories were successfully copied.

JohnFLewis:
@Reception123 I don't get what we're waiting for here. There are no directories for this extension.

Reception123:
There were the SQL files, but I guess @NDKilla did merge them so I'll go ahead and merge this as it's only extloadwiki in any case.

Reception123:
@JohnFLewis Apparently, the database changes weren't made per #809

LegoFan4000:
Maybe run Update.php

Reception123:
@LegoFan4000 We do not run update.php except while doing a MediaWiki update or in extreme cases.

JohnFLewis:
The daabase tables definitely exist. The issue is the LC cache wasn't rebuilt.

LegoFan4000:
How can we fix that.

Reception123:
I have already fixed it by changing the extension configuration back to require_once. It has been now been enabled.

PR #801, opened 2016-10-09
Add articlefeedbackv5

LegoFan4000:


Reception123:
This needs to be added to the Mediawiki repo, no use to just add it here if it's not actually installed. And the tabbing is wrong.

PR #811, opened 2016-10-10
add ArticleRatings for T848

ImBoPhil:

PR #812, opened 2016-10-10
Revert "add ArticleRatings for T848"

LegoFan4000:
Reverts miraheze/mw-config#811


Reception123:
There is no need for this as extloadwiki is used for testing and it is fine for it to be down for a while. If it is not done in >30 then of course I will revert but it can be left like this for now.

PR #813, opened 2016-10-10
fix error with ArticleRatings

LegoFan4000:


Reception123:
This is not the error! The error is the branch.

PR #818, opened 2016-10-11
Revert "add ArticleRatings for T848"

LegoFan4000:
Reverts miraheze/mw-config#811


Reception123:
@LegoFan4000 Please stop creating useless PRs for reverting all the files. I will disable it on extloadwiki now for the time being but there is no need to completely remove it.

PR #823, opened 2016-10-12
change Logo + Favicon (imperiuswiki) T874

ImBoPhil:


ImBoPhil:
fixed. As someone had previously added it in non-alphabetical order I did not notice it, will be more careful next time

Reception123:
Indeed, it should have been in alphabetical order.

NDKilla-public:
@ImBoPhil @Reception123 No problem, and kind of my bad. If you look back at the commit history it used to be in alphabetical order, because it used to be "blackmyth" or something. When the database was renamed, I replaced all instances of the old db name with the new one, without alphebatizing it. Any reference to this wiki is probably not abc'd, but feel free to fix it

PR #828, opened 2016-10-13
Add SQL files for ArticleRatings for T848

ImBoPhil:


ImBoPhil:
I assume they need to be ran on every wiki using sql.php again

JohnFLewis:
Yep, they do.

PR #832, opened 2016-10-14
Allow copy uploads on applewikiwiki for T791 and disable instant inst…

LegoFan4000:
…ant commons on pso2wiki for T883


Reception123:
As commented above by @ImBoPhil has been redone due to tabbing issues

PR #825, opened 2016-10-12
add ArticleFeedbackv5 for T847

ImBoPhil:

PR #839, opened 2016-10-14
add SQL files for ArticleFeedbackv5

ImBoPhil:
#825 im not sure if all of them are needed


Reception123:
Waiting for @JohnFLewis to reply on IRC as I'm not sure if all of them need to actually be added or not..

Reception123:
@ImBoPhil Per @JohnFLewis only one is needed which is the first one, ArticleFeedbackv5.sql .

PR #840, opened 2016-10-14
Enable comments on att for T761

LegoFan4000:


Reception123:
@JohnFLewis Doesn't the permission need to be added to wgGroupPermissions for it to work?

LegoFan4000:
@Reception123 It's a group created by the extension.

PR #841, opened 2016-10-15
Change to fablabesds

FelixINX:
Hi here is my change for my wiki.

- VisualEditor
- ContactPage
- Feedback

Also how can we change the default language of the wiki? I accidently choose frc instead of fr.

Thanks!

### DIFF
--- a/ContactPage.php
+++ b/ContactPage.php
@@ -42,3 +42,23 @@ if ( $wgDBname == 'christipediawiki' ) {
 		'RLStyleModules' => array(),
 	);
 }
+
+if ( $wgDBname == 'fablabesds' ) {
+	$wgContactConfig['default'] = array(
+		'RecipientUser' => 'Contact COOP FabLab ESDS',
+		'SenderEmail' => $wgPasswordSender,
+		'SenderName' => 'Ne pas répondre FabLab ESDS',
+		'RequireDetails' => true,
+		'IncludeIP' => false, // No privy
+		'AdditionalFields' => array(
+			'TitreArticleDemande' => array(
+				'label' => 'Titre de l\'article demandé (si c\'est la cas)',
+				'type' => 'text',
+				'required' => false,  // Either "true" or "false" as required
+			),
+		),
+		'DisplayFormat' => 'table',
+		'RLModules' => array(),
+		'RLStyleModules' => array(),
+	);
+}
--- a/ContactPage.php
+++ b/ContactPage.php
@@ -43,7 +43,7 @@ if ( $wgDBname == 'christipediawiki' ) {
 	);
 }
 
-if ( $wgDBname == 'fablabesds' ) {
+if ( $wgDBname == 'fablabesdswiki' ) {
 	$wgContactConfig['default'] = array(
 		'RecipientUser' => 'Contact COOP FabLab ESDS',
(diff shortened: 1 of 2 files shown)


LocalSettings.php
```diff
@@ -649,6 +649,7 @@
 	'wmgUseArticleFeedbackv5' => array(
 		'default' => false,
 		'applewikiwiki' => true,
+		'fablabesds' => true,
```
LegoFan4000: Wikiname + wiki

LocalSettings.php
```diff
@@ -740,6 +741,7 @@
 		'default' => false, // Add wiki config to ContactPage.php
 		'ayrshirewiki' => true,
 		'christipediawiki' => true,
+		'fablabesds' => true,
```
LegoFan4000: Wikiname + wiki

LocalSettings.php
```diff
@@ -1345,6 +1347,7 @@
 		'evawiki' => true,
 		'extloadwiki' => true,
 		'ezdmfwiki' => true,
+		'fablabesds' => true,
```
LegoFan4000: Wikiname + wiki

LocalSettings.php
FelixINX: @LegoFan4000 please elaborate. I am missing something?

FelixINX:
Thanks @LegoFan4000 . Any idea for the language?

LegoFan4000:
That is done on meta by an admin. As I am not an admin I cannot do that.

FelixINX:
Thanks I'll post an issue on phabricator

Reception123:
@LegoFan4000 It is not done by an admin but by a wiki creator or a system administrator. The administrator permission on Meta has nothing to do with global requests like this.
@FelixINX The rest is good now, except the "No reply user" which I'm not currently sure of.

ContactPage.php
```diff
@@ -42,3 +42,23 @@
 		'RLStyleModules' => array(),
 	);
 }
+
+if ( $wgDBname == 'fablabesdswiki' ) {
+	$wgContactConfig['default'] = array(
+		'RecipientUser' => 'Contact COOP FabLab ESDS',
+		'SenderEmail' => $wgPasswordSender,
+		'SenderName' => 'Ne pas répondre FabLab ESDS',
```
Reception123: I think this has to be "Miraheze No Reply" but I'm not sure.

JohnFLewis: It implies a 'Do Not Reply' therefore it is fine, it's only a conical name and has no impact.

FelixINX: @JohnFLewis Do I need To change it?

Reception123:
@FelixINX Leogfan forgot to mention that parsoid needs to be enabled for VE.

LegoFan4000:
Yes to do that, add ypur wiki[here](https://github.com/miraheze/puppet/blob/master/modules/parsoid/manifests/init.pp).

FelixINX:
Sorry about the delay, now done.

@Reception123 Should the `SenderName` be change?

Reception123:
@FelixINX As JohnLewis said, the SenderName is fine.

FelixINX:
Ok thanks.

PR #849, opened 2016-10-16
enable extensions + change logo and favicon (takethatwikiwiki) T909

ImBoPhil:
waiting for miraheze/puppet/pull/276


Reception123:
As SocialProfile is involved, I will need to copy the directories. Hopefully I do it right this time.

PR #850, opened 2016-10-17
Enable VE and change logo and favicon on karniarutheniawiki for T902

LegoFan4000:


ImBoPhil:
Redone in #851 because of syntax errors.

PR #855, opened 2016-10-18
enable VE (sidemwiki) T911

ImBoPhil:


ImBoPhil:
waiting for miraheze/puppet/pull/278

PR #857, opened 2016-10-21
Reenable search

LegoFan4000:


Reception123:
This shall be done at the discretion of an operation member, when we are sure that it can be enabled.

PR #865, opened 2016-10-23
Try to fix refreshed per a comment on phab

LegoFan4000:


labster:
OK so this is a chicken-and-egg problem.  extloadwiki is what we use to load all extensions, so that we can build the localization cache.  Because both Refreshed and MobileFrontend have system messages, they both need to be enabled on extloadwiki.  If this needs to be disabled for cache to load, we may have to 'trick' it on extloadwiki, like we did here in LocalExtensions.php:

```php
if ( $wmgUseHeaderTabs ) {
	require_once "$IP/extensions/HeaderTabs/HeaderTabs.php";
	if ( $wgDBname == 'extloadwiki' ) {
		// load LC despite SectionHide's strong objections
		unset( $htUseHistory );
	}
}
```
But remember, it doesn't have to work correctly on extload, it just has to run.

Anyway, closing this, as it is the wrong approach.

PR #882, opened 2016-10-28
Create PrivateWikiWhitelist

Reception123:


Reception123:
@Southparkfan @JohnFLewis It is not completed but it has most languages on. It should be right but I prefer if you merged it just in case.

JohnFLewis:
I don't see the file being included in LS

JohnFLewis:
Looks okay

PR #888, opened 2016-10-30
Add global notice

LegoFan4000:


Reception123:
There was no actual downtime where the sites were completely accessible so there's no point to make a sitenotice.

PR #890, opened 2016-11-01
Enable string functions on marcosschriekwiki for T812

LegoFan4000:


Reception123:
This will be on hold ATM, per @NDKilla 's comment on https://phabricator.miraheze.org/T812 .

Reception123:
I've checked mediawiki.org and this would be integrated into ParserFunction. Will still wait for another sysadmins approval before merging.

LegoFan4000:
@Reception123 that what this is for.

JohnFLewis:
Why is this adding an undefined wmg variable?

Reception123:
(I might be wrong but) I think its supposed to be defined in LocalExtensions.php and then used as a wmg.

Reception123:
It's not done like that. You would need to define it under the extension, so therefore this could actually go to GlobalExtensions (but again not sure this is the correct approach)

LocalSettings.php
```diff
@@ -3746,4 +3751,3 @@ function onSiteNoticeAfter( &$siteNotice, $skin ) {
 
 	return true;
 } */
-
```
Reception123: please add this line back

JohnFLewis:
Why are all these changes being made? wg in LocalSettings was correct as this is a global extension!

JohnFLewis:
Please make a new request. The amount of changes to this is getting way too much.

Reception123:
@LegoFan4000 Sorry, but I did tell you that I wasn't sure, you should have waited for a confirmation.

PR #896, opened 2016-11-03
Add FlaggedRevs for T966

ImBoPhil:


ImBoPhil:
Not sure which configuration is needed for which wiki, so for now I'll leave it like that

ImBoPhil:
Should I create a separate PR for the wiki configuration on the wiki or do it in this one? (as someone told me that it first only has to be enabled on extload)

PR #900, opened 2016-11-04
Disable URLshortner on espiralwiki for T974

LegoFan4000:


Reception123:
The issue mentioned says that it is not working and "If the fix is not trivial, you can safely remove this feature. It's not important for us." that does not necessarily mean that they want to remove it. It is better to try to fix the error first, if this is not possible please reopen this PR.

PR #902, opened 2016-11-05
settings for FlaggedRevs (pso2wiki) T970

ImBoPhil:
Not entirely sure if everything is correctly done


Reception123:
It looks correct to me but I don't want to merge before confirmation from another sysadmin. The numbers might have to be inside **'** but again I am not sure about this.

Reception123:
@ImBoPhil Please note that it is not necessary for the DB name to have a + if they're all false.

PR #909, opened 2016-11-09
Change toftcricketclubwiki logo

davidblackburn:


Reception123:
@LegoFan4000 There's no need to add comments on standared PRs. If it is more complicated configuration it is ok, but generally there is no need to review PRs.

PR #913, opened 2016-11-11
add LocalisationUpdate for T923

ImBoPhil:


ImBoPhil:
I made a pull request did it for LocalExtensions first so it can be ran on extloadwiki, but then should I move it to GlobalExtensions.php?

Reception123:
Not sure, but for now this can be left like that.

PR #916, opened 2016-11-11
custom domain settings (wikia22wiki)

ImBoPhil:
Wating for https://github.com/miraheze/puppet/pull/296 to be merged first and I can't find the task on Phabricator.


LegoFan4000:
Before this merged the cert should be fixed as it is not valid for that domain name. The one it has now is for .miraheze.org

Reception123:
@LegoFan4000 What do you mean by the cert being "fixed"? I have generated a cert for the domain name, and it needs to be merged by an Operations member, then this PR can be merged as well. If by "fixed" you meant merged there's no need to comment things that are already known.
@ImBoPhil The task is on Phabricator, but is limited to members of the Security group (Staff members).

PR #926, opened 2016-11-12
Change refreshed to wfload

LegoFan4000:


Reception123:
Will check this out later.

Reception123:
Will merge but if anything goes wrong I will have to revert.

Reception123:
@LegoFan4000 As you can see above, it didn't work.

labster:
Of course it's not going to work if you don't put quote marks around the name of the skin.

labster:
Actually never mind that last comment, I forget how terrible a language PHP is.

PR #929, opened 2016-11-12
implement LocalisationUpdate T923

ImBoPhil:
I'm unsure about this, there could be some mistakes, or it should have been done in GlobalExtensions.php, so I'm not sure if this is correct.


JohnFLewis:
The thing that needs to be done is the update.php script in the extension. Otherwise this does nothing.

PR #938, opened 2016-11-14
Add namespaces for flaggedrevs for T1014

LegoFan4000:


JohnFLewis:
This should be added for wikibridge only.

JohnFLewis:
Missing + in the dbname

PR #947, opened 2016-11-15
Extensions on porpwiki for T1019

LegoFan4000:


LegoFan4000:
VE stills needs parsoid

PR #948, opened 2016-11-15
FlaggedRevs setting on wikibridgewiki for T1018

LegoFan4000:


JohnFLewis:
wg doesn't work for local extensions that don't merge properly.

LegoFan4000:
What should I do then?

JohnFLewis:
the same that has been done to all the other variables.

JohnFLewis:
Now you just need to define the variable

PR #951, opened 2016-11-16
disable WikiForum (allthetropeswiki) T1025

ImBoPhil:


ImBoPhil:
Not sure if this doesn't get rid of the database tables, as @Arcane21 said "so simply disable the extension and keep the database tables,"

PR #953, opened 2016-11-16
Enable MultimediaViewer on bttestwiki for T1024

LegoFan4000:


LegoFan4000:
Was don't before #952

ImBoPhil:
I actually did it 4 minutes ago

PR #955, opened 2016-11-17
change logo (eternalconflictwiki) T1032

ImBoPhil:


LegoFan4000:
I have uploaded a new version of the file that is already set which makes this PR pointless.

PR #964, opened 2016-11-18
move InputBox to GlobalExtensions

ImBoPhil:


JohnFLewis:
Why have you change it from the recommended loading method to the old method?

LegoFan4000:
Redone in #965

PR #970, opened 2016-11-19
change possible extensions to wfLoad

ImBoPhil:
Checked for extension.json for all the extensions changed

PR #974, opened 2016-11-19
change possible extensions to wfLoad

ImBoPhil:
Redo per comment at #970

### DIFF
--- a/LocalExtensions.php
+++ b/LocalExtensions.php
@@ -57,9 +57,8 @@ if ( $wmgUseCategoryTree ) {
 }
 
 if ( $wmgUseCharInsert ) {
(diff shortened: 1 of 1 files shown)


LocalExtensions.php
```diff
@@ -354,7 +353,7 @@
 }
 
 if ( $wmgUseShortURL ) {
-	require_once( "$IP/extensions/UrlShortener/UrlShortener.php" );
+	wfLoadExtension ( 'ShortURL' );
```
Southparkfan: Should be UrlShortener

Southparkfan:
I prefer we don't do more than 5 extensions each commit. Can you only change the require_once -> wfLoadExtension for 5 extensions now?

PR #981, opened 2016-11-20
user rights and settings on wikijokewiki and wikiarchitecturewiki for T1045

LegoFan4000:


JohnFLewis:
wmgFlaggedRevsAutoReview should probably be set to true on default really.

PR #985, opened 2016-11-21
Remove redundant managewiki rights

LegoFan4000:


JohnFLewis:
Are they redundant?

LegoFan4000:
Since we are now using managewiki, the managewiki page has disappeared.

JohnFLewis:
But it's the same infra. The managewiki[-restricted] rights are still in use.

PR #992, opened 2016-11-22
remove Variables enable VE (wikipucwiki) T1051

ImBoPhil:
parsoid already done


LegoFan4000:
Does not matter which or is merged they were both done at the same time

PR #993, opened 2016-11-22
Variables -> VE on wikipucwiki per T1051

LegoFan4000:


LegoFan4000:
Does not matter which pr is merged, they were opened at the same time.

PR #1013, opened 2016-11-26
enable MsUpload + assign founder permissions (wikicanadawiki) T1064

ImBoPhil:


Reception123:
@JohnFLewis @NDKilla Founder permission needs to be assigned to user.

JohnFLewis:
The group has no permissions.

Reception123:
@ImBoPhil Please add a 'read' permission for the founder group, so that it will work properly.

PR #1017, opened 2016-11-26
User Rights on wikicanadawiki for T1064

LegoFan4000:


Reception123:
As seen on line 3261, wgRevokePermissions doesn't work, therefore you need to do this in the overrides.

LegoFan4000:
Done although it seems to work on login wiki and testwiki

Reception123:
@LegoFan4000 As per what @JohnFLewis said in a previous commit, you need to add a "read" permission for the "banned" group as groups can't be assigned without permissions.

LegoFan4000:
Adding the read permission would defeat the purpose of this PR as we are revoking the read permission

Reception123:
@LegoFan4000 Yes, I did not realize that. Please assign another arbitrary permission then.

JohnFLewis:
Repealing read is not the same as revoking read.

This first commit was correct to me.

JohnFLewis:
Though you don't want to revoke read from *, you want to repeal that.

LocalSettings.php
```diff
@@ -4105,11 +4113,6 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
 	$wgGroupPermissions['sysop']['abusefilter-modify-restricted'] = false;
 }
 
-if ( $wgDBname == 'wikicanadawiki' ) {
-	$wgGroupPermissions['*']['read'] = false;
```
JohnFLewis: This was correct.

LocalSettings.php
```diff
@@ -3265,6 +3265,14 @@
 				'editmyoptions' => true,
 			),
 		),
+		'wikicanadawiki' => array(
+ 			'*' => array(
+ 				'read' => true,
+ 			),
+ 			'banned' => array(
+ 				'read' => true,
```
JohnFLewis: This is correct.

JohnFLewis:
Though I dont have shell right now so i prefer @Reception123 merging right now.

PR #1029, opened 2016-11-29
Revert "Read access for non-logged in users on wikicanadawiki per T1078" and whitelist some pages for T1078

LegoFan4000:


LocalSettings.php
```diff
@@ -4166,6 +4166,13 @@ function wfModifyMetaTags( OutputPage $out ) {
 	require_once( "$IP/extensions/OpenGraphMeta/OpenGraphMeta.php" );
 }
 
+if ( $wgDBname == 'wikicanadawiki' ) {		
+	$wgGroupPermissions['*']['read'] = false;	
+	$wgWhitelistRead = array(
+    "Main Page", "Special:UserLogin", "Special:UserLogout", "Special:PasswordReset", "MediaWiki:Common.css", "MediaWiki:Common.js"
```
Reception123: Why is Special:PasswordReset necessary if it's an anon account?

LegoFan4000:
It applies to the banned group as well

LocalSettings.php
JohnFLewis: If you need to reset your password, your not logged in.

PR #1035, opened 2016-11-30
Fix my mistake from earlier

LegoFan4000:


LocalSettings.php
```diff
@@ -3290,6 +3291,7 @@
 				'autopatrolled',
 				'rollback',
 				'bot',
+				'sysop',
```
NDKilla: the entire crat section is uneeded i think since crats already have this (and this array is just adding to existing permissions...)

LegoFan4000:
Moderator, autopatrolled, confirmed, bot are requested by DeltaQuad and somehow crats cannot add sysops.

Reception123:
@LegoFan4000 That makes no sense. The default MediaWiki configuration is that crats can add sysops, except if done in the overrides there is no way that crats can't add sysops

LegoFan4000:
According to Special:Version on wikicanada they can't

Reception123:
You mean https://wikicanada.miraheze.org/wiki/Special:ListGroupRights . I will check it out, but it shouldn't be happening, as I said by default crats should be able to add sysops on all wikis.

Reception123:
Issue was caused by the fact that the rollbacker permission was added as "rollback"

PR #1042, opened 2016-12-02
Disable SMW version of maps

LegoFan4000:
Per instructions in https://github.com/JeroenDeDauw/Maps/issues/286


JohnFLewis:
ive already acknowledged this change.

PR #1054, opened 2016-12-07
Update LocalSettings.php

1-2-1-2-3-Miraheze:


Reception123:
There is already a PR for this ( #1024 ) but the extension itself isn't installed. As this seems to be a priority for you I will go ahead and install it now.

PR #1057, opened 2016-12-07
Update LocalSettings.php

1-2-1-2-3-Miraheze:

PR #1058, opened 2016-12-07
Userrights on wikicanadawiki

LegoFan4000:
Redoing #1057

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2717,9 +2717,10 @@ $wgConf->settings = array(
 			),
 		),
 		'+wikicanadawiki' => array(
-			'Founder' => array(
+			'founder' => array(
 				'bureaucrat',
 				'banned',
+				'founder',
 			),
 			'moderator' => array(
 				'autopatrolled',
@@ -3320,9 +3321,10 @@ $wgConf->settings = array(
 			),
 		),
 		'+wikicanadawiki' => array(
-			'Founder' => array(
+			'founder' => array(
 				'bureaucrat',
 				'banned',
+				'founder',
 			),
 			'moderator' => array(
 				'autopatrolled',


Reception123:
@LegoFan4000 You forgot sysop

LegoFan4000:
@Reception123 crats can already add sysops by default so it's not necessary.

Reception123:
Whoops. Anyways I can't merge until I am able to login and change the current founder to be able to merge this.

1-2-1-2-3-Miraheze:
What do you mean by "can't log in" @Reception123 ?

1-2-1-2-3-Miraheze:
And why do you have to de-founder my sister? Can't you just change the group name to lower case and add the ability to remove the group, without doing anything else? Shouldn't the software automatically update the abilities of my sister?

LegoFan4000:
No it does not do that automaticly. It has to be done manualy by a steward.

LegoFan4000:
He cant log in because of our current server situation.

Reception123:
Yes, due to our server (db2) having issues, we had to migrate to another one and therefore there are some login issues. The removal of your sister as a founder is not permanent, it is only to change the name, after that she will be added back. Yes, unfortunately there is no way of doing this "automatically".

1-2-1-2-3-Miraheze:
@Reception123 the simple solution I use is checking the "Keep me logged in" box on every Miraheze wiki that I log in to.

Reception123:
@1-2-1-2-3-Miraheze I'd be happy if that were the solution. Unfortunately, it still doesn't work for me. Sorry for the inconvenience and I hope that I can get back to your request tomorrow, when the database issue will hopefully be resolved.

LegoFan4000:
@NDKilla can you do it? I know that you were able to login

NDKilla:
@1-2-1-2-3-Miraheze No. Even though renaming the group in the code implies it's the same group with the same permissions, the software thinks it is a brand new group. After the rename, your sister would still have the old group name listed on her account and in the database, and it would be unremovable because the group doesn't exist.

Also, repeatedly creating unconstructive pull requests will result in your ability to contribute to the configuration being restricted. Our actual configuration that affects servers is not the place to discuss the global removal of CheckUser, which, based on my comment on that PR, won't ever happen.

PR #1060, opened 2016-12-07
Update LocalSettings.php

1-2-1-2-3-Miraheze:
A few other things on request of DQ

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3174,6 +3174,7 @@ $wgConf->settings = array(
 			),
 			'sysop' => array(
 				'protectsite' => true
+			),	
 		
 		),
 		'+wikipucwiki' => array(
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3170,6 +3172,9 @@
 				'autopatrolled' => true,
 				'supressredirect' => true,
 			),
+			'sysop' => array(
+				'protectsite' => true
+
```
LegoFan4000: this array needs to be closed i.e. put ), in  the empty space

1-2-1-2-3-Miraheze: Done

labster:
I have to say, I'm opposed to this being merged.  We have someone in this commit who is not a bureaucrat of that wiki assigning user rights to themselves.  If we keep letting commits like this get through, it's really going to undermine the trust our users have in us.

PR #1061, opened 2016-12-08
Update GlobalExtensions.php

1-2-1-2-3-Miraheze:
I've been informed that CU is hardly even used, so why do we even need it?


NDKilla:
Really? Because it's hardly used **but** it is still used. If we did this then only system administrators would have access to the information (and it's much more annoying to access).

PR #1067, opened 2016-12-10
Disable CheckUser on WikiCanada

1-2-1-2-3-Miraheze:
If I can't use it, just get rid of it so it's not an issue


NDKilla:
Sorry but that's not exactly how this works. Also, that variable doesn't exist... also, syntax errors :D

PR #1077, opened 2016-12-11
Assign interwiki permissions on all wikis

LegoFan4000:
Not sure why it wasn't done before


Reception123:
@LegoFan4000 Most things are done for a reason :) When users get the Interwiki permission it allows them to edit the interwiki for every wiki on Miraheze (including Meta) and we don't want that.

PR #1078, opened 2016-12-11
add wiki.drones4nature.info to URLShortner whitelist

LegoFan4000:


JohnFLewis:
That list is the least standardised one ever.

PR #1081, opened 2016-12-12
Restrict all actions related to tags on WikiCanada

AmandaQuad:
All actions related to change tags are now revoked from the user group and restricted to sysops and a new "tag manager" group

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2755,6 +2755,7 @@ $wgConf->settings = array(
 			),
 			'sysop' => array(
 				'moderator',
+				'tag manager'
 			),
 			'bureaucrat' => array(
 				'moderator',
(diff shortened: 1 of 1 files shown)


JohnFLewis:
Syntax errors and this is not how revokepermissions works. This would revoke the permission from all users who are registered.

AmandaQuad:
@JohnFLewis well, the _least_ you could do is re-form my request to include the proper syntax, etc.

LocalSettings.php
```diff
@@ -3193,8 +3194,12 @@
...
 			),
 			'sysop' => array(
-				'protectsite' => true
-			),	
+				'protectsite' => true,
+			),
+			'tag manager' => array(
+				'changetags' => true,
+				'applychangetags' => true,
+				'managechangetags' => true,
```
JohnFLewis: The issue is this isn't closed.

JohnFLewis:
I've commented the relevant issue.

AmandaQuad:
I don't know how to fix the syntax error without creating a whole new PR. In all regards, the description I provided is exactly what I want to do (I might of made an error or two in the edit itself)

JohnFLewis:
You can push the change to https://github.com/AmandaQuad/mw-config/tree/patch-10 and it should appear here. You should then be able to re-open the request. If not - just comment and I'll merge it.

LegoFan4000:
You have to open a new one be cause John closed it and you can't reopen it unless you were the one that closed it.

LegoFan4000:
Also, user group names should have no spaces.

PR #1085, opened 2016-12-12
Update LocalSettings.php

AmandaQuad:
Would this be technically/safely possible to enable? If not, why is it even listed?

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2785,8 +2785,8 @@ $wgConf->settings = array(
 				'skipcaptcha' => true,
 			),
 			'bureaucrat' => array(
-				'renameuser' => false,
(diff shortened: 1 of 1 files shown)


LegoFan4000:
No it would not. You would then be able to use CU and OS so please close this using the button next to the comment button. Renaming is also done globally.

LegoFan4000:
In other words, this PR is declined.

JohnFLewis:
Renames are handled globally by stewards.

Group additions must be defined in wgAddGroups, userrights is not permitted as it allows assigning all user group including restricted ones.

LegoFan4000:
Its listed because, by default MW assigns those rights to crats, therefore they must be revoked.

PR #1086, opened 2016-12-12
rm sitenotice

LegoFan4000:
Has been up for a while now


JohnFLewis:
Remove the block rather than comment it out please. We don't need two if them here.

LegoFan4000:
Like that?

JohnFLewis:
I said remove it not change how it's commented out

PR #1090, opened 2016-12-14
calexitwiki settings

calnation:


labster:
Looks good.  (Though I'm still wondering how y'all are going to deal with the USN base in San Diego when we vote leave.)

calnation:
For better or worse, that pattern wouldn't be [anything new to the US](https://en.wikipedia.org/wiki/List_of_United_States_military_bases#Overseas_3).

PR #1094, opened 2016-12-15
Allow only founder to change local rights on WikiCanada

AmandaQuad:


NDKilla:
Probably wont work as expected because you don't revoke the wgAddGroups or whatever from bureaucrat, so they wont be able to add default groups but bureaucrats still have default rights

AmandaQuad:
@NDKilla Again, I'm still learning. I may have done it wrong, but I don't know how to fix it. The exact thing that I want to do is disallow sysops, bureaucrats, and moderators from accessing the Special:UserRights page altogether. The only person who should be able to access that page is the founder (me/DQ). I should be able to edit all rights except Steward, CU, and OS.

AmandaQuad:
@JohnFLewis @NDKilla @LegoFan4000 @Reception123 can one of you please re-do/correct this if indeed I did make a mistake?

NDKilla:
Done in commits the following commits
- 1911b2819a114156584bc7c96645bfa8dbb6b1f5
- e2f392de4a574d027f5772c846325e0be4f89645
- 301f91326989c2d8f8ea34094382816bf863b8bf

AmandaQuad:
Still missing tagmanager and rollbacker

PR #1096, opened 2016-12-15
User group permissions on WikiCanada

AmandaQuad:
Remove moderator group - merge former moderator permissions into founder


AmandaQuad:
Scratch that

PR #1105, opened 2016-12-20
Requesting global imput on status update

AmandaQuad:


JohnFLewis:
Unfortunately this is not an appropriate use of global notices. We use them rather for discussions that globally impact every single user or to communicate status of technical issues/outages.

PR #1109, opened 2016-12-24
Database issue

DeltaQuad-DQ:
MySQL seems to have failed - all wikis and Phabricator cannot connect at the moment


labster:
A site notice will never display when wikis aren't rendering, so this PR will never be useful to a single user.  Declined.

labster:
As an update, I tried to add a message to the error page, but that didn't work either.  I opened https://phabricator.wikimedia.org/T154076 as a follow-up on this issue.

PR #1110, opened 2016-12-24
allrightsreserved licsense for soundboxwiki per req T1179

Reception123:
Not sure if we accept a license like this so I need another sysadmins approval before merging this


JohnFLewis:
This is already used on another wiki

PR #1115, opened 2016-12-26
Simplify sitenotice

DeltaQuad-DQ:
Put notice into more simple language for users who may not have a clue what MariaDB is


LulzKiller:
I think this just removes detail rather than simplifying it.

Reception123:
I agree with LulzKiller. The issue was a few days ago so I think it's safe to remove the sitenotice now anyways.

PR #1120, opened 2016-12-27
Add read-only settings

DeltaQuad-DQ:


Reception123:
We don't use a "wg-ReadOnlywikis" variable. If you would like this on your on wiki please make a different PR

PR #1122, opened 2016-12-27
Add founder permissions on Elements

DeltaQuad-DQ:


LocalSettings.php
```diff
@@ -3270,6 +3270,11 @@
 		'+elementswiki' => array(
 			'founder' => array(
 				'read' => true,
+				'globalblock-whitelist' => true,
+				'interwiki' => true,
```
Reception123: Interwiki is not given to userrights on wikis as it edits Miraheze's global interwiki.

LocalSettings.php
```diff
@@ -3270,6 +3270,11 @@
...
+				'editusercss' => true,
+				'edituserjs' => true,
+				'siteadmin' => true,
```
Reception123: Not sure what this permission is/does.

JohnFLewis: Locking/unlocking a database.

DeltaQuad-DQ:
Sorry, interwiki should have been editinterface. The siteadmin permission is a direct way of doing indefinite read-only, rather than having to go through a new PR every time. I don't expect to need to use it frequently, but it is one of those things that it would be handy to have around, just in case.

Reception123:
@DeltaQuad-DQ From what I know you can't have siteadmin, but I am not completely sure of that. If you want this PR merged now please remove it and you may make another one for siteadmin and another sysadmin will say if you can have it or not.

DeltaQuad-DQ:
I have made a Phabricator task regarding the siteadmin permission

PR #1129, opened 2016-12-27
Bureaucrats can now manage a few user groups

DeltaQuad-DQ:


DeltaQuad-DQ:
@Reception123 now crats can add/remove more groups than I assigned in LocalSettings.php

DeltaQuad-DQ:
Nvm - reverting

PR #1133, opened 2016-12-27
Does this work?

DeltaQuad-DQ:
Just for my knowledge, does this code work to have a local wiki opt-out of global extensions?

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -86,6 +86,7 @@ if ( $wgDBname == 'elementswiki' ) {
 	$wgGroupPermissions['sysop']['importupload'] = false;
 	$wgGroupPermissions['sysop']['import'] = false;
 	$wgGroupPermissions['sysop']['unblockself'] = false;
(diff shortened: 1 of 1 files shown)


DeltaQuad-DQ:
I'm going to assume that the answer is no, since the default check has failed

JohnFLewis:
Indeed, nope.

DeltaQuad-DQ:
Is there a possible way to do this @JohnFLewis

JohnFLewis:
Nope.

DeltaQuad-DQ:
So local wikis can opt out of global sitenotices and global defaults of permissions, but they can't opt out of global extensions?

JohnFLewis:
Yes because those are handled in different ways. Inclusion of global extensions is loading everything in a file and then including it.

DeltaQuad-DQ:
And I gather it's not technically possible to unload everything out of a local file and then exclude it from specified wikis

JohnFLewis:
Nope

PR #1134, opened 2016-12-27
Enable Flow on Elements wiki for T1207

DeltaQuad-DQ:
I believe that there is more to do than this, but this is all I know how to do

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -964,6 +964,7 @@ $wgConf->settings = array(
 		'dicficwiki' => true,
 		'drones4allwiki' => true,
 		'dtswiki' => true,
+		'elementswiki' => true,
 		'ernaehrungsrathhwiki' => true,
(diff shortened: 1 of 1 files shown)


DeltaQuad-DQ:
Hello? Is this being ignored? @Reception123 @JohnFLewis

JohnFLewis:
It's not.

Needs [parsoid](https://github.com/miraheze/puppet/blob/master/modules/parsoid/manifests/init.pp) config however to work.

DeltaQuad-DQ:
@JohnFLewis I don't know how to do that. As noted above, this request contains the settings that I know how to configure. Someone else will need to take care of the rest.

JohnFLewis:
Okay, I've done it at miraheze/puppet@3433bead595ba9f5fbe68d19d428bdb58144df07.

PR #1135, opened 2016-12-27
Remove suppress permission for flow

DeltaQuad-DQ:


JohnFLewis:
Any particular reason?

DeltaQuad-DQ:
Deletion - which bureaucrats will hold - should be enough.

DeltaQuad-DQ:
Besides, I don't want to purposely be adding user rights to groups that I will never have access to

JohnFLewis:
Also an interesting scenario!

You didn't add the permissions to the oversight group - the extension did. By removing them, you're adding code so you're not really removing anything but instead adding :wink:

DeltaQuad-DQ:
@JohnFLewis The extension not only added the suppression right to the oversight group - it also created a brand new "Suppressor" user group that has this right and nothing else. It isn't really necessary to have the right in 2 different groups - neither of which I have access to - so what code would overwrite and remove the new extra group? Also, what is the purpose of the "Flow Bot" user group?

DeltaQuad-DQ:
@Reception123 Can you look at this if @JohnFLewis doesn't get to it?

Reception123:
@DeltaQuad-DQ I would but I don't think removing the default groups created by the extension is possible, nor would I know how to do it if you could.

DeltaQuad-DQ:
@JohnFLewis per above, do you know if removing the "Suppressor" user group is possible, since the sole permission that it has is already included in the oversight group?

JohnFLewis:
You simply remove all permissions from the group.

DeltaQuad-DQ:
But the group would still technically exist, right? Or could it be renamed to "null group" or something?

DeltaQuad-DQ:
@JohnFLewis this failed entirely. The extra group still exists on the wiki.

The-Voidwalker:
You might want to use:
`unset( $wgGroupPermissions['suppress'] );`

If it still exists, you can set the following to hide the group from menus:
`$wgImplicitGroups[] = 'suppress';`

FYI, [`$wgFlowGroupPermissions`](https://github.com/miraheze/mw-config/blob/master/LocalWiki.php#L97) doesn't exist.

DeltaQuad-DQ:
@The-Voidwalker do you have any idea how to fix https://phabricator.miraheze.org/T1211

PR #1137, opened 2016-12-28
Merge CU and OS rights into Steward group for Elements

DeltaQuad-DQ:
Reducing the total number of user groups on the wiki


DeltaQuad-DQ:
The "oversight" and "check user" groups will then be overwritten locally

Reception123:
This would not be possible per the policy. Stewards have to assign the rights, they don't get them by default.

PR #1139, opened 2016-12-28
Not exactly sure what this right is for

DeltaQuad-DQ:
But it doesn't have access to any restricted interfaces, so....

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2864,6 +2864,7 @@ $wgConf->settings = array(
 				'autopatrolled',
 				'rollbacker',
 				'researcher',
+				'Flow bot',
 			),
 		),
 	),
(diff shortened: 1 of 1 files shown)


Reception123:
I would merge but I'm not 100% sure if spaces are allowed for permissions. Please wait for someone else to double check.

DeltaQuad-DQ:
The user group as defined by the extension, as defined in Special:UserRights and as defined in Special:ListGroupRights has a space. Also note that the "check user" group has a space without issues.

Reception123:
@DeltaQuad-DQ I checked the file and there it seems to be possible so I will merge.

JohnFLewis:
Though in future, please use lowercase.

DeltaQuad-DQ:
@JohnFLewis The group is defined as uppercase in the extension

JohnFLewis:
But we prefer lowercase. It's consistency among us.

DeltaQuad-DQ:
Err... that didn't work. That created a brand-new "Flow bot" user group and didn't add the permissions to the existing predefined one like I wanted. If you go to Special:ListGroupRights on elements.miraheze.org you will see what I am talking about.

JohnFLewis:
https://github.com/wikimedia/mediawiki-extensions-Flow/blob/bc94b5d247990d4214d08e7daf54333b5894df55/Flow.php#L201

The group is called "flow-bot" which matches our standards of lowercase and no spaces.

DeltaQuad-DQ:
Oops. Will fix now.

PR #1146, opened 2016-12-28
Override again!

DeltaQuad-DQ:


Reception123:
@DeltaQuad-DQ I know that what you mean is good but please try to not create so many PRs in one day.

DeltaQuad-DQ:
That should be the last one regarding elements wiki... at least until DQ returns.

DeltaQuad-DQ:
You know... I kind of wish that GitHub configuration settings could allow administrators to "blacklist" certain additions/deletions/modifications of code from being made. Any changes that do not trigger the blacklist would be automatically merged, and PR's would only be required for changes that do trigger the blacklist. I know that's not technically possible, but it would be nice, and would definitely reduce the amount of PR's per day - for all users.

PR #1154, opened 2016-12-30
De facto autopatrolled for Elements Wiki founder

DeltaQuad-DQ:


labster:
I don't know why this has to exist.  Can't you just assign rights to yourself on-wiki?

Reception123:
Oops. Sorry labster for some reason your comment didn't load until after I merged.

DeltaQuad-DQ:
@labster technically I can, but I like to remain only in the founder group except when I need other permissions. If you look at the user rights log on my wiki, you'll see that I'm treating advanced permissions like CU and OS - only granting when needed.

PR #1160, opened 2017-01-01
restrict perms per T1237

The-Voidwalker:


Reception123:
@The-Voidwalker If possible please include the wiki name in the PR title too.

PR #1161, opened 2017-01-01
Abuse filter settings on Elements wiki per T1234

DeltaQuad-DQ:


LocalSettings.php
```diff
@@ -136,6 +136,30 @@
 		'default' => false,
 		'metawiki' => true,
 	),
+	'wgAbuseFilterAvailableActions' => array(
+		'default' => array(
+			 'flag' => true,
```
Reception123: please fix the tabbing here

DeltaQuad-DQ: @Reception123 what do you mean?

Reception123: 'flag' is tabbed wrongly it is ahead of the others. like
 'flag' instead of
'flag'

DeltaQuad-DQ: Fixed

PR #1163, opened 2017-01-02
IP's blocked by abuse filter shouldn't be indefinite

DeltaQuad-DQ:


DeltaQuad-DQ:
Failed check = syntax error

Syntax error = uh oh

DeltaQuad doesn't know where the error is = bigger uh oh

@Reception123

LocalSettings.php
```diff
@@ -160,6 +160,13 @@
 			'tag' => true,
 		),
 	),
+	'wgAbuseFilterBlockDuration' => array(
+		'default' => 'indefinite',
+	),
+	'wgAbuseFilterAnonBlockDuration' => array(
+		'default' => null
```
JohnFLewis: Here. You can simply it by just making the default "6 months" as we really shouldn't be indefinitely blocking IPs

DeltaQuad-DQ:
@JohnFLewis fixed

PR #1164, opened 2017-01-02
Unset CU group on Elements Wiki

DeltaQuad-DQ:
The use of CheckUser locally on Elements Wiki is prohibited per https://elements.miraheze.org/wiki/Elements:Privacy_policy#CheckUser barring exceptional circumstances. If one of those exceptional circumstances should arise, stewards can perform CU globally via Meta anyway.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -102,7 +102,7 @@ if ( $wgDBname == 'elementswiki' ) {
 	$wgGroupPermissions['*']['flow-hide'] = false;
 	unset( $wgGroupPermissions['suppress'] );
 	unset( $wgGroupPermissions['rollbacker'] );
-	unset( $wgGroupPermissions['checkuser'];
+	unset( $wgGroupPermissions['checkuser'] );
 	$wgGroupPermissions['sysop']['flow-delete'] = false;
 	$wgGroupPermissions['sysop']['flow-edit-post'] = false;
 	$wgGroupPermissions['sysop']['flow-lock'] = false;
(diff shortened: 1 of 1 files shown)


JohnFLewis:
Stewards can't perform global CheckUsers.

DeltaQuad-DQ:
@JohnFLewis really? I would think that because of CentralAuth, running a CU on Meta would reveal something like:

* 1 January 2000 JohnDoe123 used IP 123.456.789 on example.miraheze.org

JohnFLewis:
Nope. A CU on meta only reveals things that happen on meta. Plus loginwiki is the 'global' wiki in term of CentralAuth not meta but still only shows things that happen on loginwiki.

DeltaQuad-DQ:
@JohnFLewis So if merging this would technically prevent all CheckUsers on Elements Wiki, how else do I enforce the local privacy policy?

JohnFLewis:
The current method does comply with your local policy. For a steward to be able to run a CU on your wiki, there needs to be a log produced [here](https://meta.miraheze.org/wiki/Special:log/rights).

DeltaQuad-DQ:
Yeah, but because I don't have the "userrights" permission, if I notice a steward granting themselves CU rights on Elements Wiki, and there was no local request and I didn't authorize it, what do I do? I can't remove the rights, and unless my RFC on global rights passes, stewards could undo any sanctions that would prevent them from carrying out the CU.

JohnFLewis:
The odds are a steward will have carried out a CU before you have time to react unless you monitor the rights log 24/7.

If it directly contravenes with your local policies, then the only thing you can really do is make a case following [this policy section](https://meta.miraheze.org/wiki/Stewards#Revocation_of_rights).

DeltaQuad-DQ:
*sigh* I guess I'll just have to keep my fingers crossed that stewards read the local polices before taking any technical actions. I mean, links to them are in big text on the main page, so you really can't miss them unless you're being totally ignorant.

DeltaQuad-DQ:
Cause honestly, a no-confidence vote would get shot down for being opened just for violation of one local policy, although in theory, that should be enough to revoke rights.

PR #1167, opened 2017-01-04
Sysop-only editing on cosiadventurewiki per T1247

DeltaQuad-DQ:


ImBoPhil:
This is done here: https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L438

PR #1169, opened 2017-01-04
Revert editing restrictions on cosiadventure per T1255

DeltaQuad-DQ:


JohnFLewis:
I've looked at the request. sysop needs to be added, there's technically nothing wrong with this needing a revert.

ImBoPhil:
Fixed in https://github.com/miraheze/mw-config/commit/97c489635260e02cb70f02f2504ce194e1786386

PR #1171, opened 2017-01-05
Wiki creators are not technically Miraheze staff

DeltaQuad-DQ:


revi:
Wiki creators only create the wikis, they do not close the wiki. Sysadmin do. So it's still staff.

DeltaQuad-DQ:
@revi wiki creators also have the "managewiki" userright, so they can close wikis.

PR #1175, opened 2017-01-06
NDKilla said this was already defined...

DeltaQuad-DQ:
...but it seems it is not

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1469,7 +1469,7 @@ $wgConf->settings = array(
 		'extloadwiki' => true,
 		'qwertywiki' => true,
 	),
-	'wmgUseScributno' => array(
+	'wmgUseScribunto' => array(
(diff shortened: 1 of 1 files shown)


JohnFLewis:
This is enabled on every wiki by default already.

DeltaQuad-DQ:
@JohnFLewis then why is this not working? https://elements.miraheze.org/wiki/Template:Infobox_element

DeltaQuad-DQ:
@JohnFLewis how do I fix said error?

JohnFLewis:
Making an edit to the page will likely fix it

DeltaQuad-DQ:
@JohnFLewis that completely broke it! Look what happened to the page

DeltaQuad-DQ:
@JohnFLewis What has happened to the template? It's completely broken.

PR #1179, opened 2017-01-07
Enable VisualEditor on plasticssongcontestwiki per T1268

DeltaQuad-DQ:


DeltaQuad-DQ:
@JohnFLewis @Reception123 can this be merged now? I have a feeling that the affected user will start to get impatient.

PR #1182, opened 2017-01-08
Update ContactPage.php

DeltaQuad-DQ:


ContactPage.php
```diff
@@ -43,6 +43,20 @@
 	);
 }
 
+if ( $wgDBname == 'deltaquadtestwiki' ) {
+	$wgContactConfig['default'] = array(
+		'RecipientUser' => 'DeltaQuad Test Wiki',
+		'SenderEmail' => $wgPasswordSender,
+		'SenderName' => 'noreply@deltaquadtestwiki.org',
```
Reception123: might have to be 'Miraheze No Reply'. Also deltaquadtestwiki.org is not a registered domain name therefore it would look dubious for people to receive emails from noreply@miraheze.org but have the sender name be an email address from an inexistant domain name.

DeltaQuad-DQ:
@Reception123 changed

Reception123:
@DeltaQuad-DQ The Recipient User needs to be a user on the wiki that has a verified email.

PR #1183, opened 2017-01-08
enabling extensions and namespaces for calexitwiki

calnation:


Reception123:
@calnation Conflicts need to be fixed. I recommend you refork miraheze/mw-config or make a new patch from the current version as there's some weird spaces all around on this PR

PR #1191, opened 2017-01-09
Lock editing on howtoturtlewiki per T1277

DeltaQuad-DQ:


DeltaQuad-DQ:
@JohnFLewis @Reception123 can this be merged now? I have a feeling that the affected user will start to get impatient.

JohnFLewis:
Should be disabled by setting "anon => true" in $wmgEditingMatrix. https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L438

DeltaQuad-DQ:
Fixed in #1192

PR #1192, opened 2017-01-09
Fix #1191 (lock anon editing on howtoturtlewiki for T1277)

DeltaQuad-DQ:


LocalSettings.php
```diff
@@ -533,6 +533,9 @@
 			'user' => true,
 			'sysop' => true,
 		),
+		'howtoturtlewiki' => array(
```
JohnFLewis: needs a +

DeltaQuad-DQ:
Syntax fixed

PR #1206, opened 2017-01-16
SecurePoll on bigforestwikipediawiki per T1310

The-Voidwalker:


LocalSettings.php
```diff
@@ -1492,6 +1492,7 @@
 	),
 	'wmgUseSecurePoll' => array(
 		'default' => false,
+		'bigforestwikipedia' => true,
```
JohnFLewis: lacking wiki suffix

PR #1225, opened 2017-01-25
userrights on bigforestwikipediawiki per T1331

The-Voidwalker:


LocalWiki.php
```diff
@@ -68,6 +68,76 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
...
+	$wgGroupPermissions['judge']['delete'] = true;
+	$wgGroupPermissions['judge']['deleterevision'] = true;
+	$wgGroupPermissions['judge']['undelete'] = true;
+	$wgGroupPermissions['judge']['browsearchive'] = true;
+	$wgGroupPermissions['judge']['deletedtext'] = true;
+	$wgGroupPermissions['judge']['securepoll-create-poll'] = true;
+	$wgGroupPermissions['judge']['managechangetags'] = true;
+	$wgGroupPermissions['judge']['tboverride'] = true;
+	$wgGroupPermissions['judge']['abusefilter-view-private'] = true;
+	$wgGroupPermissions['judge']['abusefilter-log-private'] = true;
```
Reception123: This would reveal IPs so we can't have it. Please remove the three -private

PR #1234, opened 2017-02-05
Changing localsettings.php per T1392

BenjaminSchool:
This resolves the request of T1392.


Reception123:
@BenjaminSchool Unfortunately this would allow you to edit Special:Interwiki on all Miraheze wikis, so therefore for the meantime this feature is denied. If you have any things you want to add to Special:Interwiki that you think could be used globally please request them in your Phabricator task and a sysadmin will take a look.

PR #1235, opened 2017-02-05
userrights on bigforestwikipediawiki per T1331

The-Voidwalker:


NDKilla:
I didn't comment on the initial edit of user rights but is there a reason that wg(Add|Remove)Groups has a line per group for some groups and an array of groups for other groups?

The-Voidwalker:
@NDKilla That's because the one's where I use arrays are new usergroups, but when altering existing usergroups, I just added a new item into the array.

Reception123:
@NDKilla

PR #1237, opened 2017-02-08
grant editothersprofiles to sysops

The-Voidwalker:
~~In it's current configuration, the group can only be added by Stewards. It may be better to just `unset( $wgGroupPermissions['staff'] );` and assign `$wgGroupPermissions['sysop']['editothersprofiles'] = true;`, but this is the least (possibly) controversial of the two actions.~~
See T1397 for reference.

### DIFF
--- a/LocalExtensions.php
+++ b/LocalExtensions.php
@@ -404,6 +404,7 @@ if ( $wmgUseSiteScout ) {
 
 if ( $wmgUseSocialProfile ) {
 	require_once( "$IP/extensions/SocialProfile/SocialProfile.php" );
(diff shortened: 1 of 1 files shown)


The-Voidwalker:
Now, I'm just assigning the `editothersprofiles` to the sysop group.

Reception123:
Shouldn't this just be specific for att?

The-Voidwalker:
@Reception123 well, the thing is, I'm ~~pretty~~ sure all wikis using social profile have the same problem. Actually, I just checked Special:ListGroupRights for all the wikis currently using social profile, and all but two have this issue.

PR #1238, opened 2017-02-10
Usawiki: create special user group for Miraheze staff

AgraveBanks:


Reception123:
These permissions cannot be used by normal wikis as Miraheze is a wiki farm. Only stewards have access to these permissions.

AgraveBanks:
@Reception123 this "staff" group would be taking the place of the "steward" group locally on usawiki. I wouldn't have access to it - only users with the 'userrights-interwiki' permission (stewards and sysadmins) would be able to add themselves to the group.

AgraveBanks:
@Reception123 if you look at Special:ListGroupRights on usawiki, you'll see that the "steward" group has the CentralAuth permissions and that there are "checkuser" and "oversight" groups as well. All I'm doing is merging the three groups into one and adding the "renameuser" permission to it.

Reception123:
@AgraveBanks Ok. Please call the entire permission "steward", though.

AgraveBanks:
@Reception123 done. Please also merge the other PR that you closed, because that one unset the "oversight" and "checkuser" groups so that the permissions would not be in two places. In addition, I want to unset the default "steward" group and replace it with this one so that I can create a custom project page for "stewards". Currently, the "steward" link in Special:ListGroupRights soft-redirects to the Meta policy.

labster:
PR by a ban evader. Closing PR.

PR #1239, opened 2017-02-10
Follow up to #1238

AgraveBanks:


Reception123:
per last comment on PR

PR #1251, opened 2017-02-14
userrights configuration on Picturepedia per IRC

Trevor12Davis:


The-Voidwalker:
This would be invalid, since it would remove all of the previously defined userrights from the steward group.

Trevor12Davis:
@The-Voidwalker fixed

PR #1255, opened 2017-02-15
Enable VisualEditor for bchwiki

MrSimonC:
Rather than log an issue on phabricator - I've found this is what volunteers end up doing, so have simply added code to enable VisualEditor for bchwiki directly.


Reception123:
For VisualEditor to work you also need to add your wiki name (no 'wiki' ending) here https://github.com/miraheze/parsoid/blob/master/parsoid.yaml

PR #1290, opened 2017-03-05
Userrights on Americanopedia #1

Antonio12345678:
I'm not sure if I did this correctly as I am not familiar with the wiki farm configuration setup. I have the "allow maintainer edits" feature turned on, so please edit my commit and fix any errors that I have.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -61,6 +61,20 @@ if ( $wgDBname === 'allthetropeswiki' ) {
 	$wgNamespaceContentModels[NS_REVIEWS] = CONTENT_MODEL_FLOW_BOARD;
 }
 
+if ( $wgDBname ==== 'americanopedia' ) {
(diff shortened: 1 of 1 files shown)


Antonio12345678:
What does the "all checks have failed" error mean? Is that expected or not? It didn't happen with my other commit.

LocalWiki.php
```diff
@@ -61,6 +61,20 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
 	$wgNamespaceContentModels[NS_REVIEWS] = CONTENT_MODEL_FLOW_BOARD;
 }
 
+if ( $wgDBname ==== 'americanopedia' ) {
```
The-Voidwalker: you have an extra `=` there in the conditional

Reception123:
Sock of LP

PR #1291, opened 2017-03-05
Userrights on Americanopedia #2

Antonio12345678:
I'm not sure if I did this correctly as I am not familiar with the wiki farm configuration setup. I have the "allow maintainer edits" feature turned on, so please edit my commit and fix any errors that I have.


Reception123:
Sock of LP

PR #1293, opened 2017-03-07
Uncommented and added volunteers-needed sitenotice

calnation:


gfn1227:
I currently oppose as this has been hastily rushed by two users; and current volunteers have been going with the flow and tolerating it, but I will not. Current volunteers need to be able to figure things out before we can invite more volunteers into the mix--especially the ones who have elevated rights on all wikis and are able to do considerable amounts of damage in the wrong hands. I'm concerned about this, and I hope that is addressed before a global site-notice is put across all of our wikis.

labster:
Ugh, a link on `this page`?  What is this, 1997?  [Click here](https://www.youtube.com/watch?v=gvdf5n-zI14) to find out!

Southparkfan:
I'm against site notices that can't be closed with a single click, just to avoid complaints. I had this same issue with the many site notices regarding the database crashes, but these were critical.

The current text seems like it should be permanently showed (use of 'always', no 'currently' or so). A temporary site notice can be OK, a permanent one is very unlikely. Also I think we should work this out a little bit more first.

revi:
I think this should be done with [CentralNotice](https://github.com/wikimedia/mediawiki-extensions-CentralNotice), if possible.

(well, looking at the code it looks like it is centralized for Wikimedia use like UploadWizard. See also [T121](https://phabricator.miraheze.org/T121).

labster:
We have DismissableSiteNotice on for everyone right now, so it can be dismissed with a single click, as proposed above.  Not that I think we're ready for this.

Reception123:
No activity. When this is needed/wanted feel free to reopen the PR

PR #1326, opened 2017-03-14
Configurations for WikiHub

JPGumby:


JPGumby:
@revi I want my wiki to use QuestyCaptcha instead of the default - how else would I do this?

PR #1340, opened 2017-04-04
Update LocalSettings.php

bs-yoonho:
Added settings for permission at KSTARTUPSWIKI


Reception123:
@bs-yoonho This PR is approved, but is redundant without the "LocalWiki" PR as these settings are already like this by default so after you create that one this one can be merged too

PR #1343, opened 2017-04-04
Update LocalWiki.php

bs-yoonho:


bs-yoonho:
Should be merged with #1342 
Please consider asap.. ㅠㅠ @revi

PR #1356, opened 2017-04-16
Userrights on WikiCanada

Amanda1234567:


LocalSettings.php
```diff
@@ -4100,6 +4111,38 @@
...
+				'import' => true,
+				'markbotedits' => true,
+				'mergehistory' => true,
+				'abusefilter-modify' => true,
+				'abusefilter-modify-restricted' => true,
+				'changetags' => true,
+				'applychangetags' => true,
+				'editcontentmodel' => true,
+			),
+			'sysadmin' => array(
```
revi: These are handled by global groups.

Reception123: We've explained this several times. CheckUser and Oversight are "steward" permissions and not "sysadmin" permissions

Amanda1234567:
@Reception123 fixed

Reception123:
@Amanda1234567 As revi said, these are "global groups". Is it really necessary that they be unset to a single group? What difference does it make for you?

Amanda1234567:
The difference that it makes is that there is only one group that I cannot access, rather than three. I like to be able to edit the maximum number of user groups that I can, so by restricting all of the unauthorized interfaces to a single group, now there's only group that I cannot edit.

Reception123:
@Amanda1234567 "unauthorized interfaces" - I don't see why you're calling them "unauthorized" they are perfectly authorized and part of Miraheze's global policies.

revi:
Still, global groups bypasses local groups.

Amanda1234567:
"Unauthorized" as in "not authorized for local use except by stewards and sysadmins".

I'll remove the `globalgroupmembership` and the `globalgrouppermissions` userrights if you want, but I don't want to have a separate "check user" and "oversight" group.

Amanda1234567:
@Reception123 @revi If you don't think that this PR is acceptable, I'll close it and one or both of you can send a PR to https://github.com/WikiCanada/mw-config so that I know what you do consider "acceptable".

Amanda1234567:
@Reception123 @JohnFLewis updates?

Amanda1234567:
C'mon Travis-ci - pick up the pace

PR #1362, opened 2017-04-16
WikiCanada userrights #2

Amanda1234567:


JohnFLewis:
can you use -'s instead of spaces please for group names. It makes the config look a lot nicer.

Amanda1234567:
@JohnFLewis done

PR #1366, opened 2017-04-16
AbuseFilter and anti-spam config for WikiCanada

Amanda1234567:


LocalSettings.php
```diff
@@ -178,17 +178,31 @@
 			'rangeblock' => false,
 			'tag' => true,
 		),
+		'wikicanadawiki' => array(
+			'flag' => true,
+			'throttle' => true,
+			'warn' => true,
+			'disallow' = true,
```
The-Voidwalker: missing angle bracket '`>`'

Amanda1234567:
Fixed

LocalSettings.php
```diff
@@ -178,17 +178,31 @@
 			'rangeblock' => false,
 			'tag' => true,
 		),
+		'wikicanadawiki' => array(
```
Reception123: Can't you just use rangeblock => true, and remove the rest as they're already default values?

Amanda1234567:
@Reception123 done

PR #1369, opened 2017-04-17
move bigforestwiki to db3

Reception123:
second largest wiki, I think it should be move to db3

### DIFF
--- a/Database.php
+++ b/Database.php
@@ -4,6 +4,7 @@ $wgLBFactoryConf = array(
 	'class' => 'LBFactoryMulti',
 	'sectionsByDB' => array(
 		'allthetropeswiki' => 'c2',
+		'bigforestwiki' => 'c2',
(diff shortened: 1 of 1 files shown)


JohnFLewis:
Not the largest wiki on db2.

```
root@db2:/srv/mariadb# du -sh * | grep G
1.1G    bigforestwiki
1.5G    bpwiki
1.2G    mysql
1.2G    piwik
```

db2 has more space than db3 as well.

```
root@db2:/srv/mariadb# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/simfs      120G   88G   33G  73% /

root@db3:~# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/simfs      100G   72G   29G  72% /
```

Nothing to suggest bigforestwiki will grow at a rate worth moving it across servers.

PR #1370, opened 2017-04-17
WikiCanada userrights again

Amanda1234567:


Reception123:
There is no problem at all but if possible try to make less PRs for the same subject but of course if you really need to it is fine.

PR #1383, opened 2017-04-19
Removing all permissions from CheckUser group

Amanda1234567:
Since the group cannot be deleted and since checkuser rights have been assigned to the steward group, removing all permissions from the CheckUser group should invalidate it.


Reception123:
I guess this is the same as unsetting the permission

PR #1384, opened 2017-04-19
Define "banned" user group

Amanda1234567:


LocalSettings.php
```diff
@@ -4203,6 +4203,16 @@
 			'bot' => array(
 				'editinterface' => true,
 			),
+			'user' => array(
```
Reception123: Except chat these are all default user permissions, so again they don't need to be included

LocalSettings.php
```diff
@@ -4430,6 +4440,22 @@
 				'editmyoptions' => true,
 			),
 		),
+		'wikicanadawiki' => array(
```
Reception123: This is unneeded, only the LocalWiki settings are.

LocalSettings.php
Amanda1234567: @Reception123 by default these permissions are assigned to the '*' group, but I'm revoking them from the '*' group and assigning them to the 'user' group instead (because the '*' group would override the "banned" group)

Amanda1234567: @Reception123 for some reason my above comment is all screwed up. It is supposed to read like this:

by default these permissions are assigned to the 'all' group, but I'm revoking them from the 'all' group and assigning them to the 'user' group instead (because the 'all' group would override the "banned" group)

LocalSettings.php
Amanda1234567: @Reception123 I know that RevokePermissions works, because you aren't supposed to set anything to "true" in LocalWiki (I believe)

Reception123: Well you would set it false if it were in LocalWiki. As I'm not sure on this I'm leaving someone else to merge.

Amanda1234567:
@JohnFLewis ?

PR #1403, opened 2017-04-24
Setting Westmarcheswiki to allow external images.

Ithiloneth:
We wish to display external images and other media on our site as to comply with copyright law for copyrighted material. Our admin, Ithiloneth ahs spoken to a +JohnLewis on the Miraheze IRC about this to get a feel for whether it might be appropriate or not before asking for this pull.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2398,7 +2398,7 @@ $wgConf->settings = array(
 	// Allow External Images
 	'wgAllowExternalImages' => array (
 		'default' => false,
-		'westmarcheswiki' => true
+		'westmarcheswiki' => true,
 	),
 
 	// FlaggedRevs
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2394,6 +2394,12 @@
 		'wisdomsandboxwiki' => '_blank',
 		'yacresourceswiki' => '_blank',
 	),
+	
+	// Allow External Images
+	'wgAllowExternalImages' => array (
+		'default' => false,
+		'westmarcheswiki' => true
```
NDKilla: I believe this should have a comma after the false (based on the above array) however I don't believe it's acruallly a syntax error.

Ithiloneth: That seems indeed to be right. My bad; I am still learning PHP.

Ithiloneth: I see what NDKilla means. I will be making another pullrequest with the fix implemented and I hope doing so is the right method for going about things... If not; I am sorry and new ^^

PR #1405, opened 2017-04-24
Enable AccessControl on westmarcheswiki per T1712

Amanda1234567:


Reception123:
Please reopen when you can confirm (and link the on-wiki page here) that they agree that Miraheze is not responsible for any data leaks that could be caused by this extension.

Ithiloneth:
Admin on Westmarches here:
I've added a statement on our rules page which is available to all those with an account on our Wiki. Since the Wiki is private that should suffice, but if not I will make it appear on our Main Page or Terms of Use?

As we wish the ability to restrict not just editing and deleting of pages, but also the reading, an extension will have to be installed. Such an extension degrades the security of the page such that Miraheze (our service providers) can no longer take any responsibility for the protection of our content. This is notice that they will not, and cannot, be held responsible for any leak of content from this Wiki from the time of extension installation.

PR #1406, opened 2017-04-24
Disable TorBlock on WikiCanada

Amanda1234567:
I don't see a need for it, for a few reasons:

A). There has only been one instance of serious spam. All other attempts have been stopped before they got out of hand
B). I have no evidence that any of the vandals/spammers were connecting through Tor
C). I don't want to be restrictive as to block an entire web provider from accessing my wiki

I will revert this change if I see clear reason to


Amanda1234567:
Closed per comments by @NDKilla on IRC

PR #1410, opened 2017-04-25
Enable <img> tag on Westmarches

Ithiloneth:
I hope that setting [default = false] is correct. I have tried to verify that it is not being set elsewhere in this file and that false should be the defualt value.

https://www.mediawiki.org/wiki/Manual:%24wgAllowImageTag

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2406,6 +2406,12 @@ $wgConf->settings = array(
 		'default' => false,
 		'westmarcheswiki' => true,
 	),
+	
+	// Allow HTML <img> tag
+	'wgAllowwgAllowImageTag' => array (
+		'default' => false,
+		'westmarcheswiki' => true,
+	),
 
 	// FlaggedRevs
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2406,6 +2406,12 @@
 		'default' => false,
 		'westmarcheswiki' => true,
 	),
+	
+	// Allow HTML <img> tag
+	'wgAllowwgAllowImageTag' => array (
```
Amanda1234567: You don't need a space between "array" and the (

Ithiloneth:
Thank you Amanda1234567.

LocalSettings.php
```diff
@@ -2406,6 +2406,12 @@
 		'default' => false,
 		'westmarcheswiki' => true,
 	),
+	
+	// Allow HTML <img> tag
+	'wgAllowwgAllowImageTag' => array(
```
Reception123: there should only be one wgAlowImageTag, I presume it's a typo

PR #1411, opened 2017-04-25
Enable AccessControl on Westmarches

Ithiloneth:
This is based on a previous pull request which was closed and unable to be re-opened.

Please see here for confirmation of leak responsibility denial notice on our Wiki: https://westmarches.miraheze.org/wiki/Main_Page


Reception123:
I confirm text is on wiki.

PR #1413, opened 2017-04-25
Change TestWiki userrights abilities per IRC

Amanda1234567:


Reception123:
This is not the error

PR #1414, opened 2017-04-25
TestWiki userrights changes per IRC

Amanda1234567:
A). Allowing sysops to add/remove testgroup

B). Completely remove exampleuser as group is defunct

C). Remove duplicate group assignement/revocations

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3524,35 +3524,17 @@ $wgConf->settings = array(
 		),
 		'+testwiki' => array(
 			'bureaucrat' => array(
-				'testgroup',
 				'bureaucrat',
 				'sysop',
-				'confirmed',
-				'autopatrolled',
-				'rollbacker',
 			),
 			'consul' => array(
 				'bot',
-				'bureaucrat',
-				'consul',
-				'exampleuser',
-				'testgroup',
-				'sysop',
-				'confirmed',
-				'autopatrolled',
-				'rollbacker',
 			),
 			'steward' => array(
 				'consul',
-				'bot',
-				'bureaucrat',
-				'consul',
-				'exampleuser',
+			),
+			'sysop' => array(
 				'testgroup',
-				'sysop',
-				'confirmed',
-				'autopatrolled',
-				'rollbacker',
 			),
 		),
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3524,35 +3524,17 @@
...
 				'bureaucrat',
 				'sysop',
-				'confirmed',
-				'autopatrolled',
-				'rollbacker',
 			),
 			'consul' => array(
 				'bot',
-				'bureaucrat',
-				'consul',
```
The-Voidwalker: consuls still need to assign other consuls

Amanda1234567:
@The-Voidwalker fixed

NDKilla:
I just looked at the unexpanded diff but doesn't this make it so that Consuls can't add bureaucrats?

Amanda1234567:
Bureaucrats can add other bureaucrats. All consuls are also bureaucrats, so we don't need two groups being able to add the same group.

Amanda1234567:
blockquote, div.yahoo_quoted { margin-left: 0 !important; border-left:1px #715FFA solid !important; padding-left:1ex !important; background-color:white !important; } Bureaucrats can add other bureaucrats. All consuls are also bureaucrats, so we don't need two groups being able to add the same group.


On Tuesday, April 25, 2017, 7:27 PM, Justin <notifications@github.com> wrote:


I just looked at the unexpanded diff but doesn't this make it so that Consuls can't add bureaucrats?

NDKilla:
So if a Consul, for whatever reason, has their bureaucrat removed, they shouldn't be able to add themselves back?
That's not how it's ever worked and not how it should work. The groups shouldn't contain duplicate user rights but multiple groups can add the same groups.

Amanda1234567:
@NDKilla I've enabled the "allow maintainer edits" feature on this PR - you can make any changes you would like befor merging

revi:
It is usually requester's responsibility to fix your own mistakes/requested changes. Sysadmins might fix the changes locally if it is trivial but this doesn't seem to be the case here.

Amanda1234567:
I'm closing this as I do not agree with the changes suggested and I'm honestly fine with leaving things as they are

PR #1417, opened 2017-04-26
Disable anonymous editing on hlptestwiki per T1720

Amanda1234567:


Reception123:
@Amanda1234567 This is done with wmgEditingMatrix in LocalSettings.php, not like this.

Amanda1234567:
@Reception123 ????

PR #1421, opened 2017-04-27
Define local "sysadmin" group on Meta

Amanda1234567:
I'm doing this for a couple of reasons.

A). The wiki creator policy states that "wiki creator rights are granted at the discretion of a sysadmin". However, currently, non-steward sysadmins need to assign "userrights" in the global sysadmin group and then assign local wiki creator rights on Meta. It would be nice if sysadmins could just assign wiki creator directly

B). Technically, the managewiki-restricted right should be sysadmin only, as the top of Special:ManageWiki says "if you cannot change all settings you should contact a system administrator".

C). It would be good if sysadmins could revoke bureaucrat rights on Meta, since we have more sysadmins than stewards currently

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3499,6 +3499,12 @@ $wgConf->settings = array(
 				'commentadmin',
 			),
 		),
+		'+metawiki' => array(
+			'sysadmin' => array(
+				'wikicreator',
+				'sysadmin',
+			),
+		),
 		'+trexwiki' => array(
 			'co' => array(
 				'ceo',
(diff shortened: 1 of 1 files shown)


JohnFLewis:
A) Requests are so rare currently are usually deferred to operations to decide on by others (who are all stewards right now).

B) Stewards have this in order to implement any sort of consensus they need to without defaulting everything to sysadmins. The only thing covered by "restricted" is public/private which is a steward-specific matter at it relates wholly to communities and nothing technical.

C) This applies everywhere though? Stewards are the only ones able to remove bureaucrats on most wiki and as this is a community issue, it's not appropriate for sysadmins to intervene. Furthermore there only 2 bureaucrats on meta.

PR #1434, opened 2017-05-04
Userrights update for WikiCanada

Amanda1234567:

PR #1435, opened 2017-05-04
Redo #1434

Amanda1234567:
Closed improperly by Labster


Reception123:
It was not closed improperly. This group is pointless and all it does is
create confusion. There is no point in this group as the founder cannot add
themselves anyway.

Amanda1234567:
Notice that the new group is not the only change in this PR.

Also the purpose of the group was to have checkuser and oversight rights assigned to a separate group, as "steward" typically refers to global stuff

Reception123:
That still does not explain the point of this group. You can call it
"checkuser and oversight" but not a confusing name

Amanda1234567:
Can I call the group "staff"?

LulzKiller:
The build failed so you'd need to make changes for this to be accepted regardless of the actual issue of user groups.

JohnFLewis:
I don't get why this cant be left as stewards or put back in the respect CheckUser/oversight groups. I'm not going to merge another rearrangement of cu/os that make it so stewards have to give themselve local steward to then assign rights locally when the current system is a simple assign/revoke system.

Also there is a a syntax error I have comnented inline on it.

LocalSettings.php
```diff
@@ -4267,35 +4266,31 @@
...
-				'checkuser' => true,
-				'checkuser-log' => true,
 			),
-			'edit-filter-manager' => array(
-				'applychangetags' => true,
-				'changetags' => true,
-				'abusefilter-modify' => true,
-				'abusefilter-modify-restricted' => true,
-				'abusefilter-revert' => true,
 			),
```
JohnFLewis: Unopened array. Thus line needs to go.

PR #1437, opened 2017-05-04
Add $wgAbuseFilterRestrictions

Amanda1234567:


Arcane21:
I looked over this commit, and I'd advise Miraheze Staff against it.

It effectively leaves the Abuse Filter toothless except to block IP ranges on the wikicanada wiki, and I see no good reason to defang that extension on any individual wiki.

The-Voidwalker:
~~Actually, all it does is makes rangeblock the only abusefilter function that requires the right `abusefilter-modify-restricted` on wikicanadawiki.~~

~~It has no effect on the current global abusefilter setup.~~

The above would apply if the proper syntax had been used.

`'default' => array( 'block', 'degroup', 'blockautopromote', 'rangeblock' ),`
`'wikicanadawiki' => array( 'rangeblock' ),`

Amanda1234567:
Happened to wake up for a minute and saw this comment. 

I will not be active for the rest of the evening as I am resting from a cold

The-Voidwalker:
Hold on, it might be that the documentation is not up to date. Something is not adding up

The-Voidwalker:
Ok, turns out you had it right the first time, except the variable name is `wgAbuseFilterRestrictions`. My apologies. I'm going to go edit the documentation on mediawiki.org.

Arcane21:
Glad to hear I was in error. 

I'm just glad to see the ability of staff to globally police all wikis regardless of individual preferences per wiki is not compromised in any way.

PR #1439, opened 2017-05-05
Userrights changes for WikiCanada

Amanda1234567:


Reception123:
Staff is more okay as a name but why not just leave it at Steward?

Amanda1234567:
@Reception123 because when I see "steward" I think of only global CentralAuth actions. I want one group that has all restricted rights, and "staff" is a less misleading name

JohnFLewis:
Stewards can only add groups that exist on metawiki.

Amanda1234567:
@JohnFLewis I thought that stewards can do userrights-interwiki from Meta and add any/all groups that exist on the local wiki specified

JohnFLewis:
No, they have to exist on meta to be assigned. It's a MediaWiki bug that's been waiting years for a fix.

Amanda1234567:
Okay, so is it possible to configure settings so that the "founder" group can add/remove the "staff" group to all users **except** themselves?

(i.e. If a steward needs rights from the staff group, I would add the group to their account locally, but I wouldn't be able to assign the right to myself)

JohnFLewis:
Not a possibility in MediaWiki.

Amanda1234567:
@JohnFLewis I've changed it back to steward

LocalWiki.php
```diff
@@ -195,6 +194,7 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
 	unset( $wgGroupPermissions['oversight'] );
 	unset( $wgGroupPermissions['autoreview'] );
 	unset( $wgGroupPermissions['editor'] );
+	unset( $wgGroupPermissions['steward'] );
```
JohnFLewis: needs to be removed

Amanda1234567:
@JohnFLewis done

PR #1448, opened 2017-05-06
Create local "sysadmin" group on WikiCanada

Amanda1234567:
This is a null group, similar to the Wikimedia Foundation's "OTRS Member" gloabl group. It contains basic administrator and user permissions (nothing advanced) and is used solely to distinguish global Miraheze sysadmins in the WikiCanada user list

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3681,6 +3681,7 @@ $wgConf->settings = array(
 				'importer',
 				'uploader',
 				'template-editor',
+				'sysadmin',
 			),
 		),	
 		'+wikidolphinhansenwiki' => array(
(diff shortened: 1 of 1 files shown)


Reception123:
I see absolute no point in this group, as sysadmins can already be seen here: https://wikicanada.miraheze.org/w/index.php?title=Special%3AGlobalUsers&username=&group=sysadmin 

We have no need for "null" groups and sysadmins already have these (and more) global permissions, so the local ones are not needed. 
Also, two sysadmins are already blocked on your wiki so there is no point being blocked but being in the "sysadmin" user group.

Reception123:
And also, founder being able to add sysadmin is also unacceptable as you could add any user you wanted and it would cause confusion

PR #1451, opened 2017-05-07
Prevent global abuse filter 12 from taking actions on WikiCanada

Amanda1234567:
I don't like the fact that global abuse filter 12 has blocking and removing groups enabled. IMHO those restricted actions should never be enabled for global filters. As a result, I am preventing said filter from taking those actions on WikiCanada

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -203,6 +203,10 @@ $wgConf->settings = array(
 			'degroup' => false,
 			'rangeblock' => true,
 		),
+		'wgAbuseFilterDisallowGlobalLocalBlocks' => array(
+			'default' => false,
+			'wikicanadawiki' => true,
+		),
 	),
 
 	// Anti-spam


Reception123:
This is for me very suspicious, that you want to remove global abuse filters. Abuse filter 12 is directed to prevent spam to other links. Spam/Advertising goes against Miraheze policies, and by doing this you would be allowing it on a Miraheze wiki.

Amanda1234567:
@Reception123 not necessarily. I'd be fine if the filter:

A). Had a more descriptive name than just "Anti-harassment filter", and
B). Was **only** set to disallow. I don't approve of blocking and degrouping being enabled globally

Reception123:
I do not understand what problem you have with this filter. It disallows a harassment only site. It is very unlikely that you will trigger this filter, and honestly I don't understand why you'd want to get rid of it.

Amanda1234567:
@Reception123 I **don't** want to get rid of it. All I want is for it to disallow the site, but not to block and degroup when tripped. It's fine if you want to block and degroup on Meta, but I don't approve of those actions being taken on WikiCanada unless I specify them (and so far I've only specified them once)

labster:
The only reason you want to remove Abuse Filter 12 is so that you can take criminal action against me personally.  Sorry, not going to happen.

Amanda1234567:
Not true. Quit the personal attacks Labster

Reception123:
I've removed the two things as they don't do much anyway and it has only been triggered once. If any sysadmin disagrees feel free to add back.

labster:
I've restored those two things, as they are important to me.

Amanda1234567:
Clearly nobody is listening to me. For the last time:

**I DO NOT WANT TO REMOVE THE FILTER.** The only thing I want is for the filter to not be blocking and degrouping globally. That's all.

Reception123:
We understood. Please don't use caps, it's not necessary

PR #1452, opened 2017-05-07
Opt-out WikiCanada from global sitenotices

Amanda1234567:


Amanda1234567:
@JohnFLewis It didn't work

PR #1459, opened 2017-05-08
Disable BootstrapMediaWiki

FelixINX:
So I was playing around with the theme, created the template Bootstrap:Subnav and then nothing works! It always give me a 500 error and I don't wont to crash the server. Even if I want to go back to the settings page and change my theme to the default one, I can't! :cry: 

Here is the end of the story between my wiki and the skin :stuck_out_tongue_winking_eye:


Reception123:
The server would probably not crash from that ;)

PR #1465, opened 2017-05-10
Fix Settings for "PruebaWiki"

AlvaroMolinaCL:
Fix Settings for FlaggedRevs Extension on "PruebaWiki". Sorry for the repeated pull requests, but I've needed to see live configurations to know what they consist of and see how well they are. Thanks,

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3587,9 +3587,6 @@ $wgConf->settings = array(
 		),
 		'+pruebawiki' => array(
 			'sysop' => array(
-				'confirmed',
-				'rollbacker',
-				'autopatrolled',
 				'autoreview',
 				'editor',
 				'reviewer',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3591,6 +3586,15 @@
 			),
 		),
 		'+pruebawiki' => array(
+			'sysop' => array(
+				'confirmed',
```
Reception123: These rights (confirmed, rollbacker and autopatrolled) are already in "default" so there is no need to re-add them.

LocalSettings.php
```diff
@@ -4517,6 +4521,15 @@
...
```
Reception123: Same comment as above.

LocalSettings.php
AlvaroMolinaCL: Done.

LocalSettings.php
AlvaroMolinaCL: Done.

PR #1466, opened 2017-05-10
Remove AbuseFilter degrouping on WikiCanada

Amanda1234567:
The "Abuse filter" account does not have the technical ability to degroup as editing userrights is restricted to founder and stewards


Amanda1234567:
Nevermind. I realized that the Abuse filter account is exempt from userrights such as degrouping, etc.

PR #1479, opened 2017-05-15
Add local CVT group on Meta

Amanda1234567:
Per recently closed RFC

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3584,6 +3584,11 @@ $wgConf->settings = array(
 				'commentadmin',
 			),
 		),
+		'+metawiki' => array(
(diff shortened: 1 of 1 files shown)


JohnFLewis:
nothing against it - I just have it done in a set of commits I'm pushing in a few! Plus sysops shouldn't be able to give cvt

Amanda1234567:
@JohnFLewis I just thought sysops should as there are only two bureaucrats on Meta...

Otherwise we would have to create a "sysadmin" group locally on Meta

JohnFLewis:
Stewards exists? Plus they're the ones who grant the rights.

PR #1480, opened 2017-05-15
Add Draft: and Official: namespaces to tmewiki

stranger195:
I don't know why but for some reason there are changes I didn't type. Edit conflict?


JohnFLewis:
Clashes with changes I made just before this was opened. Above commit makes said changes, thanks anyway!

PR #1481, opened 2017-05-15
Move TorBlock to GlobalExtensions

Amanda1234567:
TorBlock has been in LocalSettings with default = true for a while now, I disabled it on WikiCanada and then was immediately spammed. I think that it is something Miraheze should have globally


Amanda1234567:
Sysadmins may support/oppose/comment on this but please do not close it. Thanks!

PR #1484, opened 2017-05-16
Enable 2FA for founder on WikiCanada

Amanda1234567:
I think this is the best way to prevent my account(s) from being compromised


Amanda1234567:
I'll close this for now as SPF is not highly active as of lately.

PR #1492, opened 2017-05-17
Allow crats to modify CVT + allow CVT to remove themselves

Amanda1234567:
I think that this is appropriate.

Meta Bureaucrats will almost always be stewards or sysadmins, and there's no reason why CVT's shouldn't be able to remove themselves.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3603,6 +3603,11 @@ $wgConf->settings = array(
 				'commentadmin',
 			),
 		),
+		'+metawiki' => array(
+			'bureaucrat' => array(
+				'cvt',
+			),
+		),
 		'+trexwiki' => array(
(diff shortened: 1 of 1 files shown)


JohnFLewis:
CVT can't edit global group permissions so there not able to remove themselves.

Bureaucrats can also not be stewards and crats can't edit global groups. 

Therefore this is ineffective.

Amanda1234567:
Couldn't CVT's remove themselves without global permissions with `$wgGroupsRemoveFromSelf`

NDKilla:
No. They could remove themselves from the local group but it's moot since they can't remove themselves from the global group. 

Membership = global. 

The local group only exists to enforce logging on MetaWiki, like Steards

PR #1497, opened 2017-05-17
Override default CentralAuth rights

Amanda1234567:


Reception123:
Is there a particular reason you want to disallow these rights?

revi:
Well, stewards must do all actions on meta so it's no meaning anyway.

Amanda1234567:
> Well, stewards must do all actions on meta so it's no meaning anyway.

Indeed, and therefore I'm trying to cleanup my ListGroupRights a little

PR #1498, opened 2017-05-18
Officially restrict WikiCanada

Amanda1234567:
Currently being managed by ProtectSite and AbuseFilter but this makes it official


Amanda1234567:
Nevermind

PR #1508, opened 2017-05-21
Enable FlaggedRevs and change group permissions in isvwiki

thelevking:
Group permissions:
- Stop editing of CSS/JS by anons
- Remove default patrolling from pages
- Replace it with FlaggedRevs-based settings for same groups
- Move restrictions for users

FlaggedRevs:
- Articles/Files/Templates/Categories/Modules for review
- Protection-like settings
- One status setting
- No autopromotion

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1204,6 +1204,7 @@ $wgConf->settings = array(
 	'wmgUseFlaggedRevs' => array(
 		'default' => false,
 		'extloadwiki' => true,
+		'isvwiki' => true,
 		'pruebawiki' => true,
 		'styleguidesheetwiki' => true,
 		'trexwiki' => true,
@@ -2491,6 +2492,13 @@ $wgConf->settings = array(
 			NS_HELP,
 			NS_PROJECT,
 		),
+		'isvwiki' => array(
+			NS_MAIN,
+			NS_FILE,
+			NS_TEMPLATE,
+			NS_CATEGORY,
+			828,
+		),
 		'trexwiki' => array(
 			NS_ARTIKEL,
 			NS_FILE,
@@ -2499,6 +2507,7 @@ $wgConf->settings = array(
 	),
 	'wmgFlaggedRevsProtection' => array(
 		'default' => false,
+		'isvwiki' => true,
 		'pruebawiki' => true,
 		'wikicanadawiki' => true,
 	),
@@ -2508,6 +2517,11 @@ $wgConf->settings = array(
 				'quality' => 1,
 				'levels' => 2,
 				'pristine' => 3,
+			),
+		),
+		'isvwiki' => array(
+			'accuracy' => array(
+				'levels' => 1,
 			),
 		),
 	),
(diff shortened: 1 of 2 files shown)


LocalSettings.php
```diff
@@ -2491,6 +2492,13 @@
 			NS_HELP,
 			NS_PROJECT,
 		),
+		'isvwiki' => array(
+			NS_MAIN,
+			NS_FILE,
+			NS_TEMPLATE,
+			NS_CATEGORY,
+			828,
```
Reception123: not sure this is possible

LocalWiki.php
```diff
@@ -83,6 +83,31 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
 	$wgGroupPermissions['*']['createpage'] = true;
 }
 
+if ( $wgDBname === 'isvwiki' ) {
+	$wgGroupPermissions['*']['editmyusercss'] = false;
+	$wgGroupPermissions['*']['editmyuserjs'] = false;
+	$wgGroupPermissions['*']['writeapi'] = false;
+	$wgGroupPermissions['user']['editmyusercss'] = true;
```
Reception123: True statements need to be done in wgGroupPermissions on LocalSettings

LocalWiki.php
```diff
@@ -83,6 +83,31 @@ function onClosedSiteNoticeAfter( &$siteNotice, $skin ) {
...
+	$wgGroupPermissions['sysop']['autopatrol'] = false;
+	$wgGroupPermissions['autopatrolled']['patrol'] = false;
+	$wgGroupPermissions['confirmed']['patrol'] = false;
+	$wgGroupPermissions['sysop']['patrol'] = false;
+	
+	$wgGroupPermissions['autopatrolled']['autoreview'] = true;
+	$wgGroupPermissions['bot']['autoreview'] = true;
+	$wgGroupPermissions['confirmed']['autoreview'] = true;
+	$wgGroupPermissions['confirmed']['review'] = true;
+	$wgGroupPermissions['confirmed']['autoconfirmed'] = true;
```
Reception123: same as above

LocalSettings.php
thelevking: NS_MODULE as an alternative? Scribunto will unreview the pages as I am aware.

Reception123: If you think it would work with a number, than we can leave it

thelevking: It works in https://noc.wikimedia.org/conf/highlight.php?file=flaggedrevs.php

LocalWiki.php
thelevking: Fixed (both things).

thelevking:
`[2d6456ebdd3d1c980bac1a1a] 2017-05-21 12:22:56: Fatal exception of type "Exception"`

Don’t really know why, but there’s some errors. Module thing broke?

Reception123:
@thelevking The error is  "FlaggedRevs given invalid tag value!"

thelevking:
@Reception123 maybe you can just s/accuracy/status? I copied that setting from Wikimedia config and honestly didn’t know.

Reception123:
Could be

PR #1509, opened 2017-05-21
Fixing an error on isvwiki (accuracy to status)

thelevking:
Might fix errors from #1508.


thelevking:
Still produces errors. I think I should have duplicated "pristine" & "quality" settings too? Eh, sorry for taking your time, I guess I need to produce another pull request.

PR #1547, opened 2017-06-11
2 extensions (isvwiki) & global setting for TemplateSandbox

thelevking:
- Minor style changes ("defualt", 828 changed to WMG_NS_MODULE)
- Added DPLForum with configuration to stop editing forum talks (namespace 111).
- Added VisualEditor with config settings (UseSingleEditTab = true, ShowBetaWelcome = false). Standard config for ShowBetaVelcome: https://www.mediawiki.org/wiki/Extension:VisualEditor#Complete_list_of_configuration_options
- Added Module namespace to be globally in wgTemplateSandboxEditNamespaces since Scribunto is used globally, see https://noc.wikimedia.org/conf/CommonSettings.php.txt for the same configuration in Wikimedia projects.

Before approving this request please go into Parsoid repository and add "isv" there (I hope it is not an inconvenience, I just don't want to make another fork).

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1095,6 +1095,7 @@ $wgConf->settings = array(
 		'allthetropeswiki' => true,
 		'extloadwiki' => true,
 		'inkubatorwiki' => true,
+		'isvwiki' => true,
 		'rpgbrigadewiki' => true,
 	),
 	'wmgUseDuplicator' => array(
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -5348,6 +5355,14 @@
 		'wisdomsandboxwiki' => "//$wmgUploadHostname/wisdomsandboxwiki/b/be/Sandbox_Logo.png",
 	),
 
+	// TemplateSandbox
+	'wgTemplateSandboxEditNamespaces' => array(
```
Reception123: Why does the default need to be defined if there is no alternative?

PR #1550, opened 2017-06-13
enable visualeditor on zharkunuwiki per T1905

Pix1234:


Pix1234:
Will need to fix this patch do not merge

PR #1552, opened 2017-06-13
fix

The-Voidwalker:
strings with single quotes don't parse for variables


Reception123:
I remembered that they do, but maybe I'm wrong.

PR #1563, opened 2017-06-18
T1873 - Define Tweeki as default skin for corydoctorow.miraheze.org

penguinstyles:


Reception123:
Looking at https://phabricator.miraheze.org/T1790#35384 it seems like there is some error with installing Tweeki, that's why it's not enabled on extload. Until the error is fixed it cannot be enabled on any wiki.

PR #1569, opened 2017-06-22
T1900 - Enable GuidedTours on isv.miraheze.org

penguinstyles:
**See also: https://github.com/miraheze/mediawiki/pull/87**

This pull request enables the following extensions on isv.miraheze.org:

- EventLogging
- GuidedTour

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1167,6 +1167,7 @@ $wgConf->settings = array(
 	),
 	'wmgUseEventLogging' => array(
 		'default' => false,
+		'extloadwiki' => true,
 		'isvwiki' => true,
 	),
 	'wmgUseFancyBoxThumbs' => array(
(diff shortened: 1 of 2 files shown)


LocalSettings.php
```diff
@@ -1165,6 +1165,10 @@
 		'extloadwiki' => true,
 		'lunacyindwiki' => true,
 	),
+	'wmgUseEventLogging' => array(
+		'default' => false,
```
Reception123: all extensions need to be loaded on extloadwiki first

penguinstyles: .-.

Reception123:
@penguinstyles As I want to merge this now so I can finish, I've applied the new changes myself. Extensions  have to **only** be loaded on extloadwiki first, as if they do not work it would bring the other wiki down, which we don't want

PR #1592, opened 2017-07-03
add proxybot group

The-Voidwalker:
should be everything proxybot needs to function (besides generic bot group)


Reception123:
user is steward, therefore automatic approval from that point of view

PR #1595, opened 2017-07-05
Enable ImageMap on tme.miraheze

stranger195:
Is there anything else I have to add?


Reception123:
@stranger195 what do you mean?

stranger195:
@Reception123 To make ImageMap working.

Reception123:
@stranger195 No, enabling the extension is enough. It should work in ~3 minutes.

PR #1603, opened 2017-07-13
Add second-year anniversary

revi:
I know this is going to need some proofreading.


revi:
Scheduled to publish after 10:00 UTC today.

PR #1608, opened 2017-07-17
protectsite on tnoteswiki and protectsite config per T2003

The-Voidwalker:


LocalSettings.php
```diff
@@ -5708,6 +5709,19 @@
...
+	// Protect site config
+	'wgProtectSiteLimit' => array(
+		'default' => '1 week',
+		'tnoteswiki' => 'indefinite',
+	),
+	'wgProtectSiteDefaultTimeout' => array(
+		'default' => '1 hour',
+		'tnoteswiki' => '2 hours',
+	),
+	'wgProtectSiteExempt' => array(
```
Reception123: Since this is the default, why do we have to define it?

The-Voidwalker: No idea, should I remove?

labster: I'm not sure if we must have these values, but if we have custom values for extension variables, I'd prefer that we have the default setting listed too.

Reception123: Unless a wiki wants a different value than the default, I'm not sure why we should be listing the default. That means we'd have to list all the default values for all the extensions.

Reception123: It's not a big deal if we do define the default, of course, just don't see why this should be taking up more space

Reception123:
@The-Voidwalker Sorry for the confusion. I didn't mean remove those defaults, those have to stay! I mean't removing the 'wgProtectSiteExempt' variable only.

PR #1611, opened 2017-07-20
change Default Skin for girlsfrontlinewiki per request T2013

videojeux4:
plus: reorganization of the defaultskin aray

PS: It is normal that darkvector is enabled by default ?


Reception123:
@videojeux4 It is not normal. Could you do me a favor and make the default for darkvector false and then enable it on this wiki?

videojeux4:
Done

PR #1613, opened 2017-07-20
T2012: set refreshed off and make apex default

Pix1234:
for thebb.miraheze.org


LocalSettings.php
```diff
@@ -1809,7 +1809,7 @@
 		'extloadwiki' => true,
 		'grandtheftautowiki' => true,
 		'inazumaelevenwiki' => true,
-		'thebbwiki' => true,
+		'thebbwiki' => false,
```
Reception123: If you're settings something to false, but the default is false, just remove it.

PR #1620, opened 2017-07-24
Extension request for n2go.miraheze.org

CnocBride:
Enabling VisualEditor and Youtube per request T2029


Reception123:
Numbers should go first on the list, so it should be before the other wikis.

PR #1661, opened 2017-08-24
Setup for cdcwiki T2122 T2120

Pix1234:


Reception123:
@Pix1234 you need to actually enable the extension first

PR #1668, opened 2017-08-27
Use wfLoadSkins

MacFan4000:


GlobalSkins.php
```diff
@@ -1,5 +1,7 @@
 <?php
-wfLoadSkin( 'CologneBlue' );
-wfLoadSkin( 'Modern' );
-wfLoadSkin( 'MonoBook' );
-wfLoadSkin( 'Vector' );
+wfLoadSkins( [
+	'CologneBlue',
+	'Modern' ,
```
Reception123: there's no space between the apostrophe and comma.

GlobalSkins.php
```diff
@@ -1,5 +1,7 @@
...
+	'MonoBook' ,
```
Reception123: same as above

PR #1673, opened 2017-08-28
Add translationadmin group on metawiki

MacFan4000:
To allow users to help prepare pages for translation and manage translation, without having to be put in the sysop group.


Reception123:
Is this really necessary? I've marked all the pages that you prepared or changed for translation fairly quickly after. IMO there's not really a need for this group at the time.

MacFan4000:
No, just thought it might be useful. Btw I plan to prepare more and more things for translation.

MacFan4000:
It can be useful when no sysops are online.

Reception123:
Even so, it's not that urgent to mark translations. I guess if you really think it's necessary this could be merged, but you would of course also need to request the permission on Meta.

Also, this group is unassignable (locally).

LocalSettings.php
```diff
@@ -4624,6 +4629,12 @@
 			'wikicreator' => array(
 				'createwiki' => true,
 			),
+			'translationadmin' => array(
+				'pagelang' => true,
+				'pagetranslation' => true,
+				'translate-import' => true,
+				'translate-manage' => true,
+				),
```
Reception123: this array needs one tab less, like the one above

MacFan4000:
I think I'm going to ask that this be merged.

Reception123:
I prefer that your request the right on Meta first, and to merge if the request is successful.

revi:
I'd say we currently don't see any need for this: if you need such power, trust for TA would be just same as ordinary A on meta. We aren't that big as WMF wikis.

MacFan4000:
> if you need such power, trust for TA would be just same as ordinary A on meta.
I disagree with this as its not as easy to abuse Ta as it is to abuse A. IMO, on WMF wikis the group is targeted for people who want to actively contribute to the wiki in question.

JohnFLewis:
> Actually we have more wikis on our farm the the WMF has on their's
Number isn't a measure of size here.

They have an incredibly large infrastructure spread over several datacenters with likely an average disk space of TBs and memory size of hundreds of GB. They operate one of the most accessed websites in the world storing a lot of information accessed globally.

Yes, we might have more wikis but we are by no means "big as WMF wikis".

MacFan4000:
@JohnFLewis @labster @NDKilla @Southparkfan opinions on this?

MacFan4000:
I believe @Pix1234 expressed some support for this on IRC.
PuppyKun:
Reception123: what's the point of translation admin without the translator group
The only purpose was allowing translation admins to add other translators
Reception123
Can't it mark pages for translation?
Oh wait nevermind I thought the PR for that was merged
So I'm not sure if I should add that and keep translationadmin or remove the group
Zppix
i like that idea but i think it should be completely implemented Reception123
Reception123
So you mean add the mark pages for translation permission.
PuppyKun: what do you think?
Zppix
yes

CnocBride:
I don't know if MacFan4000 is still pushing for this and I am kind of in the middle of it. On one hand, Reception is one of the best sysops for merging pull requests and doing all that kind of things (approving translations). On the other hand we all have to sleep (even Reception!) and thus the community goes slightly quiet. I find that changes made from 12AM GMT and 6AM GMT take the 
longest to be approved as most people are either in bed (there're exceptions of course).

So, yes there is a need in times for a translation admin, however, I do not think that delays are large enough to cause much disruption.

MacFan4000:
@Reception123 @CnocBride I would be happy to have this merged.

Reception123:
I'd still like to see some support on Meta other than my vote.
@CnocBride you can add the comment you wrote here on the Meta:RFP if you want to.

MacFan4000:
@Reception123 Support has now been shown by two other users.

Reception123:
Support on Meta by 4 users.

PR #1681, opened 2017-08-31
Add Score (T2142)

MacFan4000:


Reception123:
As I said on IRC, we no longer use extloadwiki for loading extensions. Please change extloadwiki with test1wiki, and add the extension to extension-list.

PR #1684, opened 2017-09-02
Update for unionnorteamericana.miraheze.org

CnocBride:
Adding Graph and Labelled Section Transclusion per request T2147

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1497,7 +1497,7 @@ $wgConf->settings = array(
 		'test1wiki' => true,
 		'sidemwiki' => true,
 		'testwiki' => true,
-		'unionnorteamericanawiki' = true,
+		'unionnorteamericanawiki' => true,
 	),
 	'wmgUseLinkSuggest' => array(
 		'default' => false,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1417,6 +1417,7 @@
 		'bigforestwiki' => true,
 		'calexitwiki' => true,
 		'test1wiki' => true,
+		'unionnorteamericanawiki' = true,
```
Reception123: please use => not = in this case.

LocalSettings.php
```diff
@@ -1496,6 +1497,7 @@
 		'test1wiki' => true,
 		'sidemwiki' => true,
 		'testwiki' => true,
+		'unionnorteamericanawiki' = true,
```
Reception123: same as above.

CnocBride: Crap, I must have forgotten to enter it in.

Reception123: still missign the second one "=>"

CnocBride:
Now if this isn't correct I dunno what will xD.

PR #1699, opened 2017-09-05
Add AddThis (T1596)

MacFan4000:


LocalExtensions.php
```diff
@@ -5,6 +5,10 @@
 //	require_once( "$IP/extensions/AccessControl/AccessControl.php" );
 //}
 
+if ( $wmgUseAddThis ) {
+	require_once "$IP/extensions/AddThis/AddThis.php";
```
JohnFLewis: consistency is really useful with it enclosed in () as all the ones below.

extension-list
```diff
@@ -1,4 +1,5 @@
 $IP/extensions/AJAXPoll/extension.json
+$IP/extensions/AddThis/extension.json
```
JohnFLewis: If this exists, why is it being loaded by using "require_once"?

LocalExtensions.php
MacFan4000: ?

JohnFLewis: require_once( "$IP/extensions/AddThis/AddThis.php" );

MacFan4000:
Oops wasn’t paying attention to that. Fixed it.

PR #1710, opened 2017-09-10
Update LocalSettings.php

robkam:
Need to give  sysops moderation rights on scruffywiki

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4644,6 +4644,11 @@ $wgConf->settings = array(
  				'unreviewedpages' => true,
  			),
  		),
+		'+scruffywiki' => array(
+			'sysop' => array(
+				'moderation' => true,
+			),
+		),
 		'+serinfhospwiki' => array(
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -4644,6 +4644,11 @@
  				'unreviewedpages' => true,
  			),
  		),
+		'+scruffywiki' => array(
+			'sysop' => array(
```
MacFan4000: Fix the indenting please.

Pix1234:
+1

LocalSettings.php
robkam: I can only see tabs in the diff?

MacFan4000: Take a look at how it’s done on other wikis. See how the individual group arrays are indented one tab more then the wiki array? Here instead of that you just indented it one space.

robkam:
Sorry I'm sure you're right but I can only see tabs. Going to try again.

PR #1717, opened 2017-09-13
ProtectSite on infectopedwiki (T2170)

MacFan4000:


LocalSettings.php
```diff
@@ -1931,6 +1931,7 @@
 		'financialfindswiki' => true,
 		'test1wiki' => true,
 		'hasanistanwiki' => true,
+		'infectopedeiki' => true,
```
JohnFLewis: e instead of w to make it eiki

PR #1718, opened 2017-09-13
Restrict editing on openkorebrasilwikiwiki (T2171)

MacFan4000:


LocalSettings.php
```diff
@@ -795,6 +795,9 @@
 			'user' => true,
 			'sysop' => true,
 		),
+		'+openkorebrasilwikiwiki' => array(
+			'anon' => true,
+			),
```
JohnFLewis: intending too much

PR #1746, opened 2017-09-24
Allow wiki admins to add/remove interwiki prefixes

MacFan4000:
It seems to work now without adding to the global table (excluding metawiki)


Reception123:
https://phabricator.miraheze.org/T2186#41399

Reception123:
Yes.. that is the point.

If I am a wiki bureaucrat, I can use interwiki links to link to any malicious site, and the users can just click on it not knowing where they will be redirected.

Unless there is some discussion on this, it's not just going to be merged.

PR #1748, opened 2017-09-27
add metrolook and make it default for appswiki per request T2243

videojeux4:


MacFan4000:
That would be for T2241

PR #1749, opened 2017-09-27
Extensions on appswiki per T2242

MacFan4000:
SocialProfile image directories will need to be copied


Reception123:
Directories copied.

PR #1760, opened 2017-10-01
Remove TranslateConfigHack

MacFan4000:
I’m not sure that this is really necessary.


Reception123:
I will probably test this on test1wiki before merging this, as we can't be sure it won't cause issues.

Reception123:
No evidence that things would work with this being gone.

Tested on test1wiki and as soon as the file is gone, Special:PageTranslation is no longer accessible, therefore obviously this is still needed and not something that we can remove.

PR #1782, opened 2017-10-18
Addition of "IDEA" Namespace in "revitwiki" for T2326

AlvaroMolinaCL:
Addition of "IDEA" Namespace in "revitwiki" for Phabricator Task [T2326](https://phabricator.miraheze.org/T2326).


Reception123:
This also needs to be defined above.

PR #1783, opened 2017-10-18
Disable Anonymous Editions in "paddelnwiki" for T2325

AlvaroMolinaCL:
Disable Anonymous Editions in "paddelnwiki" for Phabricator Task [T2325](https://phabricator.miraheze.org/T2325). Thanks.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -172,6 +172,10 @@ if ( $wgDBname === 'nenawikiwiki' ) {
 	$wgDefaultUserOptions['flow-editor'] = 'visualeditor';
 }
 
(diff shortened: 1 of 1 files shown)


MacFan4000:
This is incorrectly done. While this would work, editing should usually be restricted using wmgEditingMatrix. So you would add an entry for said wiki with the anon option set to true.

AlvaroMolinaCL:
@MacFan4000: Is it possible that you can do it? Right now I'm busy, I've enabled the option so you can also edit the file.

MacFan4000:
I’ll make a new pull request.

MacFan4000:
Redone in #1785

PR #1818, opened 2017-11-03
Make opt out and management of sitenotice easier

JohnFLewis:


Reception123:
@JohnFLewis If only we could make CentralNotice work..

PR #1834, opened 2017-11-10
Update LocalSettings.php

robkam:
config changes for sdiywiki


Reception123:
@robkam Since I have edited the file since you did, there are conflicts that need to be resolved.

Also again the title shouldn't be "Update LocalSettings.php" but it should be what you left as a comment.

robkam:
Sorry about that. As it looked like I’d messed up fonts for other wikis I deleted my request. I’ll try again in a few hours.

PR #1838, opened 2017-11-10
add draft namespace to bonobowiki, scruffywiki and sdiywiki

robkam:


robkam:
It looks already defined at line 125 and 126?

Reception123:
Whoops, yes, you are right.

PR #1850, opened 2017-11-14
two changes

videojeux4:
* Disable editing by anonymous users on lexiquewiki per request T2431
* Enable VisualEditor for modelshipreferencewiki per request T2432


Reception123:
It's preferred if you do them in separate commits

PR #1865, opened 2017-11-23
add permissions for moderation extension

robkam:
and delete empty line above


MacFan4000:
Are you sure you want the permissions to apply to all wikis with the extension or just your wikis.

robkam:
Oops, to all wikis with the extension was done in error. I'd intended the permissions to be only for sdiywiki.

PR #1871, opened 2017-11-24
add moderators group to sdiywiki, and permissions for the same

robkam:


LocalSettings.php
```diff
@@ -4430,6 +4430,13 @@
 				'rollbacker',
 			),
 		),
+		'+sdiywiki' => array(
```
Reception123: This is only for which groups can add which groups, not for actual permissions.

robkam:
not what I'd intended

PR #1873, opened 2017-11-25
sdiywiki moderation user rights

robkam:


LocalSettings.php
```diff
@@ -5402,6 +5411,13 @@
 				'rollbacker',
 			),
 		),
+		'+sdiywiki' => array(
```
Reception123: As I said before, this is wrong, only wgGroupPermissions is for defining permissions.

WgAddGroups and wgRemoveGroups is to define what groups can be removed by what groups.

LocalSettings.php
```diff
@@ -4452,6 +4452,13 @@
...
```
Reception123: As I said before, this is wrong, only wgGroupPermissions is for defining permissions.

WgAddGroups and wgRemoveGroups is to define what groups can be removed by what groups.

LocalSettings.php
```diff
@@ -5038,8 +5045,10 @@
 			),
 		),
 		'+sdiywiki' => array(
-			'sysop' => array(
-				'moderation' => true,
+			'moderators' => array(
```
Reception123: Only this needs to stay.

PR #1875, opened 2017-11-25
sysop needs to be included in moderator group, sdiywiki

robkam:


LocalSettings.php
```diff
@@ -5038,6 +5038,9 @@
 			),
 		),
 		'+sdiywiki' => array(
+			'sysop' => array(
```
Reception123: Here is where the other variables you used in your other PR come in.

You need to use wgAddGroups for allowing sysops to add moderators, and wgRemoveGroups to allow sysops to remove moderators.

PR #1878, opened 2017-11-25
rm redundant entry and incl sysop in moderator group by default, sdiy…

robkam:
…wiki


Reception123:
That configuration is the default, which mostly includes wikis that don't have Moderation, therefore adding that right for them would be pointless.

PR #1895, opened 2017-12-02
also www.sdiy.info > sdiywiki

robkam:


Reception123:
This isn't possible. There can only be one domain, you can have www.sdiy.info as a redirect only.

robkam:
www.sdiy.info as a redirect then will be fine.

PR #1906, opened 2017-12-09
Allow unlimited cats on DPL3 lists in tmewiki

stranger195:
I'm not sure whether this is the correct code. Modify if necessary.


Reception123:
First of all, the location isn't good
Second, you cannot do that with wgConf

PR #1909, opened 2017-12-11
Changing madgenderscience group permissions

taylor-grey:


LocalSettings.php
```diff
@@ -5013,6 +5013,14 @@
 				'upload_by_url' => true,
 			),
 		),
+		'+madgendersciencewiki' => array(
+			'scholar' => array(
+				'editprotected' => true,
+			),
+			'*' => array(
```
MacFan4000: The edit permission should be revoked using wgEditingMatrix. Basically add an entry for your wiki and only set anon to true.

Reception123:
@taylor-grey You also need to use wgAddGroups and wgRemoveGroups so that a group can actually add the "scholar" permission.

PR #1916, opened 2017-12-11
VE on 2 wikis per T2491 and T2514

MacFan4000:


MacFan4000:
@NDKilla the parsoid PR needs to be merged as well/. (It does not trigger the IRC bot for whatever reason).

PR #1933, opened 2017-12-21
T2539: wikielectionwiki config

Pix1234:


Reception123:
Oh and also, there has to be a +

JohnFLewis:
and a >

PR #1938, opened 2017-12-24
Enabling CharInsert for maplewiki

CnocBride:
Per request T2545


Reception123:
Please move the wikis starting with 's' in their correct spot to maintain alphabetical order

PR #1948, opened 2018-01-01
config for tmewiki extra content namespaces

stranger195:


LocalSettings.php
```diff
@@ -4268,6 +4268,22 @@
 		'+thelonsdalebattalionwiki' => array(
 			NS_GLOSSARY => true,
 		),
+		'+tmewiki' => array (
```
Reception123: please remove the space between array and the comma **array(**

PR #1952, opened 2018-01-02
On wiki change

sau226:
I'd like a couple of new user groups created. This request allows consuls to assign a group that lets a user enable 2 factor authentication and disables the enabling (not the use or disabling) of 2 factor authentication for everyone who isn't a group member. This also creates a sysadmin group which should be self assigned by a sysadmin or assigned by a steward to a sysadmin if they join the wiki and I or a community member send a request for the flagging.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4597,6 +4597,7 @@ $wgConf->settings = array(
 				'consul',
 				'testgroup',
 				'sysop',
+				'authenable',
 				'confirmed',
 				'autopatrolled',
 				'rollbacker',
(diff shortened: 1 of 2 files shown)


Reception123:
First of all, for enabling 2FA, that is not your decision, as it is a global Miraheze extension, it can be enabled on all wikis anyway from metawiki or loginwiki, so there's no point in disallowing users to use it on your wiki.

Secondly, I don't see a reason for this 'sysadmin' group or for revoking steward permissions, and steward s generally perform the actions you are trying to restrict on metawiki anyway.

PR #1957, opened 2018-01-05
ncpprcwiki editingmatrix per T2581

The-Voidwalker:


Reception123:
@The-Voidwalker Sysop should probably also be true, or else sysops cannot edit anymore.

PR #1962, opened 2018-01-09
Enable MsUpload and VisualEditor for healthvector

DimGun:


Reception123:
@DimGun Hi, when adding the VisualEditor extension, please note that you also need to add it to https://github.com/miraheze/parsoid/blob/master/parsoid.yaml

Reception123:
I've added it for you in 6bce50b

PR #1981, opened 2018-01-22
add logo and favicon for kunwokwiki per request T2638

videojeux4:


videojeux4:
Close because of c32918b

PR #1990, opened 2018-01-27
Wiki Setup

sau226:
This is a PR to set up a wiki I am bureaucrat on so that it is ready for use. All sysadmins have to do is merge this when ready and I'll take care of the config on wiki. This particular wiki is the LCARS47 wiki.


sau226:
Oh err also if a steward sees this could they grant me the new manager group created by this PR when it is merged? Thanks

PR #1991, opened 2018-01-27
Fix confirmed group globally/by default

sau226:
Autoconfirmed users never had the patrol right. This grants confirmed users the autoconfirmed users' IP based rate limit bypass and removes the confirmed users' right to patrol edits.


sau226:
Err why exactly do confirmed users need the patrol right?

JohnFLewis:
@reception123 easy merge here that’s been waiting a while.

Reception123:
@JohnFLewis I never merged this because I wasn't sure, was this done intentionally at the beginning? Or was it a mistake?

Southparkfan:
Makes sense.

PR #1993, opened 2018-01-30
GlobalUserPage + GlobalPreferences per T1681

MacFan4000:


Reception123:
Blocked by T1681

sau226:
Now we just need an RfC to render a yes or no answer

MacFan4000:
I think its a yes, as there are currently more support votes (3:6).

Reception123:
@MacFan4000 Yikes, this commit has so many new changes and conflicts.

You should either open a new PR, or if you prefer I can directly commit it when I install.

PR #2002, opened 2018-02-03
EnableScaryTranscluding of templatewiki

robkam:


The-Voidwalker:
Scary transclusion is enabled everywhere by [default](https://github.com/miraheze/mw-config/blob/a2b088a5972032ada994df5ae71a9086d94b3352/LocalSettings.php#L3475).

robkam:
then not required

PR #2003, opened 2018-02-03
Optimize travis config for the future and ensure stable is used

Pix1234:


Reception123:
Since I'm not sure this would actually work, and we don't need it for PHP 7 until we get it, so I'll probably merge then.
For now closing per: https://en.wikipedia.org/wiki/Wikipedia:If_it_ain%27t_broke,_don%27t_fix_it

PR #2009, opened 2018-02-11
improve templatewiki logo and favicon

robkam:


CnocBride:
If you make a task on Phabricator, it is common for us to refer to the task in the comment of the pull request.

Per request T2676

PR #2010, opened 2018-02-12
Adding VisualEditor to harrypotter.miraheze.org

CnocBride:
Per request T2719


revi:
Note: Depends on miraheze/parsoid#102

PR #2014, opened 2018-02-13
enable extension Capiunto

robkam:
[Extension:Capiunto ](https://www.mediawiki.org/wiki/Extension:Capiunto ), if it has been installed already [T2614](https://phabricator.miraheze.org/T2614)?


Reception123:
Will install soon.

PR #2031, opened 2018-02-25
add Inactive wiki notice

Reception123:


JohnFLewis:
This says you can remove the site notice at MediaWiki:Sitenotice but you can’t.

PR #2032, opened 2018-02-26
begin removing deleted wikis

The-Voidwalker:


The-Voidwalker:
Whoever merges, please be sure to also remove the histories.wiki cert if not already done.

The-Voidwalker:
I guess I'll revisit this later.

PR #2034, opened 2018-02-27
Enable VisualEditor for volleyball wiki

ebm5025:


ebm5025:
Related to parsoid #105

PR #2054, opened 2018-03-09
infectopedwiki settings per T2783

The-Voidwalker:


Reception123:
AFAIK wmg instead of wg is only used if it's defined in LocalExtensions

PR #2058, opened 2018-03-11
Updated group rights for jadtechwiki

PlasmaLight:


Reception123:
Unfortunately, this would grant access to all userrights (including checkuser and oversight) which is not permitted except for stewards.

PlasmaLight:
Ok, sorry. My intention was indeed to gain the ability of the checkuser right, but I didn't know that this was forbidden. I am the "owner" of the wiki and I think I should have that right, is there any allowed way to get it as an owner?

PR #2069, opened 2018-03-18
enable visualeditor on yourtechnicalservicewiki per req T2881

Wiki1776:


Reception123:
@Wiki1776 When it's VisualEditor or Flow, you also need to add an entry to github.com/miraheze/parsoid

Wiki1776:
@Reception123 Done [add yourtechnicalservice on parsoid for request T2881](https://github.com/miraheze/parsoid/pull/112)

PR #2096, opened 2018-03-25
Allow CVT to add/remove bot flag to/from themselves on metawiki

MacFan4000:
This allows them to prevent themselves from spamming RecentChanges when doing lots of Locks/Blocks etc. NOTE: This does not allow them to add anyone to the group, just themselves.


JohnFLewis:
Why not in LS?

MacFan4000:
@JohnFLewis Where would it be done in LS?

JohnFLewis:
Under group permissions sections with all the other group permission settings.

MacFan4000:
Those variables arn’t even in the file right now afaik, since they havn’t really been used on any wikis.

The-Voidwalker:
https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L5784 for remove from self
add to self needs to be added somewhere around https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L4643

MacFan4000:
Done

Reception123:
Approved by @The-Voidwalker as a steward, therefore good to go.

PR #2110, opened 2018-03-29
enabling extensions on plazmaburstwiki per req t2718

Wiki1776:
enable RelatedArticles, Popups, Tabber and Editcount.


LocalSettings.php
```diff
@@ -2862,7 +2866,6 @@
 		'pfsolutionswiki' => true,
 		'pgnwikiwiki' => true,
 		'plasticssongcontestwiki' => true,
-		'plazmaburstwiki' => true,
```
Reception123: @Wiki1776 This edit removes your last one. Is this intentional?

Wiki1776: no, it's a mistake I'm going to fix.

PR #2118, opened 2018-03-29
enable visualeditor on therubyserverwiki per req T2920

Wiki1776:


Wiki1776:
[Parsoid](https://github.com/miraheze/parsoid/pull/117)

PR #2165, opened 2018-04-14
avalicearchiveswiki, disabling InstantCommons etc.

Creepario:


LocalSettings.php
```diff
@@ -1948,6 +1953,7 @@
 	),
 	'wmgUseMobileFrontend' => array(
 		'default' => true,
+		'avalicearchiveswiki' => true,
```
Reception123: You can remove this line, MobileFrontend is enabled by default as true, so this does nothing.

Creepario: Right. I hope it's now removed.

PR #2172, opened 2018-04-15
Set read only mode for all wiki's

paladox:


paladox:
Should not be merged before 16.00 utc time

PR #2209, opened 2018-05-02
enable Youtube on wakeupsuperwiki per req T3048

Wiki1776:


LocalSettings.php
```diff
@@ -3312,6 +3312,7 @@
 		'uncyclopediawiki' => true,
 		'valentinaprojectwiki' => true,
 		'vandalismwikiwiki' => true,
+		'wakeupsuperwiki' =true,
```
MacFan4000: = should be =>

Wiki1776: Done

PR #2215, opened 2018-05-05
enable youtube and visualeditor on dobotswiki per req T3061

Wiki1776:
Requires review before approval.


paladox:
Needs to be fixed to use => please

PR #2233, opened 2018-05-14
rm readonly + notice + db2 from files

MacFan4000:


Database.php
```diff
@@ -3,7 +3,6 @@
 $wgLBFactoryConf = array(
 	'class' => 'LBFactoryMulti',
 	'sectionsByDB' => array(
-	        // 'wiki' => 'c1'
```
paladox: Re add this please.

LocalSettings.php
```diff
@@ -6179,7 +6179,7 @@ function onSiteNoticeAfter( &$siteNotice, $skin ) {
 	return true;
 }
 */
-
+/*
 // leaving here as placeholder for future db work
 if ( $wgDBname == 'airwiki' ||
```
paladox: Remove all these but keep two for examples please.

JohnFLewis: or remove them all because we already have a site notice thing commented out

Database.php
MacFan4000: This isn’t needed

paladox: Yes it is for reference as we need docs.

PR #2240, opened 2018-05-23
Enable CookieWarning globally

paladox:
This is due to the gdpr comming into force on the 25 of may 2018.

### DIFF
--- a/GlobalExtensions.php
+++ b/GlobalExtensions.php
@@ -11,6 +11,7 @@ wfLoadExtensions( [
 	'Cite',
 	'CiteThisPage',
 	'CodeEditor',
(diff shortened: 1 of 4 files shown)


GlobalExtensions.php
```diff
@@ -40,3 +41,7 @@
 	'WikiEditor',
 	'cldr'
 ] );
+
+// Geolocate here to determine to whom to show the cookie warning
+$wgCookieWarningEnabled = true;
```
MacFan4000: put this in $wgConf with the other CookieWarning config please

JohnFLewis:
Can we kill the ugly and unnecessary comments in LS?

PR #2246, opened 2018-05-29
Update LocalSettings.php

lbt12to12:
added a few extensions to alwikiwiki


paladox:
You can use Special:ManageWiki to do this for your self now :)

PR #2250, opened 2018-06-02
Update Altverse logo

centrist16:
Requesting logo change for Altverse Wiki


paladox:
Hi, this is done through https://altverse.miraheze.org/wiki/Special:ManageWiki now for logos and favicons.

PR #2254, opened 2018-06-10
Grant privileges to Abuse Team on metawiki

sau226:
This patch allows abuse team members to use 2 centralauth features: global account rename and global account oversight.

Global oversight is to be used to hide abusive usernames at the CentralAuth account level when necessary
Global rename is in case an abusive username or normal username needs to be renamed as part of the Abuse Team's duties

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3893,6 +3893,7 @@ $wgConf->settings = array(
 		'+metawiki' => array(
 			'abuse' => array(
 				'centralauth-lock' => true,
+				'centralauth-oversight' => true,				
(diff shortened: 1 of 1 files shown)


Reception123:
There's hideuser and as you say global oversight so I don't think renaming users is necessary.

sau226:
Hideuser is for blocks and is local, global oversight is for centralauth and rename user is for "This allows the staff member to globally rename users from meta, usually for staff account renames or the rare abusive name that is unable to be hid that is found during their work." (as per a certain very large site that does assign the right as such).

PR #2279, opened 2018-07-02
Revert "olegcinemawiki config fixes T3263"

The-Voidwalker:
seems to have brought down the wiki


The-Voidwalker:
Debugging would be preferable, but reverting is faster.

MacFan4000:
Fixed by @NDKilla

PR #2285, opened 2018-07-18
Move wgDefaultSkin to ManageWiki

paladox:


ManageWiki.php
```diff
@@ -1031,6 +1031,13 @@
  * overridedefault: a string/array override default when no existing value exist.
  */
 $wgManageWikiSettings = array(
+	'wgDefaultSkin' => array(
+		'name' => 'Default Skin',
+		'requires' => false,
+		'type' => 'text',
```
JohnFLewis: so I can enter "123" without issue?

ManageWiki.php
```diff
@@ -1031,6 +1031,13 @@
...
+		'overridedefault' => null,
```
JohnFLewis: wgDefaultSkin can never be null.

ManageWiki.php
```diff
@@ -1011,6 +1011,8 @@
 		),
 );
 
+
+
 /**
```
MacFan4000: Why add extra space?

PR #2298, opened 2018-07-25
Add wgLocaltimezone to ManageWiki

paladox:


JohnFLewis:
Why do you have code in the middle of config? Why can't this be in an extension?

PR #2304, opened 2018-07-26
Configure vsfan per T3397

paladox:


LocalSettings.php
```diff
@@ -4672,31 +4676,9 @@
...
-				'project-edit' => true,
-			),
-		),
-		'+zhdelwiki' => array(
-			'confirmed' => array(
-				'createpage' => true,
-				'edit' => true,
-				'move' => true,
-				'upload' => true,
+				'edit' => false,
```
JohnFLewis: => false doesn't work in this variable.

Also please use the relevant $wmgEditingMatrix config

PR #2317, opened 2018-07-30
User perms identical to testwiki for svwiki per T3421

Amanda-Catherine:


The-Voidwalker:
FYI, there's also settings in LocalWiki.php that would need to be copied as well.

PR #2318, opened 2018-07-30
Add cc-by-sa

paladox:


JohnFLewis:
Anyone changing to this license with existing contribs violates the terms of the license.

You can not downgrade a license as you revoke conditions, while you can upgrade a license as you extend conditions.

paladox:
@JohnFLewis this was for the incurbatorwiki who requested it, what do you mean by “Manually add this license and take it off MangeWiki as anyone violating the terms of CC, we can not host.”

Then this license field will have to be added to CreateWiki so that users have a choice of the licenses we allow. 3.0 or 4.0.

paladox:
@JohnFLewis i don't think i we can do manually add the license.

Does that mean i have upgrade incurbatorwiki back to 4.0?

paladox:
@JohnFLewis sould you be able to upgrade incutbatorwiki to 4.0, just so they understand why they cannot use 3.0.

Reception123:
@Paladox Just add it to LocalSettings.php

PR #2376, opened 2018-08-14
Downgrade isvwiki to Wikipedia's CC BY-SA

thelevking:
Most of the content was licensed under CC BY-SA 3.0 and we just didn't notice that it was changed (and that it was a significant change).

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2147,6 +2147,7 @@ $wgConf->settings = array(
 	'wgRightsIcon' => array(
 		'default' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
 		'incubatorwiki' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
+		'isvwiki' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
 		'revitwiki' => "//$wmgUploadHostname/revitwiki/d/d8/All_Rights_Reserved.png",
 	),
 	'wgRightsPage' => array(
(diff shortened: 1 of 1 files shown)


Reception123:
@thelevking For a downgrade, it is mandatory to have the permissions of all contributors to re-license their contributions. This must be written somewhere on-wiki.

thelevking:
Well, the thing is that most of the content was under CC BY-SA 3.0 before, most of my (and others) translations are also from CC BY-SA 3.0 (as they must be licensed under the same terms), so I don't think it can be under the question. If anything, I'd say we haven't noticed that Miraheze was updated to CC BY-SA 4.0, not the other way.

JohnFLewis:
Creative Commons Licenses are intrinsically designed to be backward compatible. Something licensed under 1.0 can be used under 2.0->4.0, something under 2.0 can be licensed under 3.0->4.0, something licensed under 3.0 can be licensed under 4.0. Upgrading a license is a legitimate thing within the licencing terms

LocalSettings.php
```diff
@@ -2147,6 +2147,7 @@
 	'wgRightsIcon' => array(
 		'default' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
 		'incubatorwiki' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
+		'isvwiki' => 'https://meta.miraheze.org/w/resources/assets/licenses/cc-by-sa.png',
```
MacFan4000: Is this needed? We use that image as the default.

JohnFLewis:
Please reopen when permission to relicense has been gained on wiki.

PR #2383, opened 2018-08-20
[Need review] add extension Moderation on Special:ManageWikiExtensions

Wiki1776:
**[Need review]**.


JohnFLewis:
Moderation wasn’t in ManageWiki because it has critical config unassigned (user rights mostly).

Also this doesn’t remove it from LS as expected.

@reception123

Reception123:
@JohnFLewis Oh, sorry, was not aware. There was a request waiting so that is why I wanted to get this done.

PR #2392, opened 2018-08-23
add website on wgAllowExternalImagesFrom for astrobiologywiki and doomsdaydebunkedwiki per T3504 + adiction of Wiki1776wiki

Wiki1776:


LocalSettings.php
```diff
@@ -1865,10 +1865,15 @@
 		'astrobiologywiki' => array(
 			'https://www.science20.com',
 			'https://quora.com',
+			'http://robertinventor.com',
```
paladox: I think https will only work for us (switch to https here please)

LocalSettings.php
```diff
@@ -1865,10 +1865,15 @@
...
 		),
 		'doomsdaydebunkedwiki' => array(
 			'https://www.science20.com',
 			'https://quora.com',
+			'http://robertinventor.com',
```
paladox: I think https will only work for us (switch to https here please)

PR #2393, opened 2018-08-24
Fix strange userrights on weatherwiki

Amanda-Catherine:
Somehow confirmed users got the ability to create newsletters, and account creation got restricted to autoconfirmed users.


Amanda-Catherine:
Never mind

PR #2403, opened 2018-08-27
UserRights for enigmawiki

CnocBride:
So, I think I disabled edit rights for users on enigma.miraheze.org? Then I tried to create a new user group called "scribe". I actually don't know if what I did was correct, so please correct me if I did anything wrong.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4600,6 +4600,18 @@ $wgConf->settings = array(
 				'edit' => true,
 			),
 		),
+		'+enigmawiki' => array(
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -4600,6 +4600,18 @@
 				'edit' => true,
 			),
 		),
+		'+enigmawiki' => array(
```
MacFan4000: Edit is not revoked because your appending to the default and not overiding, but edit perm is controlled using wgEditNatrix

PR #2413, opened 2018-08-29
Add Hub namespace for statisticswiki T3534

Amanda-Catherine:


LocalSettings.php
```diff
@@ -2896,6 +2896,10 @@
 			NS_BOILERPLATE => 'Boilerplate',
 			NS_BOILERPLATE_TALK => 'Boilerplate_talk',			
 		),
+		'statisticswiki' => array(
```
paladox: need to define the namespace.

Amanda-Catherine: @paladox Huh?

paladox: See where we use define() at top of LS

Amanda-Catherine: Done

PR #2417, opened 2018-08-30
Autopromotion settings for igrovyesistemywiki T3536

Amanda-Catherine:


LocalSettings.php
```diff
@@ -6096,6 +6096,41 @@
 				APCOND_EMAILCONFIRMED,
 			),
 		),
+		'+igrovyesistemywiki' => array(
```
The-Voidwalker: this array is not closed

Amanda-Catherine: @The-Voidwalker I don’t see what you are seeing

The-Voidwalker: The array needs to be closed on a newline added after line 6133

PR #2420, opened 2018-08-31
initial deployement of ManageWikiPermissions

Reception123:


JohnFLewis:
also, it's late now buuuuut removing "default" means they no longer exist on any wiki (except managewiki permissions wikis. I kinda missed then first time reviewing)

PR #2423, opened 2018-09-02
Re-remove oversight and checkuser groups from weatherwiki

Amanda-Catherine:
These groups are grandfathered into the steward group, but got re-added during the deployment of ManageWikiPermissions.


JohnFLewis:
$wgGroupPermissions is made using the ext, removing stuff this way is now super inefficient

PR #2425, opened 2018-09-03
Add $wgPFEnableStringFunctions for paliwiki T3554

Amanda-Catherine:


LocalSettings.php
```diff
@@ -4040,6 +4040,12 @@
 	'wgPageTriageInfinitScrolling' => array(
 		'default' => true,
 	),
+	
+	// ParserFunctions
+	'wgPFEnableStringFunctions' => array(
```
paladox: This is already defined here https://github.com/miraheze/mw-config/blob/20397f36409a9151ec27ff9cf6ee2225628ffb0a/LocalSettings.php#L1263

Wiki1776: Ibidem!

PR #2430, opened 2018-09-05
Trim Down ManageWikiPermissions blacklist

Amanda-Catherine:
IMHO the only rights that need to be blacklisted are rights that local wiki bureaucrats should not have the ability to add to themselves or others. All of these rights are already granted to very low-level groups by default (*, user, autoconfirmed, etc) and therefore really don’t need to be restricted and are just adding extra lines of code IMHO.

**Sysadmins: Feel free to merge this whenever you feel it is safe to do so, and also feel free to add comments/make suggestions. However, please DO NOT close this PR. Thank you.**


LocalSettings.php
```diff
@@ -2381,23 +2379,16 @@
 			'managewiki-editdefault',
 			'mwoauthmanageconsumer',
 			'mwoauthmanagemygrants',
-			'mwoauthproposeconsumer',
 			'mwoauthsuppress',
-			'mwoauthupdateownconsumer',
 			'mwoauthviewprivate',
 			'mwoauthviewsuppressed',
-			'read',
```
JohnFLewis: it should not be possible to revoke read from * except by making the wiki private.

LocalSettings.php
```diff
@@ -2381,23 +2379,16 @@
...
 			'renameuser',
 			'requestwiki',
 			'siteadmin',
-			'skipcaptcha',
```
JohnFLewis: allows people to disable captcha on the whole wiki.

LocalSettings.php
```diff
@@ -2381,23 +2379,16 @@
...
 			'suppressionlog',
 			'suppressrevision',
-			'torunblocked',
```
JohnFLewis: allows people to whitelist tor for everyone.

Amanda-Catherine: @JohnFLewis TorBlock was only made a global extension because of LP - prior to that it was an optional opt-in extension. I don’t think we should be forcing users to block Tor unless its a problem for them

LocalSettings.php
Amanda-Catherine: @JohnFLewis some wikis might only want CAPTCHA for anonymous users and they may want to disable it on the whole wiki for any user who has an account, whether it’s an attached global account or an independent locally created account

LocalSettings.php
Amanda-Catherine: @JohnFLewis notice “Open proxies (or similar networks, such as web hosts) that MAY be used abusively“. Not all Tor networks will be used abusively.

Web hosts is one area where I feel that we should only be reactive and block when abused, not block every single one because of the chance of abuse

JohnFLewis: Tor are open proxies, anyone can use them. Further anyone can abuse them and they are infamous in being even harder to predict usage and prevent abuse than even open proxies on the surface web.

You have an issue with a global policy, RfC is the place, not here. This change won't be merged until this right is readded or another valid non-policy based argument is presented.

Amanda-Catherine:
@JohnFLewis torunblocked, skipcaptcha, and read have been re-added

PR #2431, opened 2018-09-06
Remove another non-restricted right from MW blacklist

Amanda-Catherine:
This right is granted to all logged in/registered users by default, and therefore does not appear to be sensitive/restricted


Reception123:
It is restricted because bureaucrats shouldn't be allowed to take that away from users, not because it's sensitive.

PR #2438, opened 2018-09-07
Add composer.json to run phplint test

paladox:
Instead of php inbuilt php lint lets use php-parallel-lint


JohnFLewis:
What does this gain?

paladox:
answered on irc

PR #2450, opened 2018-09-09
Trying something different per MacFan4000 on IRC

Amanda-Catherine:


LocalSettings.php
```diff
@@ -2593,13 +2593,6 @@
 				'oauth-enable' => true,
 				'managewiki' => true,
 				'managewiki-restricted' => true,
-			),
```
MacFan4000: the build fails because you removed this

Amanda-Catherine: Oh pfft

PR #2454, opened 2018-09-12
disable wgStructuredChangeFiltersShowPreference on apunteswiki, centralwiki, destinoswiki, mexicopediawiki, repositoriowiki, ucroniawiki

Wiki1776:
I also remove unionwiki, tallerwiki and unidoswiki configurations because they were deleted.


LocalSettings.php
```diff
@@ -2782,28 +2785,6 @@
 			NS_EXPLANATION => 'Explanation',
 			NS_EXPLANATION_TALK => 'Explanation_talk',
 		),
-		'tallerwiki' => array(
```
paladox: These look like unrelated changes?

PR #2458, opened 2018-09-15
Switch createwiki, matomo and centralauth to use mhglobal as db

paladox:


LocalSettings.php
```diff
@@ -1590,7 +1590,7 @@
 		'weatherwiki' => false, // let me do the blocking on my wiki, please
 	),
 	'wgGlobalBlockingDatabase' => array(
-		'default' => 'centralauth', // use centralauth for global blocks
+		'default' => 'mhglobal', // use centralauth for global blocks
```
JohnFLewis: comment needs to be changed!

PR #2462, opened 2018-09-17
Request to un-blacklist some user rights.

bonnedav:
skipcaptcha - So I can make sub-admin groups that can bypass the captcha.
read - So I can make sub-admin groups able to read if I make the wiki private temporarily.
torunblocked - So I can make sub-admin and exception groups able to bypass tor blocking.
mwoauthviewprivate - So admins can manage and approve consumers.
mwoauthmanageconsumer - So admins can manage and approve consumers.
bigdelete - So admin can delete pages as necessary.
viewpmlog - extension MediaWikiChat is not it manage wiki (maby add it?), and Most chat systems let admins see pms.
mwoauthmanagemygrants - User has it.


JohnFLewis:
>  skipcaptcha - So I can make sub-admin groups that can bypass the captcha.
The reason this is restricted is so that people can't grant this right to * and allow complete evasion of our Captcha.
> read - So I can make sub-admin groups able to read if I make the wiki private temporarily.
The reason this is restricted is the only way to change read, except for doing it in LS, should be making a private wiki. Also note, when the wiki is private, a member group exists anyway.
> torunblocked - So I can make sub-admin and exception groups able to bypass tor blocking.
Allows evasion of TOR blocks, so shouldn't be assignable to *
> mwoauthviewprivate - So admins can manage and approve consumers.
OAuth management should happen on meta only.
> mwoauthmanageconsumer - So admins can manage and approve consumers.
OAuth management should happen on meta only.
> bigdelete - So admin can delete pages as necessary.
For technical reasons, so someone can't overload the database and bring the farm down by doing excessive history deletion.
> viewpmlog - extension MediaWikiChat is not even installed.
It is.
> mwoauthmanagemygrants - User has it.
Should always be a right User has, shouldn't be revokable.

You can modify any of these rights using $wgManageWikiPermissionsAdditionalRights.

bonnedav:
ok. maybe this shows the need to change managewiki to have a blacklist for *? This would be on top of the exsiting blacklist ad block the listed rights from being added to *.

bonnedav:
I changed viewpmlog after I saw the extention was installed jut not listed in managewiki. Why is it restricted? @JohnFLewis

JohnFLewis:
It doesn't really. The implementation of a per user right blacklist would be far to complicated and not address every issue possible.

These rights are blacklisted not only for technical reasons for policy and social reasons and the implementation of a new blacklist wouldn't remove these from this list realistically except perhaps skipcaptcha and torunblocked which are two incredibly niche rights that are rarely necessary and if ever used esp since they are assigned as appropriate to the user groups which would need them.

MediaWikiChat isn't on there because no one has added it.

bonnedav:
the special blacklist would only exist for * and would mainly be for skipcaptcha and torunblocked.

bonnedav:
Thank you for you input and explanations, it is much appreciated. @JohnFLewis

PR #2463, opened 2018-09-17
Add extension MediaWikiChat to ManageWiki

bonnedav:


MacFan4000:
This doesnt need to be in ManageWiki, the permissions won’t show up unless the extension is enabled on the given wiki.

MacFan4000:
See https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L1044

JohnFLewis:
"This doesnt need to be in ManageWiki" why?

MacFan4000:
The reason is whatever the reason was that you didn’t add it @JohnFLewis

JohnFLewis:
The reason is I didn't.

bonnedav:
Why not just add it?

JohnFLewis:
It should be.

PR #2467, opened 2018-09-18
Update LocalSettings.php

paladox:


LocalSettings.php
```diff
@@ -2096,13 +2096,16 @@
 	),
 	'wgManageWikiPermissionsBlacklistRights' => array(
 		'default' => array(
+			// arrays
```
JohnFLewis: why is it necessary to add a comment? there's no comment on the other probably hundred of arrays in this file

LocalSettings.php
```diff
@@ -2096,49 +2096,55 @@
...
+				'hideuser',
+				'interwiki',
+				'managewiki-restricted',
+				'managewiki-editdefault',
+				'mwoauthmanageconsumer',
+				'mwoauthmanagemygrants',
+				'mwoauthsuppress',
+				'mwoauthviewprivate',
+				'mwoauthviewsuppressed',
+				'read',
```
MacFan4000: Remove read from any

LocalSettings.php
paladox: I've removed it now.

LocalSettings.php
```diff
@@ -2096,49 +2096,55 @@
...
+				'renameuser',
+				'requestwiki',
+				'siteadmin',
+				'skipcaptcha',
```
MacFan4000: This too

LocalSettings.php
```diff
@@ -2096,49 +2096,55 @@
...
+				'suppressionlog',
+				'suppressrevision',
+				'torunblocked',
```
MacFan4000: Same here

PR #2469, opened 2018-09-18
Localize extensions that are not absolutely required

Amanda-Catherine:
To-do list:

- [ ] Remove extensions from GlobalExtensions.php
- [ ] Add extensions to LocalExtensions.php and LocalSettings.php
- [ ] Enable ManageWikiExtensions for localized extensions
- [ ] Travis CI Check passes after all commits have been added
- [ ] Manual syntax/code review by Amanda
- [ ] Approval from sysadmins/stewards from a non-technical perspective
- [ ] Manual syntax/code review by one or more sysadmins
- [ ] Merge PR


paladox:
How do you know these are not needed?

GlobalExtensions.php
```diff
@@ -4,43 +4,29 @@
...
 	'Echo',
-	'Gadgets',
 	'GlobalBlocking',
 	'GlobalCssJs',
 	'GlobalPreferences',
 	'GlobalUserPage',
-	'InputBox',
 	'Interwiki',
 	'LocalisationUpdate',
-	'LoginNotify',
```
JohnFLewis: I'd like to keep this enforced as it's a security-related extension.

Amanda-Catherine:
@paladox they’re not required for legal reasons nor are they required for things to work technically. They’re just useful features that some users/wikis may choose to utilize, but should not be forced to utilize them.

GlobalExtensions.php
```diff
@@ -4,43 +4,29 @@
 wfLoadExtensions( [
 	'AbuseFilter',
 	'AntiSpoof',
-	'Babel',
-	'BetaFeatures',
```
JohnFLewis: I'm not sure if we use BF globally in a way (I think we used to, unsure if we do now?)

paladox:
You will need to notify all wikis are this change Amanda. Because some wikis have requested extensions already globally installed.

Amanda-Catherine:
@paladox well, I don’t have acess to CentralNotice, but I guess we could do a global sitenotice. Also, I don’t think any of these extensions have been requested in recent Phabricator tasks

JohnFLewis:
These would just become opt-out.

There would be no change except wikis can disable them in ManageWiki should be the end goal.

JohnFLewis:
and new wikis can enable them if they want them (since we'd only switch them on by default for existing)

JohnFLewis:
The rest I don't have an issue with as long as they are enabled on all existing wikis by default. New wikis get the choice and existing wikis can disable what they don't want. That is, after all, the point of ManageWiki.

NDKilla:
I feel like before merge part of this pull request needs to be setting up defaults for existing wikis and new wikis and also making sure this can be controlled in managewiki

JohnFLewis:
Needs to be added to LocalExtensions.php, LocalSettings.php, ManageWiki.php then needs to be enabled in the DB by whoever merges this.

paladox:
You need to add them to ManageWiki.php.

Amanda-Catherine:
Argh... I screwed something up while trying to do the next part of this. Please stand by.

PR #2485, opened 2018-09-24
Add bigdelete to weatherwiki stewards

Amanda-Catherine:
Another “just in case” type of thing that shouldn’t be completely unassigned


MacFan4000:
This right is already granted to stewards globally.

MacFan4000:
Test comment please ignore

PR #2489, opened 2018-09-30
Update $wmgClosedWiki permissions

Amanda-Catherine:
Even on closed wikis, existing admins/bureaucrats and/or stewards may need to be able to block/unblock and delete/undelete.


JohnFLewis:
Stewards/sysadmins aren’t affected by this.

The purpose of a closed wiki is anyone with local rights can’t do any RC logable actions. If they need to, then they should reopen the wiki or request whole wiki deletion.

PR #2493, opened 2018-10-01
Enable flag abusefilter action

bonnedav:
You have the rest enabled, why not?


bonnedav:
Seems it was removed. My bad. Seems they outta update https://www.mediawiki.org/wiki/Extension:AbuseFilter#Configuration then. Why was it removed I wonder?

PR #2501, opened 2018-10-02
Add .editorconfig

revi:
This will tell GitHub editor and various other editors (but no vim or nano) to use TAB instead of space.

And other practices covered as well. Correct me if the standards have changed meanwhile.


paladox:
LGTM

PR #2514, opened 2018-10-09
Add parameter namespace for ecoepiwiki T3691

Amanda-Catherine:


LocalSettings.php
```diff
@@ -3112,6 +3118,9 @@
 			'DC' => NS_MAIN,
 			'DC_talk' => NS_TALK,
 		),
+		'+ecoepiwiki' => array(
+			'P' => NS_PARAMETER,
```
Reception123: Let's keep it consistent and have a "talk" here as well

Amanda-Catherine: @Reception123 but they didn’t say that they wanted a talk alias, let alone what the alias key should be. I don’t want to assume things and have them come running back complaining later.

mlange-42: @Amanda-Catherine @Reception123 
It's fine with the associated talk namespace, as requested. I edited my request 2h after you adopted the task, please excuse.
Many thanks for the fast handling!

PR #2532, opened 2018-10-20
Move wgVerifyMimeType to ManageWiki

paladox:


JohnFLewis:
Tbh I’d argue this is always true

There is no reason I’m aware of that this should be false.

PR #2544, opened 2018-11-04
Use wmgSiteNoticeOptOut in CN

paladox:


JohnFLewis:
I'm confused?

Optout and important sound the same. The point of optout is you only get important ones

LocalSettings.php
```diff
@@ -455,14 +455,15 @@
 	'wgCentralHost' => array(
 		'default' => "//meta.miraheze.org",
 	),
-	// 'wiki' => 'wiki'
+	// 'wiki' => optout or important
```
JohnFLewis: also there shouldn't be anything here besides "default"

PR #2545, opened 2018-11-04
Move wmgSiteNoticeOptOut to ManageWiki

paladox:


ManageWiki.php
```diff
@@ -1615,11 +1615,19 @@
...
-		'requires' => 'Popups',
+		'requires' => 'popups',
 		'restricted' => false,
 		'type' => 'check',
 		'overridedefault' => null,
 		'section' => 'media',
 		'help' => false,
 	),
+	'wmgSiteNoticeOptOut' => array(
+		'name' => 'Opt out of site notices (only show important notices)',
```
JohnFLewis: should clarify this is Miraheze global ones

PR #2551, opened 2018-11-09
Add techdict namespace to arphilosophywiki T3780

Amanda-Catherine:


LocalSettings.php
```diff
@@ -2836,6 +2836,10 @@
 			NS_FEATURED => 'Featured_Page',
 			NS_FEATURED_TALK => 'Featured_Page_talk',
 		),
+		'arphilosophywiki' => array(
```
paladox: Need to add this to the define at the top too.

Amanda-Catherine: @paladox ?

paladox: For example "define( 'NS_TECH', 1600 );"

Amanda-Catherine: Oh pfft I knew that... *slaps face with trout* (dealing with some personal issues right now so I’m not at my best)

PR #2560, opened 2018-11-20
Enable moderation on socioatlaswiki T3806

Amanda-Catherine:


LocalSettings.php
```diff
@@ -2162,6 +2163,11 @@
 				'read' => true,
 			),
 		),
+		'+socioatlaswiki' => array(
+			'sysop' => array(
+				'moderation' => true,
```
JohnFLewis: moderation isn't a restricted right

Amanda-Catherine: @JohnFLewis ok, so why does next to wgEnableModeration it says “don’t forget to assign ‘moderation’ rights!”

MacFan4000: @Amanda-Catherine because it was installed before ManageWikiPermissions was a thing, and the comment was never removed

PR #2571, opened 2018-12-04
set min password legth

MacFan4000:


paladox:
Looks correct to me.

https://www.mediawiki.org/wiki/Manual:$wgPasswordPolicy

PR #2589, opened 2019-01-06
Make MobileDetect available as a global extension

penguinstyles:
Per Phabricator task [T3948](https://phabricator.miraheze.org/T3948)


extension-list
```diff
@@ -95,6 +95,7 @@ $IP/extensions/MassMessage/extension.json
 $IP/extensions/Math/extension.json
 $IP/extensions/MediaWikiChat/extension.json
 $IP/extensions/MirahezeMagic/extension.json
+$IP/extensions/MobileDetect/MobileDetect.php
```
JohnFLewis: Why not extension.json?

penguinstyles: The extension has not been updated to support extension.json yet.

PR #2595, opened 2019-01-09
Allow Genealogy to be used as a local extension (T3881)

penguinstyles:
Install Genealogy per T3881


LocalExtensions.php
```diff
@@ -313,6 +313,10 @@
 	$wgManageWikiSettings['wgDefaultSkin']['options']['Gamepress'] = 'gamepress';
 }
 
+if ( $wmgUseGenealogy ) {
```
paladox: Need's to be set in LS.php too (default to false) and also needs adding to ManageWiki.php

PR #2604, opened 2019-01-14
Enable ContactPage on qboxnextwiki T3993

Amanda-Catherine:


ContactPage.php
```diff
@@ -75,3 +75,17 @@
 		'RLStyleModules' => array(),
 	);
 }
+
+if ( $wgDBName == 'qboxnextwiki' ( {
+	$wgContactConfig['default'] = array
```
MacFan4000: missing (

ContactPage.php
```diff
@@ -75,3 +75,17 @@
 		'RLStyleModules' => array(),
 	);
 }
+
+if ( $wgDBName == 'qboxnextwiki' ( {
```
MacFan4000: ( should be )

PR #2605, opened 2019-01-15
Set $wgEnableCanonicalServerLink true for electowikiwiki T4007

Amanda-Catherine:


Amanda-Catherine:
> Can we please stop adding stuff to LS over ManageWiki please? cc @miraheze/operations @miraheze/mediawiki-admins

So where was I supposed to put this where the relevant variable wasn’t defined anywhere yet?

PR #2611, opened 2019-01-19
Disable “page creation log” for weatherwiki

Amanda-Catherine:
Unnecessary and unwanted


JohnFLewis:
Sounds like my polite request last time went unheeded. @paladox please don’t add stuff to LS anymore if they can go into ManageWiki. Thanks

paladox:
@JohnFLewis sorry. I've made it right now!

Can we add a note at the top of LS that says if the config can be in ManageWiki please add it to ManageWiki.php?

That message will be intended for users not staff.

PR #2613, opened 2019-01-20
fix pdfbook

The-Voidwalker:


paladox:
Needs fixing in LS.php too. But maybe just fix it in MW to be wmgUsePdfBook?

The-Voidwalker:
oh, heh

PR #2632, opened 2019-02-18
Add WikibaseClient and enable it for ffgxvwiki

paladox:


LocalSettings.php
```diff
@@ -1354,6 +1354,11 @@
 	'wmgUseWikibaseRepository' => [
 		'default' => false,
 	],
+	// Do not put in ManageWiki
```
JohnFLewis: Why?

paladox: Because of how it's configured see https://www.mediawiki.org/wiki/Wikibase/Installation/Advanced_configuration#Wikibase_Client

PR #2633, opened 2019-02-21
Update ManageWiki to new config format

paladox:


ManageWiki.php
```diff
@@ -224,24 +213,25 @@
 			'linkPage' => 'https://mediawiki.org/wiki/Extension:AutoCreateCategoryPages',
 			'var' => 'wmgUseAutoCreateCategoryPages',
 			'conflicts' => false,
-			'requires' => false,
-			'restricted' => true,
```
JohnFLewis: restricted

ManageWiki.php
```diff
@@ -286,72 +274,72 @@
...
-			'restricted' => false,
+			'requires' => [],
 		],
 		'collection' => [
 			'name' => 'Collection (PDF)',
 			'linkPage' => 'https://mediawiki.org/wiki/Extension:Collection',
 			'var' => 'wmgUseCollection',
 			'conflicts' => false,
-			'requires' => 'electronpdfservice',
-			'restricted' => true,
```
JohnFLewis: was this just restricted for the ext requirement?

ManageWiki.php
paladox: Thanks, done!

ManageWiki.php
paladox: Ah yes, i think so. This should now work (so no longer needs restricting).

PR #2635, opened 2019-02-21
ManageWiki: Add help comments (from ManageWiki#Overview (meta))

paladox:


ManageWiki.php
```diff
@@ -2099,7 +2099,7 @@
 		'type' => 'check',
 		'overridedefault' => null,
 		'section' => 'media',
-		'help' => false,
+		'help' => 'This enables Popups as a beta feature, rather then showing it to all users.',
```
JohnFLewis: than

ManageWiki.php
```diff
@@ -2115,15 +2115,15 @@
 		'restricted' => false,
 		'type' => 'check',
 		'overridedefault' => false,
-		'help' => '',
+		'help' => 'This setting allows anonymous users to change the language for them selfs.',
```
JohnFLewis: "Enabling allowing anonymous users to control the language they view the wiki in"

paladox: Thanks, done!

PR #2662, opened 2019-04-10
Move wgAutoConfirmAge to ManageWiki

paladox:


JohnFLewis:
Same as other message

PR #2680, opened 2019-05-16
Fix weatherwiki user permissions

Amanda-Catherine:


Amanda-Catherine:
@JohnFLewis for some reason, I can add the banned group to user accounts (i.e. ban users who aren't banned) but I can't remove the banned group (i.e. can't unban users who are currently banned). See the screenshot below of an LTA account that I've banned - not that I would want to unban them, but I'm concerned that I don't have the ability to do so. I can't find any typos in the commit that would be causing this?

PR #2683, opened 2019-06-13
Include "--logs --uploads" in the DataDump dumps

paladox:
This doesn't include the actual file so no copyright concerns. It only includes the log.


Reception123:
What is this waiting on?

PR #2692, opened 2019-07-03
Enable partial blocks on all wikis by default

Amanda-Catherine:


LocalSettings.php
```diff
@@ -108,6 +108,12 @@
 	'wgEnableBlockNoticeStats' => [
 		'default' => false,
 	],
+	'wgEnablePartialBlocks' => [
+		'default' => false,
```
paladox: Set this to true please and remove the wikis below.

We’re going to enable this for all wikis.

PR #2707, opened 2019-07-30
Remove Extension:Poll

RhinosF1:
T4584


RhinosF1:
This should be clear to remove now if no sysadmins have any issues

RhinosF1:
Please let me know when this gets pushed so I can try to be around (even if it is a basic change) as I like the idea of being able to say that I'm busy doing a deploy and testing it. Also see https://meta.miraheze.org/wiki/Community_noticeboard#Extension_Changes for community notice

PR #2713, opened 2019-08-01
database: Add db5 and set wgCreateWikiDatabaseClusters

paladox:


LocalSettings.php
```diff
@@ -260,6 +260,10 @@
 	 ],
 
 	// CreateWiki
+	'wgCreateWikiDatabaseClusters' => [
+		'db4',
```
Southparkfan: db4 and db5 are not clusters, c1 and c2 are

Database.php
```diff
@@ -28,8 +28,8 @@
 	'hostsByName' => [
 		// 81.4.109.166
 		'db4' => 'mediawiki-internal-db-master.miraheze.org',
-		// 168.235.110.50
-//		'db5' => 'mediawiki-internal-db-master-db5.miraheze.org',
+		// 185.52.1.89
+		'db5' => 'mediawiki-internal-db-master-db5.miraheze.org',
 	],
 	'externalLoads' => [
```
Southparkfan: Shouldn't db5 be uncommented as well? (not sure!)

paladox: In externalLoads no, because that means it'll balance between db4 and db5 (meaning a replica i think).

LocalSettings.php
paladox: Thanks, done.

JohnFLewis:
Please don’t merge. It’s untested code, incomplete code and won’t work code re CreateWiki stuff

Southparkfan:
> Please don’t merge. It’s untested code, incomplete code and won’t work code re CreateWiki stuff

We are aware (discussed on IRC). Needs support for auto populating sectionsByDB

PR #2714, opened 2019-08-01
Remove MobileFrontendAutoDetect

RhinosF1:
per IRC
@Reception123 / @paladox


paladox:
We already do this through varnish so we don't really need this config. Also this config doesn't really work as it appears to be defaulting to mobile for desktop views.

PR #2715, opened 2019-08-07
Unset cargo db variable

paladox:
This is uneeded as it uses the wiki dbname rather then a shared db and duplicates things.


NDKilla:
+1

I'm not entirely sure why this didn't work before, I think we may have manually set it to 'cargodb' or similar but I can't find the commit where this was updated/added initially.

A quick check of the extension confirms that these values default to null, and if they're null they're replaced with $wgDBname etc.

paladox:
Thanks @NDKilla for reviewing! I'll merge in a bit.

PR #2716, opened 2019-08-08
remove weatherwiki config

Amanda-Catherine:
As I don't have time or the motivation to continue maintaining it, Weather Wiki is going to be deleted within the coming days. This change removes all configuration that I could find in LocalSettings.php pertaining to weather wiki - most of it is "leftover" things that couldn't go in ManageWiki. However, DO NOT MERGE this change right now as I'm still working with DataDump and figuring out how it works, and as such the wiki should NOT be deleted yet.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2819,7 +2819,7 @@ $wgConf->settings = [
 			'random_quiz' => false,
 		],
 	],
-];
+],
 
 function efGetSiteParams( $conf, $wiki ) {
 	$site = null;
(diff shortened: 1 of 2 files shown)


RhinosF1:
Amanda, this is still failing - please see your repo's PR's for a fix

LocalSettings.php
```diff
@@ -1870,28 +1853,6 @@
 		'+whentheycrywiki' => [
 			'user' => [
 				'edit-create' => true,
-			],
```
The-Voidwalker: This line should not have been removed

LocalSettings.php
```diff
@@ -2879,7 +2819,7 @@
 			'random_quiz' => false,
 		],
 	],
-];
+],
```
The-Voidwalker: This should remain a semicolon

PR #2717, opened 2019-08-08
[DO NOT MERGE] remove weatherwiki config

Amanda-Catherine:
As I don't have time or the motivation to continue maintaining it, Weather Wiki is going to be deleted within the coming days. This change removes all configuration that I could find in LocalSettings.php pertaining to weather wiki - most of it is "leftover" things that couldn't go in ManageWiki. However, DO NOT MERGE this change right now as I'm still working with DataDump and figuring out how it works, and as such the wiki should NOT be deleted yet.


Amanda-Catherine:
Oops, duplicate

PR #2720, opened 2019-08-10
add multiboilerplate to sidemwiki

amehiko:


paladox:
You can enable the extension at https://sidem.miraheze.org/wiki/Special:ManageWiki/extensions then you can enable the setting at https://sidem.miraheze.org/wiki/Special:ManageWiki/settings#mw-section-styling

PR #2731, opened 2019-08-16
move MyVariables to prod

RhinosF1:


RhinosF1:
+1 from me - works fine, no caching observed at all

PR #2732, opened 2019-08-18
enable db5 wikis and add testwiki to db5

Reception123:
not sure why last time Paladox tried to add db5 it caused errors so waiting for approval from @JohnFLewis and/or @paladox before merging.


Reception123:
ok given by Paladox via IRC

Reception123:
Seems to be causing "(Cannot access the database: Cannot access the database: Unknown error (mediawiki-internal-db-master-db5.miraheze.org))"

any ideas why @JohnFLewis ?

PR #2737, opened 2019-08-23
make NS_MAIN index on horizonwiki

RhinosF1:
per T4669


RhinosF1:
Whoever merges this, please change the default to noindex in ManageWiki

PR #2744, opened 2019-08-25
theme: Only add wgDefaultTheme for vector and monobook.

paladox:


ManageWikiSettings.php
```diff
@@ -1370,8 +1367,12 @@
 		'from' => 'theme',
 		'restricted' => false,
 		'type' => 'list',
-		'options' => $skinOptions + [
+		'options' => [
 			'Default' => 'default',
+			'Dark (Monobook and Vector only)' => 'dark',
+			'Deepsea (Vector only)' => 'deepsea',
+			'Ping (Monobook only)' => 'pink',
```
RhinosF1: Should be pink not ping?

paladox: done

PR #2750, opened 2019-09-03
per request T4694

RhinosF1:


LocalSettings.php
```diff
@@ -1117,6 +1117,16 @@
 	'wgAllowImageTag' => [
 		'default' => false,
 	],
+	'egApprovedRevsEnabledNamespaces' => [
```
paladox: Need to remove the '' from the constants.

PR #2752, opened 2019-09-07
add .naf per T4702

RhinosF1:


Reception123:
Do not merge, waiting on Phab response

PR #2759, opened 2019-09-15
move 3 extensions to prod

RhinosF1:
+1 from me on all 3

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -426,8 +426,7 @@ $wgConf->settings = [
 	],
 	//CommonsMetadata
 	'wgCommonsMetadataForceRecalculate' => [
-		'default' => false,
-		'test1wiki' => true,
+		'default' => true,
 	],
 
 	// Delete
@@ -503,7 +502,6 @@ $wgConf->settings = [
 	],
 	'wmgUseAdvancedSearch' => [
 		'default' => false,
-		'test1wiki' => true,
 	],
 	'wmgUseAJAXPoll' => [
 		'default' => false,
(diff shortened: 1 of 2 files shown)


LocalSettings.php
```diff
@@ -426,8 +426,7 @@
 	],
 	//CommonsMetadata
 	'wgCommonsMetadataForceRecalculate' => [
-		'default' => false,
-		'test1wiki' => true,
+		'default' => true,
```
paladox: Why is this being enabled by default?

ManageWikiExtensions.php
```diff
@@ -1665,12 +1657,7 @@
 			'linkPage' => 'https://www.mediawiki.org/wiki/Extension:TextExtracts',
 			'var' => 'wmgUseTextExtracts',
 			'conflicts' => false,
-			'requires' => [
-				'permissions' => [
-					'managewiki-restricted',
-				],
-			],
-		],
+			'requires' => [],
```
paladox: Syntax error

LocalSettings.php
RhinosF1: As it allows remote files (from miraheze/wikimedia) commons to be used if enabled for this extension. As far as I can tell, those images wouldn't work without it.

paladox: This should be made configurable in MW and disabled by default.

Based on me reading https://www.mediawiki.org/wiki/Extension:CommonsMetadata/en there may be a performance hit with this enabled. Secondly this you can already use remote files.

What the extension does is it pulls meta data.

RhinosF1:
Reverting while I fix i18n to avoid issues

PR #2765, opened 2019-09-21
Convert mediawiki.org links to use Special:MyLanguage

RhinosF1:
Requested at https://meta.miraheze.org/w/index.php?title=Stewards%27_noticeboard&curid=662&diff=83574&oldid=83573


RhinosF1:
Will possibly automate the rest later

PR #2766, opened 2019-09-21
Grant bureaucrats renameuser

GustaveLondon776:


Pix1234:
Renames are handled by Stewards as they have to be done globally.

RhinosF1:
Per @Pix1234

PR #2770, opened 2019-10-12
Allow cvt block the email

GustaveLondon776:


RhinosF1:
Should be done by stewards on wiki following an RfC

PR #2776, opened 2019-10-16
Allow Steward the right of the block

GustaveLondon776:


JohnFLewis:
Please do not create pull requests which provide no constructive value. Matters regarding permissions and such should be decided by a vote of the community.

If you continue, your ability to produce pull requests to this repo may be revoked.

Thank you.

PR #2791, opened 2019-11-01
Make DarkMode on by default

RhinosF1:


JohnFLewis:
This makes it a global extension, in which case it should be done via GlobalExtensions not LE/LS

RhinosF1:
> This makes it a global extension, in which case it should be done via GlobalExtensions not LE/LS

Try now

PR #2795, opened 2019-11-07
Update ManageWikiSettings.php

kenng69:
I want to add more file extensions to the Restricted File Extensions settings


kenng69:
Invalid syntax. I will open one with correct syntax later.

PR #2824, opened 2019-12-13
Adjust protection settings

PlavorSeol:
* It makes sense to sort protection levels by hierarchy, lower to higher like `""`->`autoconfirmed`(->`extendedconfirmed`)->`sysop`, as MediaWiki core (and Wikimedia wikis in case of using `extendedconfirmed` protection level) does. As such, I adjusted [`$wgRestrictionLevels`](https://www.mediawiki.org/wiki/Manual:$wgRestrictionLevels) to display `user` protection level ("Allow only logged in users") before `autoconfirmed` and `sysop` protection levels.
* I also reordered `'create'` in [`$wgRestrictionTypes`](https://www.mediawiki.org/wiki/Manual:$wgRestrictionTypes) to be before other options since it matches better to default settings and "creating" a page is done before "editing", and enabled protecting pages from changing protection settings by default.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2305,9 +2305,11 @@ $wgConf->settings = [
 	],
 
 	// Restriction types
-	'+wgRestrictionLevels' => [
+	'wgRestrictionLevels' => [
 		'default' => [
 			'user',
+			'autoconfirmed',
+			'sysop'
 		],
 		'+bigforestwiki' => [
 			'editvoter',
@@ -2399,11 +2401,12 @@ $wgConf->settings = [
 	],
 	'wgRestrictionTypes' => [
 		'default' => [
+			'create',
 			'edit',
 			'move',
-			'create',
 			'upload',
 			'delete',
+			'protect'
 		],
 	],
 


LocalSettings.php
```diff
@@ -2305,9 +2305,11 @@
 	],
 
 	// Restriction types
-	'+wgRestrictionLevels' => [
+	'wgRestrictionLevels' => [
```
paladox: Needs to stay +wgRestrictionLevels otherwise all wikis who have been specified here will find breakages.

(+ means that it'll use the default + what ever the wiki specifies).

PlavorSeol: As far as I know, using `+` in front of configuration variable means merging settings with `DefaultSettings.php`, but in order to put `user` before `autoconfirmed` and `sysop`, the core default value must be overridden since the core default value of it is `["","autoconfirmed","sysop"]`.
Merging of wiki-specific additional level and default settings is done by using `+` in front of **wiki database name** (not variable), so for example value of `$wgRestrictionLevels` for [큰숲백과](https://bigforest.miraheze.org/) (`bigforestwiki`) will be `["","user","autoconfirmed","sysop","editvoter"]`.
Please review this again, and if it still needs `+`, I will push a new commit to prepend it.

Pix1234: Paladox is correct you must add + afaik

PlavorSeol: Well, won't the `default` value of `$wgRestrictionLevels` be weird like `['','autoconfirmed','sysop','','user','autoconfirmed','sysop']` if I add `+`?

Reception123:
@paladox

paladox:
This seems fine now, sorry for reviewing late.

PR #2831, opened 2019-12-23
Add $wgMaxCredits on ManageWiki per request T5011

Hispano76:
It needs thorough revision! It may have errors.


ManageWikiSettings.php
```diff
@@ -348,6 +348,17 @@
 		'section' => 'edit',
 		'help' => 'The number of days within which a person must make edits to be considered an "active" user.',
 	],
+	'wgMaxCredits' => [
+		'name' => 'Editing attribution',
+		'from' => 'mediawiki',
+		'restricted' => false,
+		'type' => 'integer',
+		'minint' => -1,
+		'maxint' => 0,
```
The-Voidwalker: This should probably be higher

RhinosF1: See https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgMaxCredits

This allows for all or none but probably and also name probably should be ‘Number of Editors to Attribute in footer’?

Hispano76:
> Can we add https://www.mediawiki.org/wiki/Manual:$wgShowCreditsIfMax at the same time?
check

Hispano76:
> Can we add https://www.mediawiki.org/wiki/Manual:$wgShowCreditsIfMax at the same time?

check

PF94:
i littearly made a phabricaptor request and now i'm semi-responsibale for the birth of a pull request :P

ManageWikiSettings.php
PF94: my mistake, i think the limit should be set to say like, 25?

PR #2847, opened 2020-01-04
wgMinervaEnableSiteNotice  on ManageWiki per T5057

Hispano76:
It needs thorough revision! It may have errors.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2081,7 +2081,7 @@ $wgConf->settings = [
 	],
 
 	'wgMinervaEnableSiteNotice' => [
-		'default' => true,
+		'default' => false,
 	],
 
 	// miraheze specific config
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2080,6 +2080,10 @@ $wgConf->settings = [
 		],
 	],
 
+	'wgMinervaEnableSiteNotice' => [
+		'default' => true,
+	],
+
 	// miraheze specific config
 	'wgServicesRepo' => [
 		'default' => '/srv/services/services',
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -1340,7 +1340,7 @@ $wgManageWikiSettings = [
 	],
 	'wgMinervaEnableSiteNotice ' => [
 		'name' => 'Minerva Sitenotice',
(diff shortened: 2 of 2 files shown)


paladox:
Need to also set wgMinervaEnableSiteNotice in LocalSettings

LocalSettings.php
```diff
@@ -2080,6 +2080,10 @@
 		],
 	],
 
+	'wgMinervaEnableSiteNotice' => [
+		'default' => true,
```
paladox: Should be set to false

ManageWikiSettings.php
```diff
@@ -1338,6 +1338,15 @@
 		'section' => 'styling',
 		'help' => 'If not set, then the img.line element (white divider line in the top menu, between the site name and the down arrow/"Upload file" link) won\'t be generated.',
 	],
+	'wgMinervaEnableSiteNotice ' => [
+		'name' => 'Minerva Sitenotice',
+		'from' => 'minerva',
```
paladox: mobilefrontend

LocalSettings.php
Hispano76: check

ManageWikiSettings.php
Hispano76: check

ManageWikiSettings.php
```diff
@@ -1338,6 +1338,15 @@
 		'section' => 'styling',
 		'help' => 'If not set, then the img.line element (white divider line in the top menu, between the site name and the down arrow/"Upload file" link) won\'t be generated.',
 	],
+	'wgMinervaEnableSiteNotice ' => [
+		'name' => 'Minerva Sitenotice',
+		'from' => 'mobilefrontend',
+		'restricted' => false,
+		'type' => 'check',
+		'overridedefault' => true,
```
Hispano76: Um, that should be in false, right? 'Cause you check the box in ManageWiki.

paladox: Nope, this makes sure it is true by default, allowing users to unset it.

If you set this as false it'll mean it defaults to false.

Hispano76: Okay, I wanted to make sure. Thank you.

PR #2851, opened 2020-01-05
Minerva Sitenotice choice in correct settings section per T5067

Hispano76:


ManageWikiSettings.php
```diff
@@ -1344,7 +1344,7 @@
 		'restricted' => false,
```
paladox: Need to move this config to the correct section.

PR #2852, opened 2020-01-05
Enable electron if collection is enabled

paladox:
Collection already does this but we were duplicating things by allowing users to enable it separately.


paladox:
> Needs wfLoadExtension to be added to the collection config not just removed

Did you not look at LE? It already does that :)

PR #2853, opened 2020-01-07
enable wgExtraInterlanguageLinkPrefixes on hispanowiki and Ucroniaswiki per request

Hispano76:


LocalSettings.php
```diff
@@ -1423,7 +1423,21 @@
...
+		'+hispanowiki' => [
+			'w',
+			'v',
+			'n',
+			'u',
+			'd',
+			'c',
+			'commons',
+		],
 	],
```
The-Voidwalker: This needs to go after the ucroniaswiki block

Hispano76: done

PR #2858, opened 2020-01-15
Enable wgMiserMode

paladox:
Per documentation https://www.mediawiki.org/wiki/Manual:$wgMiserMode, this is designed to reduce load (for wiki farms or large wikis).

Switching this on means that the query cache is now used and so we have to run a cron on all wikis (which will be done every 3 months).


paladox:
Added cron here https://github.com/miraheze/puppet/pull/1187

PR #2866, opened 2020-01-22
Make WikiForum disable for anon user on allthetropeswiki per T5140

HeartsDo-Dev:


paladox:
ATT can disabled this using ManageWiki. (under the edit section).

PR #2888, opened 2020-02-17
Add GPLv3 to WikiLicenses

RhinosF1:
Fixes T5202


RhinosF1:
Please check + fix syntax as I’m mobile

LocalWiki.php
```diff
@@ -265,6 +265,11 @@ function wfModifyMetaTags( OutputPage $out ) {
 		break;
 	case 'empty':
```
JohnFLewis: Empty should always be the bottom for stylistic reasons (as it's essentially blank, a terminating one)

ManageWikiSettings.php
```diff
@@ -323,6 +323,7 @@
 			'Creative Commons BY-NC-ND 4.0' => 'cc-by-nc-nd',
 			'Public Domain' => 'cc-pd',
 			'No license provided' => 'empty',
+                        'GNU General Public V3' => 'gpl-v3',
```
JohnFLewis: Same as above

PR #2894, opened 2020-02-20
add $wgExpensiveParserFunctionLimit

RhinosF1:
Default = 99 per mw org
Metawiki = 150 due to some pages needing more


ManageWikiSettings.php
```diff
@@ -1261,6 +1261,17 @@
 		'section' => 'restricted',
 		'help' => 'This is the list of preferred extensions for uploading files. Uploading files with extensions not selected in this list will trigger a warning.',
 	],
+	'wgExpensiveParserFunctionLimit' => [
+		'name' => 'Expensive Parser Function Limit',
+		'restricted' => true,
+		'from' => 'scribunto',
```
paladox: This is a core config so should be from mediawiki.

PR #2902, opened 2020-02-23
restore access to editothersprofiles

The-Voidwalker:
this right no longer provides access to private information since the introduction of editothersprofiles-private


paladox:
Done in https://github.com/wikimedia/mediawiki-extensions-SocialProfile/commit/9420e596fa991065a96f0ee52fe966ab99724485

PR #3004, opened 2020-04-12
Remove wgServer from ManageWikiSettings

paladox:


akuckartz:
Why?

Pix1234:
@akuckartz we changed the way custom domains (this setting) is handled

PR #3032, opened 2020-04-23
use $wgServer as $wmgHostname is null

RhinosF1:
or fix hostname but ...


paladox:
$wi->hostname is correct.

PR #3034, opened 2020-04-23
Move rest of wgDefaultUserOptions from LocalWiki to LocalSettings

paladox:


JohnFLewis:
Shouldn’t the wikis be + to add onto the defaults?

paladox:
Oh, your right.

PR #3039, opened 2020-04-26
Make CodeMirror on by default for solarawiki

MusikAnimal:


Reception123:
@MusikAnimal Extensions can be enabled in Special:ManageWiki/extensions, no need to edit LocalSettings :)

PR #3041, opened 2020-04-27
Adding Calendar-Wikivoyage Extension

examknow:
Adding Calendar-Wikivoyage extension per phab item T5137. Submodule has not yet been added. I will add comment when it is.

### DIFF
--- a/LocalExtensions.php
+++ b/LocalExtensions.php
@@ -237,6 +237,10 @@ if ( $wmgUseCSS ) {
 	wfLoadExtension( 'CSS' );
 }
 
+if ( $wmgUseCalendarWikivoyage ) {
+	wfLoadExtension( 'Calendar-Wikivoyage' )
+}
+
 if ( $wmgUseDarkMode ) {
 	wfLoadExtension( 'DarkMode' );
 }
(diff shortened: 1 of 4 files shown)


LocalExtensions.php
```diff
@@ -237,6 +237,10 @@
 	wfLoadExtension( 'CSS' );
 }
 
+if ( $wmgUseCalendarWikivoyage ) {
+	wfLoadExtension( 'Calendar-Wikivoyage' )
```
Reception123, line 241: Please use the original name (https://github.com/wikimedia/mediawiki-extensions-Calendar/blob/master/extension.json#L2) for this one or else it won't work

examknow: I will note that the mediawiki.org docs for this extension are incorrect
image

examknow:
Submodule added. Ready to merge when you are 😃.

LocalExtensions.php
RhinosF1: https://www.mediawiki.org/w/index.php?title=Topic:Vlax350jbs1yysjz&action=history

PR #3048, opened 2020-04-29
differenciate closing sitenotice for private wikis + fix style issues

Reception123:
T4164


Reception123:
@JohnFLewis Since this isn't my thing, is this the right way? I mean it works but in terms of whether there would be an easier way to "code" it

JohnFLewis:
Define a $siteNotice inside a if ( $cwPrivate ) { } else { }, and then you only need to run the one check on whether the wiki is closed and no need to duplicate the whole block.

PR #3052, opened 2020-04-30
Change wgArticlePath for solarawiki

MusikAnimal:
I'm not certain this won't break anything... like robots.txt. It looks like solarawiki might have Special pages indexed? Anyway we can fix that too?

If it's too much trouble, then don't bother. It'd just be neat to get rid of the /wiki/ given the top-level domain is .wiki (https://spcodex.wiki)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2774,6 +2774,7 @@ $wi->config->settings = [
 	// Server
 	'wgArticlePath' => [
 		'default' => '/wiki/$1',
+		'solarawiki' => '/$1',
 	],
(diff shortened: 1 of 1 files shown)


Reception123:
Unfortunately, we don't really do this as it would break a lot of things and require a lot of changes for one wiki.

Are you sure special pages are being indexed? That is disallowed by our robots.txt currently.

MusikAnimal:
I was saying that by removing the article path, robots.txt would need to be updated too or else Special pages would become indexed. But anyway, not worth it if it causes problems. Thanks anyway.

PR #3062, opened 2020-05-05
Define $wgGoogleSiteVerificationKey on solarawiki

MusikAnimal:


Reception123:
@MusikAnimal If this is for WikiSEO, you should find the setting in Special:ManageWiki/settings under "Edit"

MusikAnimal:
Oh there we go! I went looking for it in ManageWiki, just wasn't expecting it to be under "Edit". Thx

Reception123:
@MusikAnimal I agree, edit is quite a strange place for it to be in... 

If you need to look for a specific setting in ManageWiki you could also use Ctrl+F in ManageWikiSettings.php and find in which section is (if that's easier for you :) )

PR #3067, opened 2020-05-11
Use actual name of extension

RhinosF1:
Otherwise we just create confusion, weirdness and an unordered mess


Reception123:
This seems to disable the extension on wikis after the name is changed, so we would need to use toggleExtensions for all wikis using this

paladox:
I personally do not think it's worth the time doing this.

RhinosF1:
Per @Paladox

PR #3068, opened 2020-05-11
Add Femiwiki to ManageWiki

paladox:
Bug: T4663

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -535,7 +535,7 @@ $wgManageWikiExtensions = [
 		],
 		'erudite' => [
 			'name' => 'Erudite (Skin)',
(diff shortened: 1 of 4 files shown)


ManageWikiExtensions.php
```diff
@@ -548,6 +548,14 @@
 			'conflicts' => false,
 			'requires' => [],
 		],
+		'femiwiki' => [
+			'name' => 'Femiwiki (Skin)',
+			'linkPage' => 'https://www.mediawiki.org/wiki/Skin:Femiwiki',
```
JohnFLewis, line 553: Special:MyLanguage?

paladox: Thanks, done.

PR #3073, opened 2020-05-12
blacklist subdomains

examknow:
CreateWiki/RequestWiki will now throw an error when a wiki with the exact name of "wikiwiki", "wwwwiki", or "wikiswiki" is requested.


LocalSettings.php
```diff
@@ -294,6 +294,9 @@
 	],
 
 	// CreateWiki
+	'CreateWikiBlacklistedSubdomains' => [
```
paladox, line 297: Needs to be wgCreateWikiBlacklistedSubdomains

PR #3075, opened 2020-05-13
Fix order of wikibase extensions & run populateSites on install

RhinosF1:
Run Wikibase Site tables on client installs
Put client first in the ext-list

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -680,8 +680,7 @@ $wgManageWikiExtensions = [
 					],
 				],
 				'mwscript' => [
-					'extensions/flow/maintenance/flowCreateTemplates.php' => [
-						],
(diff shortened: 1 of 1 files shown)


ManageWikiExtensions.php
```diff
@@ -2056,6 +2059,25 @@
...
+			'requires' => [
+				'extensions' => [
+					'wikibaserepository',
+				],
+			],
+			'install' => [
+				'sql' => [
+					'wbc_entity_usage' => "$IP/extensions/Wikibase/client/sql/entity_usage.sql",
+				'mwscript' => [
+						'$IP/extensions/MirahezeMagic/maintenance/populateWikibaseSitesTable' => [],
```
paladox, line 2076: Cannot be done, because it'll error if the user re-enables wikibaseclient.

PR #3083, opened 2020-05-19
undeploy site scout

RhinosF1:


Reception123:
SiteScout has been removed from all wikis and is ready for removal.

PR #3085, opened 2020-05-19
use new arrays and = check

Reception123:


Reception123:
DO NOT MERGE! Parse error. I will have to go but I will fix this tomorrow morning unless someone else does before

PR #3087, opened 2020-05-24
unrestrict wgMaxImageArea

Reception123:
seems to be fine to allow it now


ManageWikiSettings.php
```diff
@@ -1214,10 +1214,10 @@
 	'wgMaxImageArea' => [
```
paladox, line 1214: This should have been moved under the media section and not kept in restricted.

PR #3098, opened 2020-06-04
Move $wgForegroundFeatures to new wiki name T5695

Amanda-Catherine:


LocalSettings.php
```diff
@@ -1461,6 +1461,17 @@
 	// Foreground
 	'wgForegroundFeatures' => [
 		'default' => [],
+		'marionetworkwiki' => [
+			'showActionsForAnon' => true,
+			'NavWrapperType' => 'divonly',
+			'showHelpUnderTools' => true,
+			'showRecentChangesUnderTools' => true,
+			'enableTabs' => true,
+			'wikiName' => &$GLOBALS['wgSitename'],
```
paladox, line 1470: Should be $wgSiteName not globals.

JohnFLewis:
The wiki was renamed, surely renaming the stanza from its original to its new name is better than adding it and not removing the old one?

Amanda-Catherine:
> The wiki was renamed, surely renaming the stanza from its original to its new name is better than adding it and not removing the old one?

except that they didn't specify the former URL in the task, so I have no idea what the original name was.

JohnFLewis:
They linked to the previous task in it

PR #3100, opened 2020-06-09
Fix T5708

Amanda-Catherine:


JohnFLewis:
For reference, if they enabled $wgAllowExternalImages in ManageWiki, this setting would be entirely unneeded.

PR #3102, opened 2020-06-10
simcitywiki configuration

Amanda-Catherine:


Reception123:
Well if it was disabling the application of global locks that would be a different story (and would definitely be declined) but if it is only global IP blocks it's a bit different. I'm also not too sure why this would be needed, IPs that are globally blocked are IPs that have violated global policies on multiple wikis, or open proxies. https://meta.miraheze.org/wiki/No_open_proxies_policy is a global policy and applies to all wikis, and according to current policy wikis cannot be exempted from said policy.

Amanda-Catherine:
As I said, all of these things are things that I have done before with Weather Wiki and WikiCanada, and @JohnFLewis has had no issues with them before.

PR #3113, opened 2020-06-12
Change license for airforcewiki and armywiki per T5738

soratako:
Contributors are only requesters.


soratako:
@Reception123 Since there are no contributions by other users on these wikis, I think there are no problem to marge. Please check [here](https://phabricator.miraheze.org/T5738#112012). (No need to hurry.)

PR #3115, opened 2020-06-12
Blacklist subdomain.miraheze

examknow:
Prevent users from using 'subdomain' as their dbname in wiki requests per a suggestion from @Amanda-Catherine.


Amanda-Catherine:
This doesn't seem to be working https://meta.miraheze.org/wiki/Special:RequestWikiQueue/12699

PR #3116, opened 2020-06-14
Add '.*scratch' to rw blacklist

examknow:
Adding rule per the request of @RhinosF1 who I assume will comment with more regarding the reason for this request.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -321,7 +321,7 @@ $wi->config->settings = [
 
 	// CreateWiki
 	'wgCreateWikiBlacklistedSubdomains' => [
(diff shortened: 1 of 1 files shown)


examknow:
Tbh banning a wiki topic should be a community or at the very least a steward issue.

RhinosF1:
My suggestion is this is put in place for 28 days to allow time for community discussion

sario528:
I agree that we should give the community a chance to comment on this.

examknow:
@RhinosF1 I would suggest you open up an RfC. We can keep this PR open until we have some community opinion.

RhinosF1:
A post will be made within the next day to allow discussion.

I only intend for this to be a temp solution until we can harden the content policy to prevent it.

Reception123:
@RhinosF1 status?

RhinosF1:
Waiting on RfC result

Pix1234:
Merging per [successful 45 day ban proposal ](https://meta.miraheze.org/wiki/Requests_for_Comment/Prohibit_hate-based_wikis)

PR #3131, opened 2020-06-28
Add $wgTrustedMediaFormats for polytopewiki T5804

Amanda-Catherine:


Southparkfan:
Per https://github.com/wikimedia/mediawiki/blob/REL1_34/includes/libs/mime/defines.php#L41, MEDIATYPE_PLAINTEXT should be MEDIATYPE_TEXT.

Although I wonder whether we want to remove a security warning, even if the actual files aren't actually malicious. .off files trigger this warning because the MIME type is text/plain, which is unfortunate, since it means we cannot put a specific MIME type on the allowed list.

Amanda-Catherine:
@Southparkfan I just did exactly what the user requested...

RhinosF1:
> @Southparkfan I just did exactly what the user requested...

maybe think about it?

paladox:
Need to change https://github.com/miraheze/mw-config/blob/c4893fd72b7ddd198ff2675a516fed7a14ac732e/LocalExtensions.php#L11 to 

$wi->config->settings['wgTrustedMediaFormats']['default']['application/sla'];

PR #3132, opened 2020-06-29
Misc. config for dcwiki T5838

Amanda-Catherine:

PR #3134, opened 2020-07-02
Update LocalExtensions.php

Amanda-Catherine:
technical change per @paladox on #3131


paladox:
Do this on your other changehttps://github.com/miraheze/mw-config/pull/3132

PR #3143, opened 2020-07-07
Change license for wikidemocracywiki per T5876

soratako:
All contribs by users without permission to license change, have been deleted.


Reception123:
Checked and all users that have contributed have agreed to the license change, and the other edits have been removed and revdel'd.

PR #3145, opened 2020-07-07
fix $wgRestrictionLevels for T5878

RhinosF1:


Reception123:
doesn't work on test2, seems to already be the default https://github.com/wikimedia/mediawiki-extensions-AuthorProtect/blob/master/extension.json#L32

RhinosF1:
> doesn't work on test2, seems to already be the default https://github.com/wikimedia/mediawiki-extensions-AuthorProtect/blob/master/extension.json#L32

We seem to be overriding extension.json and LE.php in LS.php. We need a way to make LS.php additional

PR #3168, opened 2020-07-14
Update LocalSettings.php

dmehus:
In the case of "wikitech," per discussion with @RhinosF1 on Discord following a declined wiki creation request. Though we may not need it imminently, it seems prudent to reserve this subdomain for potential Miraheze use as a public or private wiki in the future. Likewise, for "gazetteer," per the several declined wiki requests which @Pix1234 and I have declined, this one seeks to replicate our Gazetteer of Wikis page on Meta without some sort of community discussion and clearly articulating how we could not improve upon the data available in/provided by the WikiDiscover special page. We may want to wildcard "gazetteer" to include "gazeteer" as a misspelling.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -336,7 +336,7 @@ $wi->config->settings = [
 
 	// CreateWiki
 	'wgCreateWikiBlacklistedSubdomains' => [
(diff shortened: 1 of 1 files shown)


paladox:
Why was wikitech rejected?

RhinosF1:
@Paladox: wikitechwiki was blocked as requestor was likely LTA and writing in random foreign lauguages to throw us off that according to others were clear google translations so no objection to not blacklisting.

dmehus:
@paladox Has there been an update as to when these open pull requests can be merged into the master branch? I was going to add _subdomain_ as an additional subdomain to add to the blacklist, but saw it's already in there. So once a steward deletes [subdomain.miraheze.org](https://subdomain.miraheze.org/), that is closed due to the dormancy policy, we should be good as far no one being able to have their _subdomain_ hostname wikis accidentally created. ;)

PR #3174, opened 2020-07-15
Set batches 4 and 4 as read only, taking out batches 1 and 2 out of read only

paladox:


Southparkfan:
Setting batch 2 in RO but taking it out of RO as well? Are you sure?

paladox:
@Southparkfan thanks for catching, fixed!

PR #3188, opened 2020-07-24
Fix regex for wgCreateWikiBlacklistedSubdomains

dmehus:
Adding "sandbox" to restricted subdomains list because we've had a couple wiki request declines in the past month ([Wiki Request 12598](https://meta.miraheze.org/wiki/Special:RequestWikiQueue/12598) and [Wiki Request 13380](https://meta.miraheze.org/wiki/Special:RequestWikiQueue/13380)), and I don't want to see it inadvertently get created. Also, adding "noc" as a restricted subdomain as I didn't see it in this list, and it's highly conceivable Miraheze may want to use it for technical coordination purposes, either as a wiki or some other website. Finally, though I don't want to create it now, I was talking with @Reception123 on Discord about ways we can (a) increase Meta participation and (b) community volunteerism, and one of the ways in which I think this would be helpful would be through an Outreach wiki, so am requesting we restrict this subdomain for now.

Also, friendly pings @paladox and @RhinosF1, since they were involved in commenting and/or approving/merging my previous pull request.

PR #3189, opened 2020-07-24
Fix invalid regex

The-Voidwalker:
[0-15] does not match the sequence of numbers between 0 and 15, it instead matches the set of numbers 0, 1, or 5.

Instead we should simply match one or two digits. Having it match 0-15 would require some complexity that isn't needed (and I'm too lazy to figure that out). Wikis should not have subdomains that match these anyway.


The-Voidwalker:
Gonna try and merge these changes into #3188

PR #3197, opened 2020-07-31
[TEMP] - Up JobRunRate

RhinosF1:
We need to do something or this queue is never going to catch up. Would this work?


JohnFLewis:
Wouldn’t make any noticeable impact besides slowing down 1 in 4 web requests

RhinosF1:
Would it not run a job on 1 in 4 web requests as well though?

JohnFLewis:
Yes, but you could run 10s if not 100s via command line in the same time l

PR #3198, opened 2020-07-31
Add import option from Wikipedia to Meta-Wiki

MrJaroslavik:


Reception123:
@MrJaroslavik Minor but please insert the line alphabetically (so after the incubatorwiki one)

PR #3200, opened 2020-07-31
missing-wiki: add archive.org link

RhinosF1:
per discord suggestion

### DIFF
--- a/MissingWiki.php
+++ b/MissingWiki.php
@@ -63,7 +63,7 @@ if ( !$wgCommandLineMode ) {
 				<div style="text-align:center;">
 					<h2>What can I do?</h2>
 					<p class="lead">If you're a wiki visitor:</p>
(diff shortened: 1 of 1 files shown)


dmehus:
LGTM. My _only_ small suggestion might be to add a `,<br /> ` immediately preceding the `or` addition, for smaller monitors and screens. Aesthetically, it may look a little unbalanced otherwise with this one long line of text relative to really short lines of text elsewhere on the page?

dmehus:
This still hasn't been merged?

RhinosF1:
I was hoping for some feedback

dmehus:
@RhinosF1 It'd be nice to have some more comments...I would still add that one `<br />`, but otherwise, if no feedback, I'd take that to mean there's no opposition to it / no one really minds either way. Ah, I see you've requested a review from @paladox and @JohnFLewis, that's probably a good idea! :)

PR #3206, opened 2020-08-11
Add 'skipcaptcha' right to sysops on Meta

MrJaroslavik:
I need this change due to abuse filters and also this is uncontroversial i think

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2085,6 +2085,7 @@ $wi->config->settings = [
 			],
 			'sysop' => [
 				'interwiki' => true,
+				'skipcaptcha' => true,
 			],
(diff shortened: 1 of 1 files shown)


dmehus:
This is certainly non-controversial, and I personally have no problem with it, but I also see it as redundant. `skipcaptcha` is included within both `autoconfirmed` and `confirmed`. In the highly unlikely event a Meta `sysop` was _not_ `autoconfirmed`, they could easily just add `confirmed` to their account.

MrJaroslavik:
Also this right is in sysop toolkit by default.

Pix1234:
This is done onwiki

PR #3209, opened 2020-08-24
managewiki: migrate wgFooterIcons

RhinosF1:
WIP, something like this will work but I am on holiday. I can finish it in 10 days or someone can add the custom option as 'text' to ManageWiki


Reception123:
Conflicts. Should be re-created.

PR #3215, opened 2020-08-27
Add wgRestrictionLevels to ManageWiki per T5991

Universal-Omega:
I believe this will work, but not 100% certain.

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -349,6 +349,22 @@ $wgManageWikiSettings = [
 		'section' => 'edit',
 		'help' => 'Show more authors.',
 	],
+	'wgRestrictionLevels' => [
+		'name' => 'Restriction Levels',
+		'from' => 'mediawiki',
+		'restricted' => false,
+		'type' => 'usergroups',
+		'options' => [
+			'None' => '',
+		],
+		'overridedefault' => [
+			'',
+			'autoconfirmed',
+			'sysop',
+		],
+		'section' => 'edit',
+		'help' => 'Set of permission keys that can be selected on the "page protection" page.',
+	],
 	'wgRestrictionTypes' => [
 		'name' => 'Restriction Types',
 		'from' => 'mediawiki',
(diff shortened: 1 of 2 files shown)


RhinosF1 (commented):
What if wikis want more custom named ones?

Universal-Omega:
It should work with custom ones also per `'type' => 'usergroups'`

Universal-Omega:
Note: I did not remove the existing config for this that wikis did through GitHub PRs - I was unsure if that was needed it not, but in the past when moving configs to ManageWiki, they were removed, so I thought I'd mention that.

Reception123:
I tested this on test2 by adding the bureaucrat group as a restriction level and it didn't seem to work

Reception123:
My guess would be because https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L2262 is not fulfilled (it seems that for restriction levels the group must have its own name as a "user right").

ManageWikiSettings.php
```diff
@@ -349,6 +349,23 @@
 		'section' => 'edit',
 		'help' => 'Show more authors.',
 	],
+	'wgRestrictionLevels' => [
+		'name' => 'Restriction Levels',
+		'from' => 'mediawiki',
+		'restricted' => false,
+		'type' => 'userrights',
+		'options' => [
+			'no protection' => '',
```
paladox, line 358: Need to provide options here. The below part only makes sure it's the default.

ManageWikiSettings.php
```diff
@@ -349,6 +349,23 @@
 		'section' => 'edit',
 		'help' => 'Show more authors.',
 	],
+	'wgRestrictionLevels' => [
+		'name' => 'Restriction Levels',
+		'from' => 'mediawiki',
+		'restricted' => false,
+		'type' => 'userrights',
```
paladox, line 356: should be list-multi

ManageWikiSettings.php
paladox: Add all the options that wikis are using see https://github.com/miraheze/mw-config/blob/1e327aade41e4a316a6fda8394e243696e232581/LocalSettings.php#L2867

PR #3218, opened 2020-09-02
Update LocalSettings.php

dmehus:
While scanning this code looking for something else, noticed what appears to be a typo in 'default' and believe this should probably be corrected.

Pinging @Reception123 and maybe @RhinosF1 for good measure.


RhinosF1 (approved):
Good catch

PR #3221, opened 2020-09-03
Fix recently introduced configs

paladox:


RhinosF1:
Should this be reverted then per https://github.com/miraheze/mw-config/pull/3237#issuecomment-688703088

JohnFLewis:
That, or the override default default values changed

Universal-Omega:
When they were the opposite of what they are now, the configs did not work and we're force setting themselves. Cosmos coding is a bit different requiring a small change to ManageWikiPermissions possibly I guesz possibly.

PR #3229, opened 2020-09-04
Add RightFunctions extension (T6140)

Universal-Omega:


Reception123:
A mistake that is very hard to see, but the REL1_34 version [doesn't have an extension.json ](https://github.com/wikimedia/mediawiki-extensions-RightFunctions/tree/REL1_34) so you'll have to use require_once for this one and it can be changed when we upgrade to 1.35 after

Reception123 (changes requested):
per comment

Reception123 (approved):

PR #3234, opened 2020-09-07
add wgCosmosMainBackgroundColor to ManageWiki

Universal-Omega:


JohnFLewis:
@Reception123 the default values do not match. Why did you merge this?

Reception123:
My bad, I didn't notice that the LS value wasn't the same as overridefault

PR #3237, opened 2020-09-07
Add more Cosmos configs

Universal-Omega:
<s>Due to the amount of useful configs for the Cosmos skin, I did move it to a new "Cosmos" tab in ManageWikiSettings. Whether this is allowed or not, I am unsure, so if not, I will revert this back to the styling tab.  I will submit MirahezeMagic PR  to add `⧼settings-cosmos⧽` if this is approved.</s> reverted


LocalSettings.php
```diff
@@ -186,12 +186,36 @@
...
+	'wgCosmosContentBackgroundColor' => [
+		'default' => '#000',
+	],
+	'wgCosmosBannerBackgroundColor' => [
+		'default' => '#001e3b',
+	],
+	'wgCosmosLinkColor' => [
+		'default' => '#006cb0',
+	],
+	'wgCosmosBackgroundImageNorepeat' => [
```
Reception123, line 210: overridedefault is true but here the default is false

Reception123 (changes requested):

LocalSettings.php
Universal-Omega: That's intential. Paladox told me that that's how the boolean values have to work, as he changed all the other ones to do the same.

Reception123 (approved):
Per Paladox's previous merge

JohnFLewis:
That is wrong, override default MUST match LocalSettings.

Universal-Omega:
When they were the opposite of what they are now, the configs did not work and we're force setting themselves. Cosmos coding is a bit different requiring a small change to ManageWikiPermissions possibly I guesz possibly.

PR #3239, opened 2020-09-08
Re-add Extension:WikiHiero (T6048)

Universal-Omega:


Reception123:
Noting that this extension was already present in ext-list and LE.php from the previous install :)

PR #3247, opened 2020-09-10
add wgUFEnablePersonalDataFunctions to ManageWiki

Universal-Omega:


Universal-Omega:
Never mind. I just realized this 'should not' be enabled

PR #3253, opened 2020-09-13
Add static.miraheze.org as a default allowed value of $wgTemplateStylesAllowedUrls

Universal-Omega:
This:
* Adds static.miraheze.org as a default allowed value of `$wgTemplateStylesAllowedUrls`.
* Fixes the previous default config for wikimedia, to be back in sync with their configuration, and the default configuration of TemplateStyles.
* Removes the comment, which says to remove that when a patch on Gerrit is merged, this is removed, because that mentioned patch has been abandoned.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1390,22 +1390,23 @@ $wi->config->settings = [
 
 	// TemplateStyles config
 	'wgTemplateStylesAllowedUrls' => [
-		// Remove when https://gerrit.wikimedia.org/r/486828/ is merged
 		'default' => [
 			'audio' => [
-				'<^(?:https:)?\/\/upload\\.wikimedia\\.org\/wikipedia\/commons\/>',
-				'<^(?:https:)?\/\/static\\.miraheze.org\\.org\/>',
+				'<^(?:https:)?//upload\\.wikimedia\\.org/wikipedia/commons/>',
+				'<^(?:https:)?//static\\.miraheze\\.org/>',
 			],
 			'image' => [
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1394,12 +1394,15 @@
 		'default' => [
 			'audio' => [
 				'<^(?:https:)?\/\/upload\\.wikimedia\\.org\/wikipedia\/commons\/>',
+				'<^(?:https:)?\/\/static\\.miraheze.org\\.org\/>',
 			],
 			'image' => [
 				'<^(?:https:)?\/\/upload\\.wikimedia\\.org\/wikipedia\/commons\/>',
```
The-Voidwalker, line 1400:
```suggestion
				'<^(?:https:)?//upload\\.wikimedia\\.org/wikipedia/commons/>',
```

LocalSettings.php
```diff
@@ -1394,12 +1394,15 @@
...
+				'<^(?:https:)?\/\/static\\.miraheze.org\\.org\/>',
 			],
 			'svg' => [
 				'<^(?:https:)?\/\/upload\\.wikimedia\\.org/wikipedia\/commons\/[^?#]*\\.svg(?:[?#]|$)>',
```
The-Voidwalker, line 1404:
```suggestion
				'<^(?:https:)?//upload\\.wikimedia\\.org/wikipedia/commons/[^?#]*\\.svg(?:[?#]|$)>',
```

The-Voidwalker (changes requested):
See suggestions

The-Voidwalker (approved):
Looks good to me!

PR #3255, opened 2020-09-14
Adding media/journalism category to CreateWiki categories in LocalSettings.php

dmehus:
Adding a useful media/journalism category as we do have a number of other wikis which may appreciate this category, and per wiki request # 14221

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -483,6 +483,7 @@ $wi->config->settings = [
 			'Geography' => 'geography',
 			'Leisure' => 'leisure',
 			'Literature/Writing' => 'literature',
+			'Media/Journalism' => 'media',
 			'Medicine/Medical' => 'medical',
 			'Military/War' => 'military',
 			'Music' => 'music',


Reception123:
Around how many wikis would you think would fall into this category?

(P.S. no need for the in LocalSettings.php at the end :P)

dmehus:
> Around how many wikis would you think would fall into this category?
> 
> (P.S. no need for the in LocalSettings.php at the end :P)

@Reception123 No idea, but, conservatively, I would think that we could have at least three dozen potential media or journalism wikis currently, whether media criticism-type wikis or actual media wikis (i.e., online e-zines or digital newspaper-type wikis). I'm definitely in favour of broad categories, but we definitely need additional categories as we're still seeing a lot of "uncategorised" requests, which strongly suggests that none of our existing categories are applicable. I'll try and make time to proactively submit multiple categories in a single PR, but yeah, for now, this will be an important one to have.

PR #3256, opened 2020-09-14
set wgKartographerMapServer to openmapserver

Reception123:


Reception123:
doesn't currently work

PR #3261, opened 2020-09-18
configuration changes for hypopedia and scratchpad

Ultracane1997:
Making use of $wgRestrictionLevels, $wgRevokePermissions, $wgManageWikiPermissionsAdditionalRights, $wgManageWikiPermissionsAdditionalAddGroups, and $wgManageWikiPermissionsAdditionalRemoveGroups for my wiki projects. I'm trying to add the bureaucrat, extendedconfirmed, and leader restriction levels for both Hypopedia and Scratchpad. I'm also adding a banned user group with $wgRevokePermissions to Hypopedia and Scratchpad.


Ultracane1997:
I also removed the configuration variables for cyclonepediawiki as it no longer exists.

dmehus (commented):
This looks fine to me, based on my initial look. I encouraged @Ultracane1997 to just do the GitHub PR himself for, since he is familiar with editing the `LocalSettings.php` file for MediaWiki, rather than doing the Phabricator request first.

Ultracane1997:
I just got informed that editing the steward user group by myself is not allowed, so if that is indeed the case I could just remove that part and submit the rest of the changes.

LocalSettings.php
```diff
@@ -2115,6 +2114,36 @@
 			'bureaucrat' => [
 				'bureaucrat' => true,
 			],
+			'extendedconfirmed' => [
+				'extendedconfirmed' => true,
+			],
+			'leader' => [
+				'leader' => true,
+			],
+			'steward' => [
```
Universal-Omega, line 2123: Just wondering what you're doing modifying the steward usergroup locally? Seem to be adding oversight and CheckUser into Steward. I'm not certain, but that should be discouraged. Stewards can add themselves to those individual usergroups when really necessary.

Universal-Omega (changes requested):

LocalSettings.php
Ultracane1997: Well, I will just end up not doing it and remove that part of the changes, but I will keep the rest of the changes.

Ultracane1997: I have now updated it so that's no longer happening.

Ultracane1997:
If it's any easier to note this I have removed the local steward user group modification in the second commit.

Universal-Omega (approved):
All looks good to me now.

PR #3263, opened 2020-09-20
Add common sense History category to CreateWiki

dmehus:
Adding one of the major liberal arts/humanities academic disciplines, History, as a common sense category with lots of potential use

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -484,6 +484,7 @@ $wi->config->settings = [
 			'Fantasy' => 'fantasy',
 			'Gaming' => 'gaming',
 			'Geography' => 'geography',
+			'History' ==> 'history',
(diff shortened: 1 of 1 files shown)


Reception123:
I've seen quite a few history or alternative history wikis here and there, and I think it's a good idea to have more and more categories to slowly get less and less uncategorised

Reception123 (approved):

dmehus:
@Reception123 Yeah, my thoughts exactly. Ideally, I'd love to see us additional categories too, but we'd need to improve the GUI a bit to maybe categorize the categories themselves under several broad categories, to minimize the number of categories in the drop-down list. I definitely think it'd be a worthwhile feature request for CreateWiki, and I'll take to John about it in a few weeks.

PR #3270, opened 2020-09-27
whitelist certain pages for cineastaswiki per T6230

Reception123:


RhinosF1 (changes requested):
#L82 is listed as section for read whitelist changes, can we decide where things are going and either fix that order so it's not conflicting or put everything there?

Reception123:
@RhinosF1 I've only seen a "Per-wiki overrides" section, and in that section I saw several whitelist read variables.

RhinosF1 (changes requested):
Line 182

RhinosF1 (commented):
Per discord, my comment to be addressed in seperate PR

PR #3274, opened 2020-10-02
kartographer require jsonconfig instead of enable it with it

Universal-Omega:
For consistency with other extension requirements.


Reception123:
Script ran and JsonConfig enabled on all wikis that have Kartographer

PR #3285, opened 2020-10-09
Rename Flow to StructuredDiscussions in ManageWiki, and resort it alp…

dmehus:
…habetically in the ManageWiki extensions list

Per this request on Discord (https://discord.com/channels/407504499280707585/407537962553966603/763915864353275916), users are having some difficulty locating StructuredDiscussions in Special:ManageWiki/Extensions, since we've renamed it to its new name everywhere else on-wiki, the Extensions page, the Miraheze page, and even in ManageWiki itself. Former name should go in parentheses. I don't appear to be changing any database IDs, so likely no database changes are needed

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -662,68 +662,6 @@ $wgManageWikiExtensions = [
 				],
 			],
 		],
-		'flow' => [
-			'name' => 'Flow (StructuredDiscussions)',
-			'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:StructuredDiscussions',
-			'var' => 'wmgUseFlow',
-			'conflicts' => false,
-			'requires' => [],
-			'help' => 'Will start working 10-20 mins after enabling.',
-			'install' => [
-				'sql' => [
-					'flow_revision' => "$IP/extensions/Flow/flow.sql"
-				],
-				'namespaces' => [
-					'Topic' => [
-						'id' => 2600,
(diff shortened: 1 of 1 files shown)


Universal-Omega (approved):
LGTM!

dmehus:
@Universal-Omega Yeah, I mean we could possibly rename the _flow_ ID for the extension to _structureddiscussions_, too, so it's fully sorted alphabetically, but then we'd definitely need to update the database for the wikis that are using the extension.

Universal-Omega:
> @Universal-Omega Yeah, I mean we could possibly rename the _flow_ ID for the extension to _structureddiscussions_, too, so it's fully sorted alphabetically, but then we'd definitely need to update the database for the wikis that are using the extension.

See my reply on GitHub. Plus that would require changes to 3 other files I think maybe 2.

dmehus:
Yeah, exactly. I think this should be fine. No prejudice against someone making those other changes to rename _flow_ to _structureddiscussions_ in the DB in the future

Reception123:
@paladox how would we do this via the script? Because we'd have to remove "flow" from all wikis and then add "structureddiscussions" to all the wikis that had flow. Plus, wouldn't there be errors for the small time that "flow" would be removed?

Reception123:
Or could we first add SD, and then disable the old "flow"?

Reception123 (approved):
Since the variable name is not being changed, and just the *name* this is fine.

PR #3286, opened 2020-10-11
Add wikiapiary badge for closinglogosgroup (T6299)

Universal-Omega:


RhinosF1:
#3209 probably should be worked on soon. Will do if I get a chance.

PR #3318, opened 2020-10-26
hard-enable HAWelcome ext on csydeswiki (T6272)

Universal-Omega:


paladox:
they can enable it through managewiki.

Universal-Omega:
> they can enable it through managewiki.

@paladox See the task. Per a conversation with John, this was the only way to do this.

PR #3319, opened 2020-10-26
Add $wgWBRepoSettings['federatedPropertiesEnabled'] for paikallishistoriawikiwiki (T6353)

Universal-Omega:


Reception123:
Since there's already a whole tab just for Wikibase on ManageWiki, shouldn't we be doing this there?

PR #3330, opened 2020-10-28
Disable MinervaAlwaysShowLanguageButton for solarawiki

MusikAnimal:


Universal-Omega:
Would probably be better to not change the default configuration values right now as the default values are usually the defaults for the extension, as such it should usually remain true since that's likely the default value of the extension.

dmehus:
We should consider adding this to ManageWiki in a future ManageWiki update, maybe. I know we do that for @Universal-Omega's Cosmos skin, so it might be helpful to have additional configuration settings for the other skins, perhaps.

Universal-Omega:
> We should consider adding this to ManageWiki in a future ManageWiki update, maybe. I know we do that for @Universal-Omega's Cosmos skin, so it might be helpful to have additional configuration settings for the other skins, perhaps.

Yeah, at the moment things are a bit chaotic with the 1.35 upgrade but eventually I have plans to add *alot* of popular configuration values to ManageWikiSettings with approval from John, of course.

MusikAnimal:
> Would probably be better to not change the default configuration values right now as the default values are usually the defaults for the extension, as such it should usually remain true since that's likely the default value of the extension.

Yes the default for Minerva is `true`. But how many Miraheze wikis are the same 'family' of wiki but different language? I was assuming this is not common. So https://meta.wikimedia.org/wiki/Mission?useformat=mobile for instance has a "Language" link at the top-left. There is no de.miraheze.org or fr.miraheze.org, etc. I'm also assuming the wikis would need to be tied together via Wikibase. But again, just assumptions :) I'm not really sure.

That said, should I change the default to true, and false for my wiki? I just wanted to make sure my justification for this change was clear.

Universal-Omega:
> > Would probably be better to not change the default configuration values right now as the default values are usually the defaults for the extension, as such it should usually remain true since that's likely the default value of the extension.
> 
> Yes the default for Minerva is `true`. But how many Miraheze wikis are the same 'family' of wiki but different language? I was assuming this is not common. So https://meta.wikimedia.org/wiki/Mission?useskin=minerva for instance has a "Language" link at the top-left. There is no de.miraheze.org or fr.miraheze.org, etc. I'm also assuming the wikis would need to be tied together via Wikibase. But again, just assumptions :) I'm not really sure.
> 
> That said, should I change the default to true, and false for my wiki? I just wanted to make sure my justification for this change was clear.

Yes at least for now it would be preferable to only set it to false for your particular wiki. Thank you!

Universal-Omega (approved):

PR #3345, opened 2020-10-31
Remove deleted wikis

Universal-Omega:
Removes configurstion for deleted wikis, from the files, ContactPage.php, LocalSettings.php, and LocalWiki.php
Also:
* Fixes database name for `hrfwiki2wiki`, which was previously set as `hrfwiki2`
* Cleans up formatting
* Sets default for `$egApprovedRevsEnabledNamespaces` and `$wgPivotFeatures`
* Changes default for `$wgVisualEditorEnableVisualSectionEditing` to `mobile` (https://github.com/wikimedia/mediawiki-extensions-VisualEditor/blob/9f9966d1cda186086f74c8d810b081b4d189fe8c/extension.json#L93)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2152,31 +2152,6 @@ $wi->config->settings = [
 				'read' => true,
 			],
 		],
-		'+cyclonepediawiki' => [
-			'bureaucrat' => [
-				'bureaucrat' => true,
-			],
-			'extendedconfirmed' => [
-				'extendedconfirmed' => true,
-			],
-			'sysop' => [
-				'extendedconfirmed' => true,
-			],
-		],
-		'+dpwiki' => [
-			'bureaucrat' => [
-				'bureaucrat' => true,
-				'respected' => true,
-			],
-			'respected' => [
-				'respected' => true,
-			],
-		],
-		'+enigmawiki' => [
-			'scribe' => [
-				'read' => true,
-			],
-		],
 		'+hypopediawiki' => [
 			'bureaucrat' => [
 				'bureaucrat' => true,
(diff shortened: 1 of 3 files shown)


LocalSettings.php
```diff
@@ -117,13 +117,35 @@
 		'default' => false,
 	],
 	'wgPivotFeatures' => [
```
paladox, line 119: You'll need to set this in LocalWiki.php using $wi->config->settings['wgPivotFeatures']['default']

this is because you are using variables that won't be set in time.

paladox: you can always do +<dbname> so it latches onto the default + overrides. (this is for where you do thegreatwarwiki)

LocalSettings.php
```diff
@@ -566,19 +590,16 @@
...
-		'animatedfeetwiki' => true,
 		'animebathswiki' => true,
-		'baobabarchiveswiki' => true,
 		'beidipediawiki' => true,
 		'buswiki' => true,
 		'commonwealthwiki' => true,
 		'crappygameswiki' => true,
-		'crystalmaidenswiki' => true,
 		'cwarswiki' => true,
+		'drawnfeetwiki' => true,
```
paladox, line 602: Why is this being set? I don't see it being reordered.

Universal-Omega: Because there was a previous wiki which was removed, because it redirected to this one.

LocalSettings.php
Universal-Omega: Done

LocalSettings.php
paladox: oh I see.

LocalSettings.php
Universal-Omega: Copy from other conversation https://github.com/Universal-Omega/mw-config/blob/0427a0b205161670043f46df1202b2098f03c54e/LocalSettings.php#L1833 does work with variables.

PR #3352, opened 2020-11-01
Fix ManageWikiSettings load speed when Wikibase is enabled (Related to T6373)

Universal-Omega:


JohnFLewis:
Free input for this variable would cause a wiki to fatal if its not set to a valid local wiki database.

PR #3353, opened 2020-11-01
Remove FlaggedRevsRestrictionLevels from ManageWikiSettings

Universal-Omega:
Removing `$wmgFlaggedRevsRestrictionLevels` as it's not actually supported by ManageWiki for the same reason `$wgRestrictionLevels` is not. It uses user rights, not user groups. https://github.com/wikimedia/mediawiki-extensions-FlaggedRevs/blob/5904212d4ec3cb3a56b8bb493b8ba57f644f628e/backend/FlaggedRevs.php#L478

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -175,21 +175,6 @@ $wgManageWikiSettings = [
 		'section' => 'anti-spam',
 		'help' => 'Auto-review settings for edits/new pages.',
 	],
-	'wmgFlaggedRevsRestrictionLevels' => [
-		'name' => 'FlaggedRevs Restriction Levels',
-		'from' => 'flaggedrevs',
-		'restricted' => false,
-		'type' => 'usergroups',
-		'options' => [
-			'No Restriction' => '',
-		],
-		'overridedefault' => [
-			'',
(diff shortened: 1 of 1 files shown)


ManageWikiSettings.php
```diff
@@ -175,21 +175,6 @@
 		'section' => 'anti-spam',
 		'help' => 'Auto-review settings for edits/new pages.',
 	],
-	'wmgFlaggedRevsRestrictionLevels' => [
-		'name' => 'FlaggedRevs Restriction Levels',
-		'from' => 'flaggedrevs',
-		'restricted' => false,
-		'type' => 'usergroups',
```
paladox, line 182: You can change this to userrights

paladox:
The value shown for wmgFlaggedRevsRestrictionLevels is a group not a permission?

ManageWikiSettings.php
Universal-Omega: No we can't. Per @JohnFLewis it bloats the page and severally slows down ManageWikiSettings.

Universal-Omega:
Per #3633 I will attempt to fix this instead.

PR #3354, opened 2020-11-01
Fix flaggedrevs variable names to use original variable names rather then wmg ones

Universal-Omega:


Universal-Omega:
Done in #3356

PR #3355, opened 2020-11-01
why was darkmode a default extension?

Universal-Omega:


Pix1234:
To make darkmode possible globally afaik

Universal-Omega:
> To make darkmode possible globally afaik, PRs shouldnt be used to ask questions.

Fair point. It was more rhetorical. But why would it want to be made global? It doesn't even really work right.

Pix1234:
I think it was by request. Regardless theres no harm leaving it way it is

Universal-Omega:
> I think it was by request. Regardless theres no harm leaving it way it is

I suppose that's fair enough.

PR #3364, opened 2020-11-05
Move ContactPage to ManageWikiExtensions; Move ContactPage's RecipientUser config to ManageWikiSettings; which defaults to the wiki requester's username

Universal-Omega:


paladox:
Replace wiki founder's usage with wiki requester.

PR #3365, opened 2020-11-05
Remove Extension:Bootstrap from ManageWikiExtensions

Universal-Omega:
Extension:Bootstrap is only useful for Skin:Chameleon. If we ever end up installing that, then this extension can just be enabled with it, not through MWE as that gets confusing. https://phabricator.miraheze.org/T2841#65856


Universal-Omega:
@paladox I intentionally did not update extension-list because the extension should stay installed, just removed from MWE

paladox:
oh ok.

PR #3369, opened 2020-11-06
Use RESTbase with VisualEditor (On public wikis only) Related: T6411

Universal-Omega:
Can not be merged until restbase is configured.


dmehus:
👍. If this works, this may help to resolve our VisualEditor RESTbase errors, I think. It may also allow for better troubleshooting, too.

Universal-Omega:
per #3528

PR #3393, opened 2020-11-09
Raise $wgRCMaxAge to 720 days; convert it to days (T6417)

Universal-Omega:
`UPDATE mw_settings SET s_settings = JSON_REPLACE(s_settings, '$.wgRCMaxAge', CEIL(s_settings->>'$.wgRCMaxAge' / 86400))`
then, `rebuildrecentchanges.php`
will need run if this is merged

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -99,7 +99,7 @@ if ( (bool)$wmgWikiapiaryFooterPageName ) {
 }
 
 // $wgRCMaxAge
-$wgRCMaxAge = ceil($wgRCMaxAge/86400);  // Convert to seconds
+$wgRCMaxAge = ceil($wmgRCMaxAge/86400);  // Convert to seconds
 
 // $wgUrlShortenerAllowedDomains
 if ( !preg_match( '/^(.*).miraheze.org$/', $wi->hostname ) ) {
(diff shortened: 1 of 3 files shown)


paladox:
Any reason why we need to do this? This would have to be done on over 3,000 wikis.

Universal-Omega:
> Any reason why we need to do this? This would have to be done on over 3,000 wikis.

1) Much much easier to know what to set the value to
2) It would make the numbers smaller so they actually fit in the integer box.

LocalWiki.php
```diff
@@ -98,6 +98,9 @@ function onInactiveSiteNoticeAfter( &$siteNotice, $skin ) {
  	];
 }
 
+// $wgRCMaxAge
+$wgRCMaxAge = ceil($wgRCMaxAge/86400);  // Convert to seconds
```
paladox, line 102: Won't work. ManageWiki would override.

Universal-Omega: oh, so would `$wgRCMaxAge = ceil($wmgRCMaxAge/86400);` and make the managewiki one, wmg?

paladox: yeh.

Universal-Omega: done

BlackWidowMovie0:
@paladox Anything I can help with over here? If so, just point me in the right direction 👍🏾 . If not, that's ok, I guess 👎🏾.

Universal-Omega:
Will just use seconds

PR #3403, opened 2020-11-10
Add default value for $wgGalleryOptions

Universal-Omega:
https://github.com/wikimedia/mediawiki/blob/7e5f025690b39230c99be768880149410e411876/includes/Setup.php#L303

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1707,8 +1707,16 @@ $wi->config->settings = [
 	],
 
 	// Gallery Options
-	'+wgGalleryOptions' => [
-		'default' => [],
+	'wgGalleryOptions' => [
+		'default' => [
+			'imagesPerRow' => 0,
+			'imageWidth' => 120,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1707,8 +1707,16 @@
 	],
 
 	// Gallery Options
-	'+wgGalleryOptions' => [
-		'default' => [],
+	'wgGalleryOptions' => [
```
Reception123, line 1710: why remove the +?

Reception123 (changes requested):

LocalSettings.php
Universal-Omega: Because with the default value set, it was not needed (I don't think).

Universal-Omega: I just added the default value so it is easier to know what to change, if it is ever requested again.

Reception123 (approved):

PR #3411, opened 2020-11-10
remove wiki's config which has no effect

Universal-Omega:


Reception123 (approved):
It seems the wgModerationOnlyInNamespaces would contradict the first one, so there's no reason to keep it

PR #3421, opened 2020-11-12
Migrate $wgCompressRevisions to managewiki-restricted

Universal-Omega:


Reception123 (changes requested):
compressOld.php needs to be ran after doing this

PR #3424, opened 2020-11-12
Remove Extension:WhoIsWatching

Universal-Omega:


Reception123:
For some reason it was never enabled, but the wiki that requested it no longer exists anyway.

Reception123 (approved):

PR #3448, opened 2020-11-15
Migrate ContactPage to MWS/MWE

Universal-Omega:


The-Voidwalker:
Just so you know, this isn't working because $wgContactConfig uses an unusual structure that looks conflicting with our config management.

```php
$wgContactConfig['default'] = array(...);
```
Is equivalent to:
```php
$wi->config->settings->['wgContactConfig']['default']['default'] = array(...);
```

PR #3451, opened 2020-11-15
Fix contactpage extension name

Universal-Omega:


paladox:
Why remove that piece of information?

Universal-Omega:
> Why remove that piece of information?

Because it does not look right. If you think it is fine to keep it, we can, but it just seemed inconstant to me.

paladox:
Seems fine to me, let's people know they must set it.

JohnFLewis:
> Seems fine to me, let's people know they must set it.

Why for this extension and not for all the other extensions that have similar required configurations or even just configurations?

paladox:
> > Seems fine to me, let's people know they must set it.
> 
> Why for this extension and not for all the other extensions that have similar required configurations or even just configurations?

oh good question. I guess we can merge this then to follow the standard we've done for all the other extensions.

PR #3453, opened 2020-11-15
temporary removal of configurations per T6444

Hispano76:


LocalSettings.php
```diff
@@ -1860,20 +1860,6 @@
 			'eswiki',
 			'wikispecies',
 		],
-		'+hispanowiki' => [
-			'u',
-			'w',
-		],
-		'+privadowiki' => [
```
Universal-Omega, line 1867: privadowiki was not mentioned in T6444

Universal-Omega (changes requested):

LocalSettings.php
Hispano76: I didn't see it necessary, my wiki doesn't need that configuration anymore and I wanted to take advantage of it. I've done something like this before but forgot to mention it when I opened this change.

PR #3456, opened 2020-11-17
Fix translate right declaration

Universal-Omega:
translate-review right does not exist in the Translate extension. I believe that was meant to say translate-messagereview which does exist


dmehus:
:+1: LGTM.

PR #3458, opened 2020-11-17
Restrict SocialProfile userboard features

Universal-Omega:


Universal-Omega:
After a conversation with the developer of SocialProfile, this is not necessary.

PR #3506, opened 2020-12-01
Update LocalSettings.php

vernilla:
I think...I'm doing this right? God I feel dumb. :c


Universal-Omega:
I'll add them to ManageWiki shortly.

dmehus:
Thanks, @Universal-Omega, yeah I think these would probably be good settings to migrate to ManageWiki for local wiki bureaucrats.

Universal-Omega:
Added to ManageWiki with #3512 - however, if performance issues are caused, it will have to be reverted.

PR #3511, opened 2020-12-02
Logging: Drop wfDebug events

paladox:
I don't think we need these just yet.


paladox:
Or we can use sampling...

PR #3513, opened 2020-12-02
Un-restrict the enabling/disabling of file types

Pix1234:


Reception123 (changes requested):
While, as a proponent of this myself I support it, it's not just setting it to false.

You need to 1) move it out of the restricted // section 2) change 'section' => to something else

Universal-Omega:
We also should review and approve each individual file extension to verify they are safe before doing this.

Universal-Omega:
It has such a large conflict, that is easier to just re-do. I will do later.

PR #3530, opened 2020-12-07
Un-restrict $wgFileExtensions; add $wgFileBlacklist to still restrict dangerous ones.

Universal-Omega:


JohnFLewis:
Splitting to two settings seems unnecessary when HTMLForm has support for disabled elements in a list. Support should be added to ManageWiki for such a case, rather than leaving it down to a configuration problem

Universal-Omega:
> Splitting to two settings seems unnecessary when HTMLForm has support for disabled elements in a list. Support should be added to ManageWiki for such a case, rather than leaving it down to a configuration problem

Fair enough. Although we probably don't need wgFileBlacklist, just removing the $wgFileBlacklist extensions from $wgFileExtensions should be the same, if that'd be easier for you, since those extensions never worked anyways per https://www.mediawiki.org/wiki/Manual:$wgFileBlacklist

Pix1234 (approved):
Don't really need to be able to upload .crt but LGTM

Universal-Omega:
> Don't really need to be able to upload .crt but LGTM

5 wikis have that extension enabled for upload.

dmehus:
For what it's worth, I'm not sure about unrestricting this many extensions. I wouldn't mind seeing certain extensions like, `*.zip` and `*.7z` and a few others, remaining as `managewiki-restricted`. I think we can still be liberal with granting wikis the ability to use these extensions, but having a Steward review the request provides an appropriate check / second pair of eyes

Universal-Omega:
> For what it's worth, I'm not sure about unrestricting this many extensions. I wouldn't mind seeing certain extensions like, `*.zip` and `*.7z` and a few others, remaining as `managewiki-restricted`. I think we can still be liberal with granting wikis the ability to use these extensions, but having a Steward review the request provides an appropriate check / second pair of eyes

I can't say I completely disagree with you but I also think that we really don't need to keep zip and 7z restricted if they're almost always approved anyways, and archives such as these typically aren't to dangerous also but I suppose this could be investigated/looked into further first.

PR #3534, opened 2020-12-07
Move wgDiscordAdditionalIncomingWebhookUrls and wgSlackIncomingWebhookUrl to LS

paladox:


LocalExtensions.php
```diff
@@ -852,8 +852,10 @@
 	$wgSlackNotificationWikiUrl = $wgServer . '/w/';
 	$wgSlackShowNewUserEmail = false;
 	$wgSlackShowNewUserIP = false;
-	$wgSlackIncomingWebhookUrl =
-		$wmgWikiMirahezeSlackHooks[$wgDBname] ?? $wmgWikiMirahezeSlackHooks['default'];
+	$wi->config->settings['wgSlackIncomingWebhookUrl']['default'] = $wmgWikiMirahezeSlackHooks['default'];
```
JohnFLewis, line 855: Isn't the purpose of moving these to LS to remove them from LE?

PR #3540, opened 2020-12-07
Remove Parsoid from $wgIncidentReportingServices

Universal-Omega:


JohnFLewis:
Technically services shouldn’t be removed for historical purposes

PR #3543, opened 2020-12-07
re-order extension-list

Universal-Omega:


paladox:
Doesn't have i18n, thus this shouldn't be done.

PR #3549, opened 2020-12-08
 set $wgGenerateThumbnailOnParse to true if PdfHandler is enabled (to fix T6469)

Universal-Omega:


paladox:
This wont fix it @Universal-Omega

PR #3567, opened 2020-12-14
fix import sources for mrjaroslavikwiki

MrJaroslavik:
I ve fixed it per manual on mw.org, this should work.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2010,8 +2010,7 @@ $wi->config->settings += [
 			'wikipedia',
 		],
 		'+mrjaroslavikwiki' => [
-			'wikipedia',
-			'cswikipedia',
+		        'wikipedia' => [ 'cs', 'en' ],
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2010,8 +2010,7 @@
 			'wikipedia',
 		],
 		'+mrjaroslavikwiki' => [
-			'wikipedia',
-			'cswikipedia',
+		        'wikipedia' => [ 'cs', 'en' ],
```
Reception123, line 2013: could you please do it like you see below for zhdelwiki (ie different lines after the arrays ([])

Reception123 (changes requested):

LocalSettings.php
MrJaroslavik: Better?

Reception123 (approved):

PR #3570, opened 2020-12-15
Redo #2557

revi:
It still sucks. Why was it undid after all?

Ref: #2557

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -16,7 +16,7 @@ if ( $cwClosed ) {
 			'undelete' => true,
 		],
 	];
-	
+
 	if ( $wmgUseComments ) {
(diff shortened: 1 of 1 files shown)


paladox:
See https://github.com/miraheze/mw-config/blob/6d774a8ce7ff74e9195cdae44d31a1d0035c4b1d/LocalSettings.php#L3011

Pix1234:
Merging as i dont see where or why it was ever removed

JohnFLewis:
Per Paladox this is in LS.

Pix1234:
> Per Paladox this is in LS.

Yeah, only saw after i pressed merged as thats when the comment was loaded on the page for me, reverted

PR #3572, opened 2020-12-15
Fix setting overrides for wgDefaultUserOptions

paladox:
We create a new wmgDefaultUserOptions config which is local and then we add wmgDefaultUserOptions onto wgDefaultUserOptions by doing:

wgDefaultUserOptions = array_merge( $wgDefaultUserOptions, $wmgDefaultUserOptions );

Using + with wgDefaultUserOptions didn't appear to apply the overrides we had set in wgConf thus this change is necessary for this to work again.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2991,7 +2991,7 @@ $wi->config->settings += [
 	],
 
 	// Preferences
- 	'+wgDefaultUserOptions' => [
+ 	'wmgDefaultUserOptions' => [
  		'default' => [
  			'enotifwatchlistpages' => 0,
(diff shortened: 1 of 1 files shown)


JohnFLewis:
I’d rather look into this when I have time later. Our objective is to reduce out of LS config and wmg variables where possible.

paladox:
> I’d rather look into this when I have time later. Our objective is to reduce out of LS config and wmg variables where possible.

sure

JohnFLewis:
SiteConfiguration handles merging of pre-existing array keys really poorly in my opinion, so after looking this is the best temporary fix available.

PR #3576, opened 2020-12-16
Trying to make restriction levels ordered better

Ultracane1997:
I am attempting to see if this would make the protection levels ordered more in a way that doesn't trigger my OCD in hypopediawiki (Yes, I am a bureaucrat and administrator there.) If anyone has a better way to approach this please suggest them to me. Thank you.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3096,6 +3096,10 @@ $wi->config->settings += [
 			'member',
 		],
 		'+hypopediawiki' => [
+			'',
+			'user',
+			'autoconfirmed',
+			'sysop',
 			'bureaucrat',
 		],
 		'igrovyesistemywiki' => [


Ultracane1997:
@Reception123 by the way, if you're seeing this @dmehus suggested for me to try and make a PR for this.

LocalSettings.php
```diff
@@ -3096,6 +3096,10 @@
 			'member',
 		],
 		'+hypopediawiki' => [
+			'',
```
Universal-Omega, line 3099: This won't work right, I don't think. This would basically. Be duplication of them so would be `$wgRestrictionLevels` = [ '', 'user', 'autoconfirmed', 'sysop', '', 'user', autoconfirmed', 'sysop', 'bureaucrat' ];`

Universal-Omega:
No matter how this is done, it's either duplicative or it'll be the exact same as it is now.

PR #3579, opened 2020-12-21
Configuration Changes for r4356thwiki

R4356th:
This adds two interwiki import sources and one page to the read whitelist for r4356thwiki (a private wiki). Thank you.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2015,6 +2015,10 @@ $wi->config->settings += [
                                  'en', 
                           ],
 		],
+		'+r4356thwiki' => [
+			'scratchwiki',
+			'snapwiki'
+		]
 		'+sesupportwiki' => [
 			'mrjaroslavikwiki',
 		],
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2018,7 +2018,7 @@ $wi->config->settings += [
 		'+r4356thwiki' => [
 			'scratchwiki',
 			'snapwiki',
-		]
+		],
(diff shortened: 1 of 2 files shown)


dmehus (approved):
LGTM. 👍 

I can also confirm I've added the interwiki prefixes per @R4356th's request

LocalSettings.php
```diff
@@ -2015,6 +2015,10 @@
                                  'en', 
                           ],
 		],
+		'+r4356thwiki' => [
+			'scratchwiki',
+			'snapwiki'
+		]
```
Pix1234, line 2021: Needs a comma

Pix1234 (changes requested):

LocalSettings.php
R4356th: While that is not really necessary, I will add it anyway. Though the example snippet in CONTRIBUTING.md did not mention it.

R4356th: Done.

dmehus: Thanks, @Pix1234. I feel badly for missing the comma. @R4356th, when you say the comma is "not really necessary," do you mean the configuration would've still worked without it? If _so_, I did not know that.

R4356th: @dmehus, no, I thought Zppix was talking about line 2020, which was optional and not widely used (but used in this repository probably because it makes adding more items to arrays easier later). It would not have worked without the comma on line 2021.

PR #3584, opened 2020-12-22
Use rectalogic/notify-irc to send irc events

paladox:


MacFan4000:
for conditionals with jobs see https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

Universal-Omega:
Done in #3585

PR #3585, opened 2020-12-22
Add IRC notifications and workflow name

Universal-Omega:


MacFan4000:
per https://github.com/rectalogic/notify-irc/blob/master/action.yml#L25, we could have it login to nickserv if we wanted to.

PR #3602, opened 2020-12-26
set $wgKartographerMapServer to openstreetmap

Universal-Omega:


LocalSettings.php
```diff
@@ -2069,10 +2069,13 @@
...
 	'wgKartographerWikivoyageMode' => [
 		'default' => false,
-	 ],
+	],
 	'wgKartographerUseMarkerStyle' => [
 		'default' => false,
-	 ],
+	],
+	'wgKartographerMapServer' => [
+		'default' => 'https://a.tile.openstreetmap.org',
```
JohnFLewis, line 2077: I thought OSM's advice was don't use a specific server?

Universal-Omega: Oh, that might actually work. will remove `a.` then.

PR #3610, opened 2020-12-27
Add Import Sources for batfamilywiki, batmanwiki and snapwikiwiki

R4356th:
Adds import sources for batfamilywiki, batmanwiki and snapwikiwiki. I am a bureaucrat on all these wikis.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2025,6 +2025,9 @@ $wi->config->settings += [
 		'+simcitywiki' => [
 			'wikipedia',
 		],
+		'+snapwikiwiki' => [
+			'scratchwiki',
+		],
 		'wikitrashwiki' => [
 			'wikipedia' => [
 				'it',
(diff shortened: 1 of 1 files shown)


Universal-Omega:
As far as I'm aware this won't work for Fandom. Though could be wrong.

R4356th:
@Universal-Omega, I could do something like-
```PHP
'wikia' => [
    'batman'
]
, if that is what you mean.

Universal-Omega (dismissed):
batmanwikifandom is what I mean won't work

R4356th:
But why? I have added it as a interwiki link locally.

Universal-Omega:
> But why? I have added it as a interwiki link locally.

Because of Fandom's setup which blocks stuff like that

R4356th:
Interwikis are a MediaWiki core feature. I do not think it is possible to block these.

Universal-Omega:
> Interwikis are a MediaWiki core feature. I do not think it is possible to block these.

Fandom modifies core on their system significantly.

PR #3613, opened 2020-12-27
Adds editinterface protection level on devwiki

Naleksuh:


Naleksuh:
Adds editinterface protection level on https://dev.miraheze.org

Reception123 (approved):

PR #3616, opened 2020-12-27
Removing system sitenotice

dmehus:
As per the previous sitenotice's length of time (approximately 24 hours) that @Reception123 and I discussed on Discord

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -10,7 +10,7 @@ if ( $wmgSiteNoticeOptOut ) {
 // and don't comment it out
 $wgMajorSiteNoticeID = 55;
 
-/* // if ( !$wmgSiteNoticeOptOut ) {
+/* if ( !$wmgSiteNoticeOptOut ) {
 $wgHooks['SiteNoticeAfter'][] = 'onSiteNoticeAfter'; // show to all users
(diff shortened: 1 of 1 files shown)


RhinosF1:
This is a lot more important so we discussed a week for this. I did say that I'd get figures on how many password changes were being made.

Sitenotice.php
```diff
@@ -10,7 +10,7 @@
 // and don't comment it out
 $wgMajorSiteNoticeID = 55;
 
-// if ( !$wmgSiteNoticeOptOut ) {
+/* // if ( !$wmgSiteNoticeOptOut ) {
```
Universal-Omega, line 13: Can you please remove the `//` from here? While not really required I do think it simply looks better like that. Thanks!

Universal-Omega (changes requested):

Sitenotice.php
dmehus: @Universal-Omega Can you not make the change for me? For such a minor change, it seems a bit bureaucratic to have to make the requestor do it.

Universal-Omega: sure, I can I just typically don't like modifying others' PRs, but I can do that if you want.

dmehus:
@Universal-Omega Okay, thanks. :)

PR #3620, opened 2020-12-29
Stop installing globalblocking sql for each wiki

paladox:
We use a global table, do not need it to be installed on each wiki.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -533,8 +533,6 @@ $wi->config->settings += [
 			"$IP/extensions/CheckUser/cu_changes.sql",
 			"$IP/extensions/DataDump/sql/data_dump.sql",
 			"$IP/extensions/Echo/echo.sql",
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -533,8 +533,6 @@
 			"$IP/extensions/CheckUser/cu_changes.sql",
 			"$IP/extensions/DataDump/sql/data_dump.sql",
 			"$IP/extensions/Echo/echo.sql",
-			"$IP/extensions/GlobalBlocking/sql/global_block_whitelist.sql",
```
supertassu, line 536: this is needed locally, it allows locally disabling certain global blocks

paladox: Reverted in https://github.com/miraheze/mw-config/commit/85ea53fac46b2e381e9dea7a81c647ab62fdc965

PR #3626, opened 2020-12-30
Add ExemptFromUserRobotsControl settings for reviwikiwiki

revi:


dmehus:
@revi This is `managewiki-restricted`, no? If so, want me to change this for you?

Universal-Omega:
> @revi This is `managewiki-restricted`, no? If so, want me to change this for you?

This is not managewiki-restricted

dmehus:
@Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?

Universal-Omega:
> @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?

Eventually I plan to add it to ManageWiki. This is namespace overrides I believe. So would need to be added to MWN.

revi:
> @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?

`__NOINDEX__` is (by default) ignored on Contents Namespace.

dmehus:
> > @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?
> 
> Eventually I plan to add it to ManageWiki. This is namespace overrides I believe. So would need to be added to MWN.

@Universal-Omega Okay, but if this is a namespace override, then we _do_ already have that in Special:ManageWiki/namespaces. The fact you said you only _believe_ this is what this variable does suggests you're not 100% sure. Perhaps we should invite @JohnFLewis to clarify this variable's purpose?

dmehus:
> > @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?
> 
> `__NOINDEX__` is (by default) ignored on Contents Namespace.

@revi Oh. Can't you set the override in Special:ManageWiki/namespaces for each namespace, or is that too cumbersome, perhaps?

revi:
> > > @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?
> > 
> > 
> > `__NOINDEX__` is (by default) ignored on Contents Namespace.
> 
> @revi Oh. Can't you set the override in Special:ManageWiki/namespaces for each namespace, or is that too cumbersome, perhaps?

https://www.mediawiki.org/wiki/Manual:$wgExemptFromUserRobotsControl

dmehus:
> > > > @Universal-Omega Well, I don't understand what he's trying to do, then, as I know you can still set `__NOINDEX__` on specific pages even where the default robot policy is to allow indexing. Perhaps we should add this variable to ManageWiki then?
> > > 
> > > 
> > > `__NOINDEX__` is (by default) ignored on Contents Namespace.
> > 
> > 
> > @revi Oh. Can't you set the override in Special:ManageWiki/namespaces for each namespace, or is that too cumbersome, perhaps?
> 
> https://www.mediawiki.org/wiki/Manual:$wgExemptFromUserRobotsControl

@revi Oh, interesting. Weird we have it set to disallow the use of the magic words by default, as I know that's not the case on Meta, but yeah, weird. Thanks! 👍

revi:
As I demonstrated elsewhere, it is the case on Meta. (For the record) https://meta.miraheze.org/w/index.php?title=Stewards%27_noticeboard/Archive_16&action=info (`__NOINDEX__` and robots allowed)

PR #3645, opened 2021-01-05
Change HTML Meta and Title and SnapProjectEmbed's display name

R4356th:
Both of their names got changed and it is probably a good idea to update them in ManageWiki, too.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -33,7 +33,7 @@ $wgManageWikiExtensions = [
 			'requires' => [],
 		],
 		'htmlmetaadntitle' => [
-			'name' => 'HTML Meta and Title',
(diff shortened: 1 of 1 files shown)


ManageWikiExtensions.php
```diff
@@ -1945,7 +1945,7 @@
 			],
 		],
 		'snapprojectembed' => [
-			'name' => 'SnapProjectEmbed',
```
Pix1234, line 1948: On mw.org the ext page even refers to it as SnapProjectEmbed

R4356th: I am an upstream dev and can confirm I forgot to update it.

R4356th: @Pix1234, also see snapwiki/SnapProjectEmbed#12.

PR #3648, opened 2021-01-07
Add Translate extension-related custom configuration for snapwikiwiki

R4356th:
Adds Translate extension-related custom configuration for snapwikiwiki where I am a bureaucrat. Documentation for these can be found in the [extension's extension.json](https://github.com/wikimedia/mediawiki-extensions-Translate/blob/master/extension.json#L858).

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -182,6 +182,14 @@ if ( $wgDBname === 'simcitywiki' ) {
 	unset( $wgGroupPermissions['checkuser'] );
 }
 
+if ( $wgDBname === 'snapwikiwiki' ) {
+	$wgTranslateDocumentationLanguageCode = 'qqq';
+	$wgTranslateFuzzyBotName = 'TranslateExtension';
+	$wgTranslateSupportUrl = [
+		'page' => 'Snap! Wiki:Translation Administrators\' Noticeboard'
+	];
+}
+
 // Licensing variables
 switch ( $wmgWikiLicense ) {
 	case 'arr':
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -182,14 +182,6 @@ if ( $wgDBname === 'simcitywiki' ) {
 	unset( $wgGroupPermissions['checkuser'] );
 }
 
-if ( $wgDBname === 'snapwikiwiki' ) {
-	
-	$wgTranslateFuzzyBotName = 'TranslateExtension';
-	$wgTranslateSupportUrl = [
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -182,6 +182,14 @@ function wfDonateLink( $skin, &$bar ) {
 	unset( $wgGroupPermissions['checkuser'] );
 }
 
+if ( $wgDBname === 'snapwikiwiki' ) {
+	$wgTranslateDocumentationLanguageCode = 'qqq';
```
Universal-Omega, line 186: This can be done via Special: ManageWiki. In addition, this value ("qqq") is the default.

Universal-Omega (changes requested):

LocalWiki.php
R4356th: qqq is not default.

Universal-Omega: I noticed my mistake.

R4356th: And apologies for not noticing this before.

Universal-Omega: It should still be removed though since can be done in ManageWiki

R4356th: Done.

LocalWiki.php
```diff
@@ -182,6 +182,14 @@ function wfDonateLink( $skin, &$bar ) {
 	unset( $wgGroupPermissions['checkuser'] );
 }
 
+if ( $wgDBname === 'snapwikiwiki' ) {
```
Universal-Omega, line 185: I'm pretty sure that these can rather be added to LocalSettings.php rather then here. Thanks!

Universal-Omega (changes requested):

LocalWiki.php
R4356th: You will have to wait for a few hours then.

R4356th:
@Universal-Omega, would not adding this to ManageWiki be a better idea if we are actually moving this to LS.php?

Universal-Omega:
> @Universal-Omega, would not adding this to ManageWiki be a better idea if we are actually moving this to LS.php?

At the moment, no I need to add a new type to managewiki first that does not allow current users to be entered as if a real user is entered to this field it woud make the account a system account

Universal-Omega:
No changes made anymore?

R4356th:
> No changes made anymore?

Sorry about that. I will make a new pull request.

Universal-Omega:
> > No changes made anymore?
> 
> Sorry about that. I will make a new pull request.

No problem, and thanks!

PR #3653, opened 2021-01-10
Install purge user right with extension

Universal-Omega:


dmehus (commented):
@paladox told me before that, ideally, extensions should be adding user rights in their code when installed. Is that the case, or not the case here? If the former, then this LGTM. If the //latter//, my concern is we should fix this upstream to have extension install rights natively, otherwise unintended consequences can result

JohnFLewis:
Where is the purge user right defined? This adds it by default, but next time the group is modified, the user right is silently removed because it doesn’t exist.

Universal-Omega:
> Where is the purge user right defined? This adds it by default, but next time the group is modified, the user right is silently removed because it doesn’t exist.

It seems to still exist/available to add when I tested on Test2.

Universal-Omega:
Purge is a core user right.

JohnFLewis:
It would be useful to include this on future, As it looked like an extension specific right that doesn’t exist

Universal-Omega:
> It would be useful to include this on future, As it looked like an extension specific right that doesn’t exist

I am not to sure what you mean here?

JohnFLewis:
I’m not sure what is so hard to understand about, ‘in future, state this is a core right. Otherwise it looks like you’re removing an undefined user right that needs to be defined’

Universal-Omega:
> I’m not sure what is so hard to understand about, ‘in future, state this is a core right. Otherwise it looks like you’re removing an undefined user right that needs to be defined’

Oh, you mean in PR description. I understand now, my bad.

PR #3661, opened 2021-01-14
add wgARENamespaces to ManageWikiNamespaces

The-Voidwalker:
See https://meta.miraheze.org/w/index.php?title=Community_noticeboard&oldid=156690#Problem_with_ArticleRatings_extension for extra details

### DIFF
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -141,4 +141,14 @@ $wgManageWikiNamespacesAdditional = [
 		],
 		'requires' => [],
 	],
+	'wgARENamespaces' => [
+		'name' => 'Enable Article Ratings in this namespace',
+		'from' => 'articleratings',
+		'type' => 'check',
+		'main' => true,
+		'talk' => false,
+		'blacklisted' => [],
+		'overridedefault' => false,
+		'requires' => [],
+	],
 ];
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -148,7 +148,7 @@ $wgManageWikiNamespacesAdditional = [
 		'main' => true,
 		'talk' => false,
 		'blacklisted' => [],
-		'overridedefault' => false,
+		'overridedefault' => array_fill_keys( $wgContentNamespaces, true ),
 		'requires' => [],
 	],
 ];
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -148,7 +148,10 @@ $wgManageWikiNamespacesAdditional = [
 		'main' => true,
 		'talk' => false,
 		'blacklisted' => [],
-		'overridedefault' => array_fill_keys( $wgContentNamespaces, true ),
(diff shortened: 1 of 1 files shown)


dmehus:
This LGTM, as far as I can tell. Are there any custom configurations for this extension-added variable on existing wikis we'll need to migrate to ManageWiki/namespaces as well?

The-Voidwalker:
To my knowledge this variable has never been configured before, which is honestly impressive, considering it appears to be required to use the extension. 🙃

dmehus:
> To my knowledge this variable has never been configured before, which is honestly impressive, considering it appears to be required to use the extension. 🙃

Wow, heh. 👍

Universal-Omega (changes requested):
This needs to be added to LocalSettings.php also.

ManageWikiNamespaces.php
```diff
@@ -141,4 +141,14 @@
...
 		'requires' => [],
 	],
+	'wgARENamespaces' => [
+		'name' => 'Enable Article Ratings in this namespace',
+		'from' => 'articleratings',
+		'type' => 'check',
+		'main' => true,
+		'talk' => false,
+		'blacklisted' => [],
+		'overridedefault' => false,
```
Universal-Omega, line 151: Should be default to be on on $wgContentNamespaces (https://github.com/wikimedia/mediawiki-extensions-ArticleRatings/blob/5ca30436ade43d27edc9cb2669d2ab8395d06734/ArticleRatingsHooks.php#L10)

Universal-Omega (changes requested):

The-Voidwalker:
> This needs to be added to LocalSettings.php also.

I had looked into that, and it seems that none of the other settings in this file are in LS, so I just decided to leave it. Also, to my knowledge any setting in LS.php would be overwritten by the wiki config anyway.

As for setting the default, I'm not actually sure if `overridedefault` (in this context) works the way we'd be expecting it to. I'll take a look.

ManageWikiNamespaces.php
```diff
@@ -141,4 +141,14 @@
...
 		'requires' => [],
 	],
+	'wgARENamespaces' => [
+		'name' => 'Enable Article Ratings in this namespace',
+		'from' => 'articleratings',
+		'type' => 'check',
+		'main' => true,
+		'talk' => false,
+		'blacklisted' => [],
+		'overridedefault' => array_fill_keys( $wgContentNamespaces, true ),
```
Universal-Omega, line 151: Should be overridedefault => default => format I believe

Universal-Omega (changes requested):

Universal-Omega (approved):

ManageWikiNamespaces.php
Universal-Omega, line 151: will need to have the `'default' => false,` value also

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #3677, opened 2021-01-24
Blacklist sensitive user rights from being `*`

dmehus:
Per internal discussion with @The-Voidwalker and @JohnFLewis


dmehus:
We'll need to also run a query to determine which wikis have assigned any of these user rights to `*`, and then run a separate query to remove them from same.

dmehus:
@Universal-Omega Can you figure out the query we can run to see (a) which wikis had set any of these user rights to `*` and of those that did (b) a second query that removes the listed rights from `*`?

Universal-Omega:
> @Universal-Omega Can you figure out the query we can run to see (a) which wikis had set any of these user rights to `*` and of those that did (b) a second query that removes the listed rights from `*`?

Yes, tomorrow or later tonight just merged this to prevent more in the meantime.

dmehus:
> > @Universal-Omega Can you figure out the query we can run to see (a) which wikis had set any of these user rights to `*` and of those that did (b) a second query that removes the listed rights from `*`?
> 
> Yes, tomorrow or later tonight just merged this to prevent more in the meantime.

Okay, thanks. Can you DM me the list of wikis once you run it? I'm quite curious to know how many and which ones.

PR #3679, opened 2021-01-26
Add Naver site verification key for linkwikiwiki (T6764)

R4356th:
Adds Naver site verification key for linkwikiwiki per T6764. The wiki is private so I had to copy the DB name from a custom domain request.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -185,6 +185,15 @@ if ( $wgDBname === 'erislywiki' ) {
 	}
 }
 
+// T6764
+if ( $wgDBname === 'linkwikiwiki' ) {
+	$wgHooks['BeforePageDisplay'][] = 'addNaverVerificationKey';
+	
+	function addNaverVerificationKey ( OutputPage $out ) {
+		$out->addMeta( 'naver-site-verification', '42e328a7c17f55beed38a6ad399777b6de20d6b9' );
+	}
+}
+
 if ( $wgDBname === 'metawiki' ) {
 	$wgHooks['BeforePageDisplay'][] = 'wfModifyMetaTags';
 
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -176,9 +176,12 @@ if ( $wmgUseYandexTranslate ) {
 }
 
 // Per-wiki settings
+
+//T5981
 if ( $wgDBname === 'erislywiki' ) {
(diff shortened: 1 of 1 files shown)


RhinosF1 (commented):
The example above seems to use a different hook. Is there any difference between the two? Should one be changed?

Universal-Omega:
> The example above seems to use a different hook. Is there any difference between the two? Should one be changed?

I think this one is the better one in this case (the one R4356th used)

R4356th:
The above example is pretty odd as it is not really using the ``ParserOutput`` hook.

RhinosF1:
I stole it off a MediaWiki.org example when I wrote that.

R4356th:
> I stole it off a MediaWiki.org example when I wrote that.

Okay then maybe consider reading the docs for ``OutputPage`` or reading the example of Meta if you are still uncomfortable.

LocalWiki.php
```diff
@@ -178,13 +178,22 @@ function onInactiveSiteNoticeAfter( &$siteNotice, $skin ) {
 // Per-wiki settings
 if ( $wgDBname === 'erislywiki' ) {
 	$wgHooks['OutputPageParserOutput'][] = 'onOutputPageParserOutput';
```
Universal-Omega, line 180: Will cause errors if this hook is not also changed with removal of param in function

LocalWiki.php
```diff
@@ -178,13 +178,22 @@ function onInactiveSiteNoticeAfter( &$siteNotice, $skin ) {
...
-	function onOutputPageParserOutput( OutputPage &$out, ParserOutput $parseroutput ) {
+	function onOutputPageParserOutput( OutputPage &$out ) {
```
Universal-Omega, line 181: Function should be renamed to for new hook

Universal-Omega (changes requested):

LocalWiki.php
R4356th: Done.

Universal-Omega (approved):

Reception123:
@Universal-Omega since you approved, is this waiting for anything else before it should get merged?

Universal-Omega:
> @Universal-Omega since you approved, is this waiting for anything else before it should get merged?

Nope I just haven't had PC access in 2 days. I will today and will merge

PR #3680, opened 2021-01-28
Add LogoFunctions (T6766)

R4356th:
Adds LogoFunctions per T6766. This should be merged with miraheze/mediawiki#1092.

### DIFF
--- a/LocalExtensions.php
+++ b/LocalExtensions.php
@@ -519,6 +519,10 @@ if ( $wmgUseListings ) {
 	wfLoadExtension( 'Listings' );
 }
 
+if ( $wmgUseLogoFunctions ) {
+	wfLoadExtension( 'LogoFunctions' );
+}
+
 if ( $wmgUseLoopsCombo ) {
 	wfLoadExtension( 'Loops' );
 }
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1102,6 +1102,9 @@ $wi->config->settings += [
 	'wmgUseListings' => [
 		'default' => false,
 	],
+	'wmgUseLogoFunctions' => [
+		'default' => false,
+	],
 	'wmgUseLoopsCombo' => [
 		'default' => false,
 	],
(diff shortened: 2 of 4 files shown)


Reception123 (changes requested):
Everything else looks good, but it also needs to be added to LocalSettings.php under [this section](https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L789)

R4356th:
> Everything else looks good, but it also needs to be added to LocalSettings.php under [https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L789](this section)

Yeah, I know and am doing that. :)

Reception123:
> > Everything else looks good, but it also needs to be added to LocalSettings.php under [https://github.com/miraheze/mw-config/blob/master/LocalSettings.php#L789](this section)
> 
> Yeah, I know and am doing that. :)

Oh sorry, didn't know you weren't finished yet. My bad then

R4356th:
This can be merged but the extension adds a configuration variable, ``$wgNamespaceLogos``, which needs namespace numbers (not constants) and I am not sure which type ManageWiki supports. It should be noted that there is an [upstream bug](https://phabricator.wikimedia.org/T234098) regarding logos being incorrectly set so we should consider if we want to add this to ManageWiki as well.

Universal-Omega (approved):

Universal-Omega:
> This can be merged but the extension adds a configuration variable, `$wgNamespaceLogos`, which needs namespace numbers (not constants) and I am not sure which type ManageWiki supports. It should be noted that there is an [upstream bug](https://phabricator.wikimedia.org/T234098) regarding logos being incorrectly set so we should consider if we want to add this to ManageWiki as well.

All constants are are namespace numbers I think so both would probably work. But don't add it in the PR I will in another one. (To MWN)

R4356th:
> > This can be merged but the extension adds a configuration variable, `$wgNamespaceLogos`, which needs namespace numbers (not constants) and I am not sure which type ManageWiki supports. It should be noted that there is an [upstream bug](https://phabricator.wikimedia.org/T234098) regarding logos being incorrectly set so we should consider if we want to add this to ManageWiki as well.
> 
> All constants are are namespace numbers I think so both would probably work. But don't add it in the PR I will in another one after I look into it first. (To MWN)

Okay. Sounds good. :)

Reception123 (approved):

PR #3685, opened 2021-01-29
Add wgDiscordAvatarUrl to ManageWikiSettings

Kelvs599:


ManageWikiSettings.php
```diff
@@ -1187,6 +1187,15 @@
 		'help' => 'This allows to set whether or not it should be possible for anonymous visitors of the wiki to dismiss the site notice shown.',
 		'requires' => [],
 	],
+	'wgDiscordAvatarUrl' => [
+		'name' => 'Discord Notification Avatar URL',
+		'from' => 'mediawiki',
+		'type' => 'text',
+		'overridedefault' => false,
```
Universal-Omega, line 1194: Should be '' to match  LS I think.  Not a huge deal just what I think it should be.

ManageWikiSettings.php
```diff
@@ -1187,6 +1187,15 @@
 		'help' => 'This allows to set whether or not it should be possible for anonymous visitors of the wiki to dismiss the site notice shown.',
 		'requires' => [],
 	],
+	'wgDiscordAvatarUrl' => [
+		'name' => 'Discord Notification Avatar URL',
+		'from' => 'mediawiki',
+		'type' => 'text',
```
Universal-Omega, line 1193: Why not use 'type' => 'url'?

Universal-Omega (changes requested):

ManageWikiSettings.php
Kelvs599: Fixed it.

Universal-Omega (approved):

ManageWikiSettings.php
Universal-Omega: Thanks!

ManageWikiSettings.php
Kelvs599: Sorry, I was copying wgLogo and didn't realize.

Universal-Omega: That's alright, thanks!

PR #3698, opened 2021-02-02
Add sitenotice regarding gluster migration

dmehus:
@Universal-Omega

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -1,6 +1,6 @@
 <?php
 
-if ( $wmgSiteNoticeOptOut ) {
+/* if ( $wmgSiteNoticeOptOut ) { */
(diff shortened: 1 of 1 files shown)


Sitenotice.php
```diff
@@ -1,22 +1,22 @@
 <?php
 
-if ( $wmgSiteNoticeOptOut ) {
+ if ( $wmgSiteNoticeOptOut ) { */
```
Universal-Omega, line 3: remove the */

Universal-Omega (changes requested):

Sitenotice.php
dmehus: This was before your latest fix, I think?

PR #3699, opened 2021-02-02
Correct grammar, improve phrasing, and add detail

dmehus:
Removes redundant phrasing, improves grammar, and adds clarity and detail (i.e., "new, upgraded infrastructure") so customers see benefits.


dmehus:
@Universal-Omega See above. This phrasing will correct grammar, add customer-centric benefits, highlight detail, and reduce redundancies.

PR #3704, opened 2021-02-02
Remove short term sitenotice that uploading is available

dmehus:
Several announcements on Discord together with this short sitenotice and the fact that actually uploading files will make it clear that uploads are again available. No need for more than a few hour sitenotice.

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -1,6 +1,6 @@
 <?php
 
-/* if ( $wmgSiteNoticeOptOut ) {
+if ( $wmgSiteNoticeOptOut ) {
 	// only show important notices when optout
 	$wi->config->settings['wgNoticeProject']['default'] = 'optout';
 }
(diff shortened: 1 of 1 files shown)


Sitenotice.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-if ( $wmgSiteNoticeOptOut ) {
+/* if ( $wmgSiteNoticeOptOut ) {
```
Universal-Omega, line 3: Should not comment out this if statement.

Universal-Omega (changes requested):

Sitenotice.php
dmehus: @Universal-Omega Oh okay, where do you want me to move it to then?

Universal-Omega: To only the second portion where the sitenotice actually is.

dmehus: > To only the second portion where the sitenotice actually is.

Okay, doing that now, and hoping I get it right.

dmehus: 👍 **Done**

PR #3729, opened 2021-02-18
Fix default value for wgUsersNotifiedOnAllChanges

paladox:


Universal-Omega:
Won't work. It'll make the input display as `array()` as that's what it was originally. This config should just be removed TBH due to somewhat incompatibility as this will save the config as a string no matter what I believe, and requires explode( "\n", value ) to fix. I could be wrong but I think that's how it is.

PR #3732, opened 2021-02-19
Use Special:MyLanguage for wgCookieWarningMoreUrl

R4356th:
Per reasonable request on the Administrators' noticeboard on Meta by Sabelöga.


dmehus:
:+1: LGTM. I can't think of any issues with this, as we're just changing the standard public URL of the cookie warning message. We're not adding Special:MyLanguage to an on-wiki interface message or anything like that, so I have no concerns with this.

PR #3736, opened 2021-02-20
Add wgNoFollowLinks and wgNoFollowNsExceptions to ManageWiki

Kelvs599:


ManageWikiNamespaces.php
```diff
@@ -58,6 +58,16 @@
 		'overridedefault' => false,
 		'requires' => [],
 	],
+	'wgNoFollowNsExceptions' => [
```
Universal-Omega, line 61: Needs to be added to LocalSettings.php.

Universal-Omega (changes requested):

ManageWikiNamespaces.php
Kelvs599: Looking through LocalSettings.php, it doesn't seem like any of the ManageWikiNamespaces configs are there?

Universal-Omega: Can you please still add this there? That's something that I'll look into later. It can't hurt to add it.

Kelvs599: Done.

PR #3738, opened 2021-02-24
Add wgNaverSiteVerificationKey

paladox:


ManageWikiSettings.php
```diff
@@ -1719,6 +1719,15 @@
 		'help' => 'Site verification key for Pinterest Console.',
 		'requires' => [],
 	],
+	'wgNaverSiteVerificationKey' => [
+		'name' => 'Pinterest Site Verification Key (WikiSEO)',
```
Universal-Omega: Pinterest?

ManageWikiSettings.php
```diff
@@ -1719,6 +1719,15 @@
...
+		'from' => 'wikiseo',
+		'type' => 'text',
+		'overridedefault' => false,
+		'section' => 'seo',
+		'help' => 'Site verification key for Pinterest Console.',
```
Universal-Omega: Again

ManageWikiSettings.php
paladox: that's an old commit.

Universal-Omega: Oh. Whoops. My bad.

PR #3760, opened 2021-03-08
Remove canned responses

RhinosF1:


Reception123:
Temporarily until a solution is found for an "Other" tab

PR #3767, opened 2021-03-10
wgCreateWikiBlacklistedSubdomains: change some \d{1,2} to \d{0,2}

Universal-Omega:


dmehus (commented):
All I remember is @The-Voidwalker had a reason for using `d{1,2}`, but I can't remember what it was. We should get his comment here before merging this.

PR #3780, opened 2021-03-13
Move remaining configs out of LocalExtensions.php

Universal-Omega:


JohnFLewis:
This PR moves config unnecessarily out of LE, making no improvements except increasing the complexity of the LS file - thereby unconstructively undoing progress we’ve made over recent years.

PR #3782, opened 2021-03-14
Add Art & Architecture category to CreateWiki

dmehus:
We do get quite a few wikis that propose to create either a wiki about buildings and other real world infrastructure. Combining with "Art", as is common, also means this could potentially be used for the many artist repository wikis or even fan-made original character repositories/catalogues, too. It's also a common categorization

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -701,6 +701,7 @@ $wi->config->settings += [
 	],
 	'wgCreateWikiCategories' => [
 		'default' => [
+			'Art & Arcitecture' -> 'artarc',
 			'Automotive' => 'automotive',
 			'Community' => 'community',
 			'Education' => 'education',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -701,6 +701,7 @@
 	],
 	'wgCreateWikiCategories' => [
 		'default' => [
+			'Art & Arcitecture' -> 'artarc',
```
RhinosF1, line 704: - should be an =

RhinosF1 (changes requested):

LocalSettings.php
dmehus: @RhinosF1 Yeah, I noticed that _just_ before you replied, so already fixed it.

RhinosF1 (approved):

LocalSettings.php
RhinosF1: Seen the fix and approved the change after that change.

dmehus: Wait, don't merge this yet. I noticed another typo.

dmehus: Fixed.

PR #3787, opened 2021-03-16
Add localhost and 127.0.0.1 to wgRottenLinksExcludeWebsites (T6971)

Universal-Omega:


RhinosF1 (approved):

RhinosF1 (changes requested):
Says don't include protocol

Universal-Omega:
> Says don't include protocol

Oh, whoops. I guess I didn't look enough. My bad. Fixing then will merge. Thanks!

PR #3789, opened 2021-03-16
Add new "Song Contest" category for CreateWiki

R4356th:
We get a lot of requests for wikis related to song contests these days. A category dedicated to those would be great.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -727,6 +727,7 @@ $wi->config->settings += [
 			'Religion' => 'religion',
 			'Science' => 'science',
 			'Software/Computing' => 'software',
+			'Song Contest' => 'sc',
 			'Sports' => 'sport',
 			'Uncategorised' => 'uncategorised',
 		],
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -727,7 +727,7 @@ $wi->config->settings += [
 			'Religion' => 'religion',
 			'Science' => 'science',
 			'Software/Computing' => 'software',
-			'Song Contest' => 'sc',
+			'Song Contest' => 'songcontest',
 			'Sports' => 'sport',
 			'Uncategorised' => 'uncategorised',
 		],


LocalSettings.php
```diff
@@ -727,6 +727,7 @@
 			'Religion' => 'religion',
 			'Science' => 'science',
 			'Software/Computing' => 'software',
+			'Song Contest' => 'sc',
 			'Sports' => 'sport',
```
Universal-Omega, line 731: Why use just `sc`? Why not songcontest?

Universal-Omega (changes requested):

LocalSettings.php
R4356th: To save time. There are two other examples above.

Universal-Omega: This is not as descriptive though and that's because those include special characters that I don't think can be in this.

R4356th: Done.

Universal-Omega: Thanks!

Universal-Omega (approved):

dmehus:
We've been using Music for these wikis. Given how many existing song contest wikis are categorized as _music_ already, I personally feel as though we should merge _Song Contest_ category into the _Music_ category, and possibly rename it as either _Music (including Song Contest wikis)_ or _Music & Song Contests_. Otherwise, we'll end up (a) fragmenting how wikis are categorized, reducing the utility to the categorization scheme in WikiDiscover, (b) creating too many narrow categories, and/or (c) both.

R4356th:
> We've been using Music for these wikis. Given how many existing song contest wikis are categorized as _music_ already, I personally feel as though we should merge _Song Contest_ category into the _Music_ category, and possibly rename it as either _Music (including Song Contest wikis)_ or _Music & Song Contests_. Otherwise, we'll end up (a) fragmenting how wikis are categorized, reducing the utility to the categorization scheme in WikiDiscover, (b) creating too many narrow categories, and/or (c) both.

I think the number of requests for this kind of wikis justifies the addition.

dmehus:
> > We've been using Music for these wikis. Given how many existing song contest wikis are categorized as _music_ already, I personally feel as though we should merge _Song Contest_ category into the _Music_ category, and possibly rename it as either _Music (including Song Contest wikis)_ or _Music & Song Contests_. Otherwise, we'll end up (a) fragmenting how wikis are categorized, reducing the utility to the categorization scheme in WikiDiscover, (b) creating too many narrow categories, and/or (c) both.
> 
> I think the number of requests for this kind of wikis justifies the addition.

Well, I_do_ think it would still be better to merge the two categories into the original Music category, given how many song contest wikis are categorized in this way. Ideally, it'd be nice to have a nested wiki category system with broader and narrower categories, but since we have one level of category, we should use a broader categorization scheme as we've always done, as I _do_ know that Reception123 is hesitant to add too many narrowly focused categories.

I _do_ agree with you, though, that given the number of song contest wikis we have, we could add a parenthetical qualifier to the **Music** category that notes this category includes song contest wikis.

PR #3793, opened 2021-03-17
Comments conflicts with ProtectionIndicator (T6987)

R4356th:
Please see T6987 for details. I tested the two on my private wiki to confirm this.


Universal-Omega:
I was going to do this but decided against it, but no harm I guess.

PR #3801, opened 2021-03-22
Very Minor Grammatical Fix to a CW Canned Response

R4356th:
The period should have been out of the quote.


dmehus:
👍 LGTM. I never know whether to put the period inside or outside the quotation marks. I wonder if this could be a difference in English variations as, at least in Canada, we were always taught to never have a period outside the quotation marks? Or maybe it's something akin to the Oxford comma in that both variations are correct except those who specialize in English grammar cannot agree on which is correct? In any event, I'm not fussed about which one we want to use.

PR #3808, opened 2021-03-25
make none of the above first createwiki purpose

Reception123:
I think it makes more sense to have it default to none of the above instead. And yes, I do realise then it should probably be called "None of the options below" rather than above

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -698,6 +698,7 @@ $wi->config->settings += [
 	],
 	'wgCreateWikiPurposes' => [
 		'default' => [
+			'None of the above' => 'None of the above',
 			'Alternate history wiki' => 'Alternate history wiki',
 			'Class or group project education wiki' => 'Class or group project education wiki',
 			'Curriculum resource wiki' => 'Curriculum resource wiki',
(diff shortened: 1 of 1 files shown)


Universal-Omega:
It's not none of the above if it's the top one.

Reception123:
> It's not none of the above if it's the top one.

Heh we conflicted, I just edited my initial comment to say that

Universal-Omega:
Maybe we should make the key be '' and default it to that in CW itself?

RhinosF1:
> Maybe we should make the key be '' and default it to that in CW itself, then keep the none of the above options at the bottom but still default to it?

That makes more sense

dmehus:
@Reception123 I disagree. The idea was _not_ to have a "none of the above" purpose. I believe it should be removed, and additional purposes should be added. The initial list was a temporary, non-exhaustive list, and @Universal-Omega promised me we'd add additional purposes as required. Otherwise, we defeat the whole purpose behind CreateWiki learning from the predefined purpose field.

dmehus:
I originally wanted this to be a **mandatory field** users _had_ to select, so I'd rather prefer the top option to be an unselectable _blank_ option

Reception123:
> I originally wanted this to be a **mandatory field** users _had_ to select, so I'd rather prefer the top option to be an unselectable _blank_ option

But what if none of the purposes match?

Reception123:
> > I originally wanted this to be a **mandatory field** users _had_ to select, so I'd rather prefer the top option to be an unselectable _blank_ option
> 
> But what if none of the purposes match?

@dmehus you never answered

PR #3831, opened 2021-04-12
Do not enable DarkMode by default

R4356th:
WMF Community Tech team abandoned the extension and it is very bad quality-wise.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2573,7 +2573,6 @@ $wi->config->settings += [
 			'categorytree',
 			'cite',
 			'citethispage',
-			'darkmode',
 			'globaluserpage',
 			'minervaneue',
 			'mobilefrontend',


Universal-Omega:
I wanted to do this originally but the idea was opposed to.

RhinosF1 (changes requested):
Blocking for now, I'll check the information I got when I brought issues up last time as there was an attempt to make efforts to improve stuff in core ready for this extension but I'm not sure whether that moved. Please chase me up if I don't look at this on Friday.

R4356th:
@RhinosF1, there should be enough info on WMF Meta.

R4356th:
Found the page- https://meta.wikimedia.org/wiki/Community_Tech/Dark_mode.

RhinosF1:
I just pinged MusikAnimal so let's see if we hear back within a day or so to see if they intend to move with some of their proposed fixes but assuming not then it can probably go.

MusikAnimal:
After much persuasion, the [necessary changes to Core](https://gerrit.wikimedia.org/r/c/mediawiki/core/+/592766/) were made, but I kind of lost steam after that. I hope to return to this soon, but I definitely think it's a good idea to have this extension disabled by default. I was kind of surprised to see it deployed for all users when it was categorized as experimental and clearly unfinished.

That said, patches are most welcome :) Someone came up with a [solution](https://www.mediawiki.org/wiki/Topic:Vqoyyfiw18jjd5sh) to persist the dark mode setting, but apparently [didn't have the time to deal with Gerrit](https://www.mediawiki.org/wiki/Topic:Vd4u0iv65blgjxs2) (understandable!), so a patch was never submitted. If anyone wants to take that on, please feel free. I can review and merge it.

However, as far as I know this will not work as-is with any full-page caching mechanisms such as Varnish. The workaround there is to use JavaScript to add the `.client-dark-mode` CSS class, which will occur after the page is loaded and hence you will see a brief "flash" of non-dark mode. It's my understanding Miraheze uses Varnish even for logged-in users, so this extension even if finished would still have come with its caveats on the Miraheze farm.

RhinosF1 (approved):

R4356th:
Let's wait for https://gerrit.wikimedia.org/r/682134.

R4356th:
Patch merged. Someone needs to update DarkMode.

dmehus:
@R4356th Oh, nice, so you mean the issue(s) with DarkMode have been fixed upstream, so now there's less of a need to not enable it by default?

R4356th:
> 
> 
> @R4356th Oh, nice, so you mean the issue(s) with DarkMode have been fixed upstream, so now there's less of a need to not enable it by default?

DarkMode is now available to logged out users and logged in people get access to a hidden preference that allows toggling dark mode on all pages without clicking anything.

dmehus:
> > @R4356th Oh, nice, so you mean the issue(s) with DarkMode have been fixed upstream, so now there's less of a need to not enable it by default?
> 
> DarkMode is now available to logged out users and logged in people get access to a hidden preference that allows toggling dark mode on all pages without clicking anything.

Ah, that's great.

PR #3838, opened 2021-04-17
Add editextendedconfirmedprotected restrict. level to documentcontrolwiki

dmehus:
Per request (https://meta.miraheze.org/w/index.php?diff=171790&oldid=171789&rcid=614721)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2662,7 +2662,7 @@ $wi->config->settings += [
 				'editextendedconfirmedprotected' => true,
 			],
 		],
-    '+famepediawiki' => [
+    		'+famepediawiki' => [
 			'Extendedconfirmed' => [
 				'editextendedconfirmedprotected' => true,
 			],
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3595,6 +3595,7 @@ $wi->config->settings += [
 			'editinterface',
 		],
 		'+documentcontrolwiki' => [
+			'editextendedconfirmedprotected',
 		],
 		'+famepediawiki' => [
 			'editextendedconfirmedprotected',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3640,6 +3648,9 @@
 			'editrestrictedtemplateprotected',
 			'editimportprotected',
 		],
+		'+documentcontrolwiki' => [
```
joritochip, line 3651: The + is not needed here since it is already present in config name

dmehus: > The + is not needed here since it is already present in config name

@joritochip Oh, my understanding is that the `+` defines configuration settings that are in addition to the default configuration, no?

dmehus: @joritochip Did you mean the `+` sign for the `$wgAvailableRights` variable? If _so_, then yeah, I see the other wikis do not have that, but I'm not quite sure. Can you clarify that for me? I have also fixed this with this commit (https://github.com/miraheze/mw-config/pull/3838/commits/84565fcecf05a0d2e1ec586c0f56d028c109b19f)

Universal-Omega: + is used to append to default configuration, using + in default config appends the default key to the actual already set config is my understanding. I think it is needed here.

Reception123 (approved):

Reception123 (commented):
Conflicts need to be fixed

dmehus:
> Conflicts need to be fixed

@Reception123 Why is there a conflict, though? I did it as a separate PR because I didn't want to add to the existing PR as that had a specific title for that wiki. It should just add it to the blank space, no?

dmehus:
> > Conflicts need to be fixed
> 
> @Reception123 Why is there a conflict, though? I did it as a separate PR because I didn't want to add to the existing PR as that had a specific title for that wiki. It should just add it to the blank space, no?

@Reception123 I think this should be resolved now.

Reception123 (approved):

PR #3841, opened 2021-04-18
Add writeapi to * when flow is enabled from ManageWikiExtensions

Universal-Omega:


JohnFLewis:
Isn't writeapi already enabled on * by default?

Universal-Omega:
> Isn't writeapi already enabled on * by default?

I'm not 100% sure. I only know that this is very annoying when you get exceptions in flow when writeapi wasn't enabled on *, so assumed it wasn't by default.

dmehus:
I wonder if the better way to handle this would be to add blacklist `writeapi` to the `*` user group, so it can't be removed? We'd just have to check which wikis //don't// have it, then add it to that group on those wikis.

Universal-Omega:
It doesn't seem like it is by default. Another user just told me of this error, enabling writeapi on * resolved it and they said they never removed it.

paladox:
Default entry:

*:
```
["autocreateaccount","createaccount","edit","createpage","createtalk", "writeapi","viewmywatchlist","editmywatchlist","viewmyprivateinfo", "editmyprivateinfo","editmyoptions","abusefilter-log","abusefilter-log-detail", "abusefilter-view","autocreateaccount","centralauth-merge", "oathauth-enable"]    
```

So we do switch it on by default?

dmehus:
> It doesn't seem like it is by default. Another user just told me of this error, enabling writeapi on * resolved it and they said they never removed it.

I mean, it is the default user right in ManageWikiDefaultPermissions, per [this](https://meta.miraheze.org/wiki/Special:ManageWikiDefaultPermissions/*); so I _assume_ it's the default in MediaWiki core, but perhaps that's all we need to do is add it to the default in `LocalSettings.php` then blacklist it so users cannot remove it from `*`?

dmehus:
> It doesn't seem like it is by default. Another user just told me of this error, enabling writeapi on * resolved it and they said they never removed it.

I mean, it is the default user right in ManageWikiDefaultPermissions, per [this](https://meta.miraheze.org/wiki/Special:ManageWikiDefaultPermissions/*); so I _assume_ it's the default in MediaWiki core, but perhaps that's all we need to do is add it to the default in `LocalSettings.php` then blacklist it so users cannot remove it from `*`?

Universal-Omega:
> > It doesn't seem like it is by default. Another user just told me of this error, enabling writeapi on * resolved it and they said they never removed it.
> 
> I mean, it is the default user right in ManageWikiDefaultPermissions, per [this](https://meta.miraheze.org/wiki/Special:ManageWikiDefaultPermissions/*); so I _assume_ it's the default in MediaWiki core, but perhaps that's all we need to do is add it to the default in `LocalSettings.php` then blacklist it so users cannot remove it from `*`?

I just checked on one of my wikis which I have never changed permissions on. It isn't enabled on *.

JohnFLewis:
https://meta.miraheze.org/wiki/Special:ManageWikiDefaultPermissions/* Has it as default, so either this was added recently or there is a bug.

dmehus:
> > > It doesn't seem like it is by default. Another user just told me of this error, enabling writeapi on * resolved it and they said they never removed it.
> > 
> > 
> > I mean, it is the default user right in ManageWikiDefaultPermissions, per [this](https://meta.miraheze.org/wiki/Special:ManageWikiDefaultPermissions/*); so I _assume_ it's the default in MediaWiki core, but perhaps that's all we need to do is add it to the default in `LocalSettings.php` then blacklist it so users cannot remove it from `*`?
> 
> I just checked on one of my wikis which I have never changed permissions on. It isn't enabled on *.

Well, that's weird, because it _should_ be if it's in ManageWikiDefaultPermissions as my understanding is that ManageWikiDefaultPermissions defines the default user groups and user rights as wikis are created. Is there a conflict between ManageWikiDefaultPermissions and LocalSettings.php somewhere?

Universal-Omega:
One wiki that had the issue was created on Nov 6th 2020. But it was also closed once when that CW issue kept closing wikis. (So did my other one actually, and it's possible others with the issue did also) so maybe that's the cause of its removal? I really don't know, but I doubt it being closed would but just an idea.

dmehus:
> One wiki that had the issue was created on Nov 6th 2020. But it was also closed once when that CW issue kept closing wikis. (So did my other one actually, and it's possible others with the issue did also) so maybe that's the cause of its removal? I really don't know, but I doubt it being closed would but just an idea. Also possible it was with wikis that are originally private?

I'm confused. Are you speculating as to possible causes for why `writeapi` wasn't in the `*` group? Also, which wikis are involved? That might help us try and figure out the issue. Maybe we updated a database table and something got changed at some point, or maybe there's a bug somewhere or a configuration conflict. I'm not sure, as it could be any of those things.

Reception123:
From what I understand (and see) this already seems to be the default, so I don't see why this is necessary. If I have misunderstood feel free to reopen this.

PR #3853, opened 2021-04-24
Disable Wikibase entity search UI by default

R4356th:
There have been a lot of cases where users have enabled the Wikibase extension without a particular reason and ended up being unable to figure out why their search got "broken". I think disabling it by default would be helpful for those users.


paladox:
Needs to also be done inside the MWS file.

R4356th:
> 
> 
> Needs to also be done inside the MWS file.

Done.

Reception123 (approved):

PR #3854, opened 2021-04-24
Remove wmgUseYandexTranslate 

R4356th:
The feature has been broken for a long time because Miraheze's verification key got expired. I have made miraheze/puppet#1746 to remove the key from Puppet.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4043,9 +4043,6 @@ $wi->config->settings += [
 	'wgTranslateTranslationServices' => [
 		'default' => [],
 	],
-	'wmgUseYandexTranslate' => [
-		'default' => false,
-	],
 
 	// Uploads
  	'wmgPrivateUploads' => [
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -209,19 +209,6 @@ if ( !preg_match( '/^(.*).miraheze.org$/', $wi->hostname ) ) {
 		array_merge( $wgUrlShortenerAllowedDomains, [ preg_quote( str_replace( 'https://', '', $wgServer ) ) ] );
 }
 
-// $wmgUseYandexTranslate
-if ( $wmgUseYandexTranslate ) {
-	$wgTranslateTranslationServices['Yandex'] = [
-		'url' => 'https://translate.yandex.net/api/v1.5/tr.json/translate',
-		'key' => $wmgYandexTranslationKey,
-		'pairs' => 'https://translate.yandex.net/api/v1.5/tr.json/getLangs',
-		'timeout' => 3,
-		'langorder' => [ 'en', 'ru', 'uk', 'de', 'fr', 'pl', 'it', 'es', 'tr' ],
-		'langlimit' => 1,
-		'type' => 'yandex',
-	];
-}
-
 // Per-wiki settings
 if ( $wgDBname === 'erislywiki' ) {
 	$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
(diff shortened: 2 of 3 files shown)


Reception123:
@R4356th well why remove it instead of renewing the verification key?

R4356th:
> @R4356th well why remove it instead of renewing the verification key?

@RhinosF1 wanted to look into that a few months ago when I mentioned this but he never followed through AFAIK.

RhinosF1:
> > @R4356th well why remove it instead of renewing the verification key?
> 
> @RhinosF1 wanted to look into that a few months ago when I mentioned this but he never followed through AFAIK.

I believe I emailed the original requestor at the time and if I remember right no one knew where they key came from

MacFan4000:
Looks like if you create an account at https://translate.yandex.com/developers/keys, you can generate a key

Universal-Omega:
> Looks like if you create an account at https://translate.yandex.com/developers/keys, you can generate a key

I don't think it's free.

R4356th:
> 
> 
> Looks like if you create an account at https://translate.yandex.com/developers/keys, you can generate a key

What Universal Omega said.

dmehus:
> > Looks like if you create an account at https://translate.yandex.com/developers/keys, you can generate a key
> 
> I don't think it's free.

While it's not free for our purposes by enabling this via default, there is:

> Individuals can register (https://translate.yandex.com/developers/account) and receive a grant for $75 (5 million characters) to test the Yandex.Translate API after linking a bank card (no charges will be made).

So we could leave it enabled in ManageWiki, but make the variable `managewiki-restricted`, which a Steward could input the user's API key they e-mailed or direct messaged to the Steward, as there may be some wikis that want to still use this feature. Or it may not even need to be `managewiki-restricted`, if we are able to just allow bureaucrats to input their own API key

R4356th:
NB: The visibility should be set to private in that case.

dmehus:
> NB: The visibility should be set to private in that case.

Yeah, that'd be fine.

Reception123:
First of all, why use `managewiki-restricted` when we could instead use the same system that we use for Discord and Slack webhooks? (Now that it's secure that is of course)
Second, my issue is that there isn't an indication that any user wishes to do this, so wouldn't it be preferable to remove it and **if** someone requests it it can be readded back, instead of keeping it in ManageWiki and hope that someone requests it (which I also don't find very likely since no one has complained about it not working so far). 
Also, there might be an alternative to Yandex as well but I don't quite remember

R4356th:
> 
> 
> First of all, why use `managewiki-restricted` when we could instead use the same system that we use for Discord and Slack webhooks? (Now that it's secure that is of course)

Where does that question come from? I have proposed using visibility.

> Second, my issue is that there isn't an indication that any user wishes to do this, so wouldn't it be preferable to remove it and **if** someone requests it it can be readded back, instead of keeping it in ManageWiki and hope that someone requests it (which I also don't find very likely since no one has complained about it not working so far).

Again, that is what I am proposing?

> Also, there might be an alternative to Yandex as well but I don't quite remember

Yes, see https://github.com/wikimedia/mediawiki-extensions-Translate/blob/master/extension.json#L1016.

dmehus:
@Reception123 Can we see how many wikis are using this `$wmgUseYandexTranslate` variable, perhaps? If it's only a handful of wikis, then yeah, I guess we could probably remove this from ManageWiki, but if there's still quite a few wikis (i.e., more than a dozen or so), then my concern would be our `LocalSettings.php` file would become quite bloated, listing every wiki that wishes to use Yandex Translate

Universal-Omega (approved):
I think this should be done.

Universal-Omega:
Merging with approval from @Reception123 via Discord.

PR #3858, opened 2021-04-28
Add wgScratchBlocks4BlockVersion to ManageWiki

R4356th:
Adds $wgScratchBlocks4BlockVersion to ManageWiki.


ManageWikiSettings.php
```diff
@@ -2617,6 +2617,15 @@
 		'help' => 'When enabled, this shows Popups by default.',
 		'requires' => [],
 	],
+	'wgScratchBlocks4BlockVersion' => [
+		'name' => 'Default Scratch Blocks version',
+		'from' => 'scratchblocks',
+		'type' => 'integer',
```
Universal-Omega, line 2623: This should be float if it's a decimal I think

Universal-Omega (changes requested):

ManageWikiSettings.php
R4356th: Both are supported by the extension.

Universal-Omega: Then it should be set to just 3 at default.

LocalSettings.php
```diff
@@ -3746,6 +3746,11 @@
 			"*",
 		],
 	],
+	
+	// ScratchBlocks
+	'wgScratchBlocks4BlockVersion' => [
+		'default' => '3.0',
```
Universal-Omega, line 3752: This should be int format or MWS type changed to text and default changed to string.

Universal-Omega (changes requested):

ManageWikiSettings.php
R4356th: I set this to 3 at first as that is what anybody familiar with the extension should expect to see but this is minor so done.

LocalSettings.php
R4356th: Done.

LocalSettings.php
```diff
@@ -3746,6 +3746,11 @@
 			"*",
 		],
 	],
+	
+	// ScratchBlocks
+	'wgScratchBlocks4BlockVersion' => [
+		'default' => '3',
```
Universal-Omega, line 3752: Remove the quotes, this just needs to be int it seems as that is what default in MWS is.

Universal-Omega (changes requested):

LocalSettings.php
R4356th: Done.

R4356th:
Blocked on miraheze/mediawiki#1544.

Universal-Omega:
I updated ScratchBlocks4. Will merge this soon.

Universal-Omega (approved):

PR #3859, opened 2021-04-28
Add Sitenotice about Security Incident

RhinosF1:


Sitenotice.php
```diff
@@ -24,19 +24,19 @@ function onSiteNoticeAfter( &$siteNotice, $skin ) {
 	}
 } */
 
-/*
+
 // Specific wiki sitenotice
-if ( $wmgUseApprovedRevs && $wmgUseMyVariables ) {
+if ( $wgDiscordIncomingWebhookUrl && $wgSlackIncomingWebhookUrl ) {
```
Universal-Omega, line 29: Should be ||

Universal-Omega (changes requested):

PR #3862, opened 2021-04-28
Change html attributes to prevent Google crawling sitenotice

revi:
This ensures Google does not use the sitenotice as a snippet for their search result for Miraheze wikis who has not set one via extensions like WikiSEO. [Plaintexts here will be impacted](https://phabricator.miraheze.org/F1430638). Please ensure all future sitenotices and centralnotices has this value.

Cf. https://revi.kr/6bFF ( https://developers.google.com/search/docs/advanced/robots/robots_meta_tag?hl=en#data-nosnippet-attr )

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -24,7 +24,8 @@ $wgMajorSiteNoticeID = 56;
 	}
 } */
 
-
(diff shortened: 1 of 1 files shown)


dmehus:
Great idea, @revi! Thank you. :+1:

PR #3863, opened 2021-04-29
Use Special:MyLanguage links where possible

Universal-Omega:


dmehus:
Some of this looks fine, but I have two concerns:

1. Why remove `https://www.mediawiki.org/wiki/Skin:` from the `// Liberty`  comment? I feel this provides good attribution in a hidden comment to the skin's creator
2. The Tech: namespace pages don't need Special:MyLanguage links because they either aren't or don't need to be translated

Universal-Omega:
> Some of this looks fine, but I have two concerns:
> 
> 1. Why remove `https://www.mediawiki.org/wiki/Skin:` from the `// Liberty`  comment? I feel this provides good attribution in a hidden comment to the skin's creator
> 2. The Tech: namespace pages don't need Special:MyLanguage links because they either aren't or don't need to be translated

I removed the liberty comment because no other config does that, so done to be just like all others as for tech namespaces, that's true, can revert that.

dmehus:
> > Some of this looks fine, but I have two concerns:
> > 
> > 1. Why remove `https://www.mediawiki.org/wiki/Skin:` from the `// Liberty`  comment? I feel this provides good attribution in a hidden comment to the skin's creator
> > 2. The Tech: namespace pages don't need Special:MyLanguage links because they either aren't or don't need to be translated
> 
> I removed the liberty comment because no other config does that, so done to be just like all others as for tech namespaces, that's true, can revert that.

Okay, yeah that's fine, but my preference for the Liberty skin one would be to instead add comments for the other configs add URLs to the other skins/extensions, if possible

Universal-Omega:
> > > Some of this looks fine, but I have two concerns:
> > > 
> > > 1. Why remove `https://www.mediawiki.org/wiki/Skin:` from the `// Liberty`  comment? I feel this provides good attribution in a hidden comment to the skin's creator
> > > 2. The Tech: namespace pages don't need Special:MyLanguage links because they either aren't or don't need to be translated
> > 
> > 
> > I removed the liberty comment because no other config does that, so done to be just like all others as for tech namespaces, that's true, can revert that.
> 
> Okay, yeah that's fine, but my preference for the Liberty skin one would be to instead add comments for the other configs add URLs to the other skins/extensions, if possible

And that might be able to be done, but not right now, it's not essential, and that may further complicate LS file. But it can probably be done later.

dmehus:
> > > > Some of this looks fine, but I have two concerns:
> > > > 
> > > > 1. Why remove `https://www.mediawiki.org/wiki/Skin:` from the `// Liberty`  comment? I feel this provides good attribution in a hidden comment to the skin's creator
> > > > 2. The Tech: namespace pages don't need Special:MyLanguage links because they either aren't or don't need to be translated
> > > 
> > > 
> > > I removed the liberty comment because no other config does that, so done to be just like all others as for tech namespaces, that's true, can revert that.
> > 
> > 
> > Okay, yeah that's fine, but my preference for the Liberty skin one would be to instead add comments for the other configs add URLs to the other skins/extensions, if possible
> 
> And that might be able to be done, but not right now, it's not essential, and that may further complicate LS file. But it can probably be done later.

Okay, cool, sounds good. Yeah, rest looks good to me. :+1:

Reception123 (approved):

PR #3864, opened 2021-04-29
Fix case

dmehus:
One more fix to this sitenotice. The other thing I was wondering if for `Special:ManageWiki/settings`, I'm wondering if we could make that into a link with `<a href="https:" + $wgServer + "/wiki/Special:ManageWiki/settings">Special:ManageWiki/settings</a>`? I just never know if we need a `+` before and/or after the `$wgServer` variable?

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -34,7 +34,7 @@ function onSiteNoticeAfter( &$siteNotice, $skin ) {
 
 	$siteNotice .= <<<EOF
 			<div data-nosnippet><table class="wikitable" style="text-align:center;"><tbody><tr>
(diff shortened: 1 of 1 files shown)


Universal-Omega:
> One more fix to this sitenotice. The other thing I was wondering if for `Special:ManageWiki/settings`, I'm wondering if we could make that into a link with `<a href="https:" + $wgServer + "/wiki/Special:ManageWiki/settings">Special:ManageWiki/settings</a>`? I just never know if we need a `+` before and/or after the `$wgServer` variable?

Wouldn't be `+` I don't think, I believe it would be `.`

dmehus:
> > One more fix to this sitenotice. The other thing I was wondering if for `Special:ManageWiki/settings`, I'm wondering if we could make that into a link with `<a href="https:" + $wgServer + "/wiki/Special:ManageWiki/settings">Special:ManageWiki/settings</a>`? I just never know if we need a `+` before and/or after the `$wgServer` variable?
> 
> Wouldn't be `+` I don't think, I believe it would be `.`

@Universal-Omega Oh, cool, thanks...yeah I was just thinking of the syntax from [Iptscrae](https://en.wikipedia.org/wiki/IPTSCRAE), so that's interesting.

PR #3869, opened 2021-05-02
email notifications on privileged actions configuration T7197

Reception123:


LocalSettings.php
```diff
@@ -3684,7 +3685,9 @@
 	
 	// Rights
 	'+wgAvailableRights' => [
-		'default' => [],
+		'default' => [
+			'watched-person',
```
RhinosF1, line 3689: Others then need moving to + codes. I wonder if there's a way to just add straight to the global group.

LocalSettings.php
```diff
@@ -4489,6 +4492,14 @@
 	],
 ];
 
+// Email notifications on privileged actions configuration
+'wgMirahezeMagicLogEmailConditions' => [
+	'default' => [
+		'right' => 'watched-person',
```
supertassu, line 4498: wgMirahezeMagicLogEmailConditions supports multiple watched groups, those values need to be wrapped inside an array

LocalSettings.php
dmehus: If it's in `$wgAvailableRights` for `default`, it should be possible to then just add that user right in Special:GlobalGroupPermissions, I believe

LocalSettings.php
```diff
@@ -4487,6 +4491,16 @@
 			'XMP' => false,
 		],
 	],
+
+// Email notifications on privileged actions configuration
```
Universal-Omega, line 4495: Indentation is off I think? Needs one additional tab on everything it seems. And removal of tab on a bracket.

Reception123: Thanks, will fix that when I change my email to Owen's after testing.

PR #3872, opened 2021-05-05
Add CreatedPagesList (T7236) 

R4356th:
Adds the CreatedPagesList extension per T7236.

``mediawiki`` pull request: miraheze/mediawiki#1603.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -1191,7 +1191,7 @@ $wgManageWikiExtensions = [
 			'section' => 'specialpages',
 			'install' => [
 				'sql' => [
-					'createdpageslist' => '$IP/extensions/CreatedPagesList/sql/patch-createdpageslist.sql'
+					'createdpageslist' => "$IP/extensions/CreatedPagesList/sql/patch-createdpageslist.sql"
 				],
 			],
 		],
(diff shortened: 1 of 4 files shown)


ManageWikiExtensions.php
```diff
@@ -1182,6 +1182,19 @@
...
+		'createdpageslist' => [
+			'name' => 'CreatedPagesList',
+			'linkPage' => 'https://www.mediawiki.org/wiki/Extension:CreatedPagesList',
+			'var' => 'wmgUseCreatedPagesList',
+			'conflicts' => false,
+			'requires' => [],
+			'section' => 'specialpages',
+			'install' => [
+				'sql' => [
+					'createdpageslist' => '$IP/extensions/CreatedPagesList/sql/patch-createdpageslist.sql'
```
Reception123, line 1194: You'll need to use "" instead of '' for the SQL path.

Reception123 (changes requested):

ManageWikiExtensions.php

RhinosF1 (changes requested):
See concerns in -sre around lack of actor usage and why it can't just store page_id

Universal-Omega:
Decided to just decline the extension for now, since I don't currently have time to do upstream patch, if someone wants to then it can be re-evaluated but until then it is declined.

PR #3873, opened 2021-05-05
Test change in FlaggedRevs config for isvwiki

thelevking:
As I've reported earlier, [FlaggedRevs autoreview does not work on isvwiki](https://phabricator.miraheze.org/T7032) despite all the necessary user rights and ManageWiki settings set. Other wikis have no problem with it, however:
1. https://nomenber.miraheze.org/wiki/Special:Log?type=review&user=&page=&wpdate=&tagfilter=&subtype=autoaccept
2. https://excelletdiscordservers.miraheze.org/wiki/Special:Log?type=review&user=&page=&wpdate=&tagfilter=&subtype=autoaccept

(Thanks to Reception123 for compiling the list for me.)

As I don't know the cause for this, I've decided to test this change that makes it as close to default as possible (other than changing the default level). If there are better ways to write this, feel free to change it.

If this does not work, I'll just stop using the extension.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1794,8 +1794,8 @@ $wi->config->settings += [
 		],
 		'isvwiki' => [
 			'status' => [
+				'quality' => 1,
 				'levels' => 1,
-				'quality' => 2,
 				'pristine' => 4,
 			],
 		],


LocalSettings.php
```diff
@@ -1794,8 +1794,8 @@
 		],
 		'isvwiki' => [
 			'status' => [
+				'quality' => 1,
```
R4356th, line 1797: Quality will be removed soon once MediaWiki 1.36 is deployed, FYI. See https://phabricator.wikimedia.org/T277883.

thelevking: https://noc.wikimedia.org/conf/highlight.php?file=flaggedrevs.php still has them while they are at MediaWiki 1.37, so I am not sure about that. Either way, the whole config would need fixing then, I'm just trying to learn if this causes the issue of no autoreview.

Universal-Omega:
This is removed on 1.36, and we will be upgrading to 1.36 pretty soon, so closing this.

PR #3883, opened 2021-05-09
remove watched-person right; unused

The-Voidwalker:


dmehus:
@The-Voidwalker Isn't this still needed for `trustandsafety`, though? Or wait, was this a leftover artefact from when we were going to use user rights but switched to global groups?

dmehus (approved):
LGTM. Per discussion with @RhinosF1 on IRC, I answered my own question, so +1:

PR #3885, opened 2021-05-10
add db server recorder for logging (T7248)

Reception123:
Could be wrong

### DIFF
--- a/GlobalLogging.php
+++ b/GlobalLogging.php
@@ -19,7 +19,7 @@ $wmgMonologProcessors = [
 		'factory' => function () {
 			return function ( array $record ) {
 				global $wgLBFactoryConf, $wgDBname;
-				$record['extra']['shard'] = $wgLBFactoryConf['sectionsByDB'][$wgDBname] ?? 'c3';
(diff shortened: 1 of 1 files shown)


GlobalLogging.php
```diff
@@ -15,6 +15,14 @@
 	'web' => [
 		'class' => \Monolog\Processor\WebProcessor::class,
 	],
+	'mhconfig' => [
+		'factory' => function () {
+			return function ( array $record ) {
+				global $wgLBFactoryConf, $wgDBname;
+				$record['extra']['shard'] = $wgLBFactoryConf['sectionsByDB'][$wgDBname] ?? 'c3';
+
+				return $record;
```
Universal-Omega, line 24: Missing closing }

Universal-Omega (changes requested):

GlobalLogging.php
```diff
@@ -15,6 +15,15 @@
 	'web' => [
 		'class' => \Monolog\Processor\WebProcessor::class,
 	],
+	'mhconfig' => [
+		'factory' => function () {
+			return function ( array $record ) {
+				global $wgLBFactoryConf, $wgDBname;
```
Universal-Omega, line 21: Not sure that this is needed here.

PR #3887, opened 2021-05-10
blacklist individual option rights from user

Reception123:
There's no need/justification for these to be restricted for a user either on a wiki


dmehus (commented):
@Reception123 I don't think this is needed actually. Registered users inherit `*` rights

PR #3890, opened 2021-05-12
Move cache out of /srv/mediawiki/w directory

Universal-Omega:
Per [T7192](https://phabricator.miraheze.org/T7192)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -378,6 +378,31 @@ $wi->config->settings += [
 	'wgCheckUserForceSummary' => [
 		'default' => true,
 	],
+	'wgCheckUserEnableSpecialInvestigate' => [
+		'default' => true,
+	],
+	'wgCheckUserLogLogins' => [
+		'default' => true,
+	],
+	'wgCheckUserCAtoollink' => [
+		'default' => 'metawiki',
+	],
+	'wgCheckUserGBtoollink' => [
+		'default' => [
+			'centralDB' => 'metawiki',
+			'groups' => [
+				'steward',
+			],
+		],
+	],
+	'wgCheckUserCAMultiLock' => [
+		'default' => [
+			'centralDB' => 'metawiki',
+			'groups' => [
+				'steward',
+			],
+		],
+	],
 
 	// Citizen
 	'wgCitizenThemeDefault' => [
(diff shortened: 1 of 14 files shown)


LocalSettings.php
```diff
@@ -774,7 +774,7 @@
 		],
 	],
 	'wgCreateWikiCacheDirectory' => [
-		'default' => '/srv/mediawiki/w/cache'
+		'default' => '/srv/mediawiki/cache'
```
Southparkfan, line 777: 1) Requires a puppet patch to create a writable directory
2) Please do not merge this _before_ denying access to this new folder: https://github.com/miraheze/puppet/blob/master/modules/mediawiki/templates/mediawiki-includes.conf.erb#L11-L13
3) Per the https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard, /var/cache may be the better location?

Southparkfan (changes requested):

LocalSettings.php
Universal-Omega: 1. yep that's why I marked this one as draft for now so it doesn't get merged before that's done.
2. Oh, right, thanks!
3. That actually may be better, yeah. Thanks!

paladox:
We could do /srv/mediawiki/cache (where i18n & MW go).

So /srv/mediawiki/cache/i18n & /srv/mediawiki/managewiki.

JohnFLewis:
> That actually may be better, yeah.

If you think /var/cache as SPF suggested is better, why did you revert that change?

Universal-Omega:
> > That actually may be better, yeah.
> 
> If you think /var/cache as SPF suggested is better, why did you revert that change?

Because I'm not exactly sure what to do. In all honesty after more thought /srv/mediawiki/cache does make more sense here, but I also still want to take SPFs advice here. And there's been some conflicting recommendations and I'm not really sure what to do here.

Universal-Omega:
Puppet PR: miraheze/puppet#1772

Reception123:
Seems fine to me, though would like @The-Voidwalker and/or @Southparkfan to have a final look at this and puppet before we make the move

The-Voidwalker:
Approved from a code perspective. I think everything that would need to be changed was updated. Regarding `/srv/mediawiki/cache` vs `/var/cache/mediawiki`, I don't really have much to say. It makes sense to me to keep all the mediawiki stuff together, and I'd approve as-is unless there's good reason or precedent (in our infrastructure) to do otherwise.

PR #3893, opened 2021-05-13
#T7283 - add protection level to allpediawiki

dmehus:
Add `extendedconfirmed` protection level to `allpediawiki` per [T7283](https://phabricator.miraheze.org/T7283)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2655,6 +2655,11 @@ $wi->config->settings += [
 				'userrights' => true,
 			],
 		],
+    		'+allpediawiki' => [
+			'extendedconfirmed' => [
+				'editextendedconfirmedprotected' => true,
+			],
+		],
 		'+autocountwiki' => [
 			'authors' => [
 				'torunblocked' => true,
@@ -3657,6 +3662,11 @@ $wi->config->settings += [
 			'autoconfirmed',
 			'sysop'
 		],
+    		'+allpediawiki' => [
+			'extendedconfirmed' => [
+				'editextendedconfirmedprotected' => true,
+			],
+		],
 		'+ahinfoboxeswiki' => [
 			'editrollbackprotected',
 			'edittemplateprotected',
@@ -3741,6 +3751,11 @@ $wi->config->settings += [
 	// Rights
 	'+wgAvailableRights' => [
 		'default' => [],
+    		'+allpediawiki' => [
+			'extendedconfirmed' => [
+				'editextendedconfirmedprotected' => true,
+			],
+		],
 		'ahinfoboxeswiki' => [
 			'editrollbackprotected',
 			'edittemplateprotected',


LocalSettings.php
```diff
@@ -3657,6 +3662,11 @@
 			'autoconfirmed',
 			'sysop'
 		],
+    		'+allpediawiki' => [
```
Universal-Omega, line 3665: Should be ordered alphabetically

LocalSettings.php
```diff
@@ -3741,6 +3751,11 @@
 	// Rights
 	'+wgAvailableRights' => [
 		'default' => [],
+    		'+allpediawiki' => [
```
Universal-Omega, line 3754: Should be ordered alphabetically

Universal-Omega (changes requested):

LocalSettings.php
Universal-Omega: Also doesn't need the `+`

LocalSettings.php
dmehus: > Should be ordered alphabetically

It is. The only one above it is `default`, and we always sort `default` at the top.

Universal-Omega (changes requested):

LocalSettings.php
dmehus: > Should be ordered alphabetically

It is, `default` is always sorted above the optional wiki configurations. See prior revision.

LocalSettings.php
Universal-Omega: Should be moved down below ahinfoboxeswiki I think

LocalSettings.php
Universal-Omega: Below ahinfoboxeswiki

LocalSettings.php
dmehus: @Universal-Omega Oh. I see what you mean. Sorry, thought I had it alphabetized, but yeah I _do_ think `default` should be at the top.

LocalSettings.php
Universal-Omega: Also doesn't need the `+` here.

LocalSettings.php
dmehus: Done. Just waiting on CI to finish.

LocalSettings.php
dmehus: Done

LocalSettings.php
```diff
@@ -3747,6 +3757,11 @@
 			'editrestrictedtemplateprotected',
 			'editimportprotected',
 		],
+    		'+allpediawiki' => [
```
Universal-Omega, line 3760: Doesn't need `+`

Universal-Omega (changes requested):

LocalSettings.php
dmehus: Done. Just waiting on CI to conclude.

LocalSettings.php
```diff
@@ -3747,6 +3757,11 @@
 			'editrestrictedtemplateprotected',
 			'editimportprotected',
 		],
+    		'allpediawiki' => [
```
Universal-Omega, line 3760: This is the wrong array syntax for here. Would  drop the first array level, see others for example.

Universal-Omega (changes requested):

LocalSettings.php
dmehus: @Universal-Omega Huh? This is the way I've done it before, three times, following others' lead, and Reception123's merged them each time? I believe this is correct.

Universal-Omega: There is no other like this way.

dmehus: > There is no other like this way. Not for this config.

If this is the bottom variable, yes, they're all like that.

Universal-Omega:
```php
'famepediawiki' => [	'editextendedconfirmedprotected',
'edittemplateprotected',
],
``` snippet of another. Which doesn't match yours.

dmehus: @Universal-Omega Yeah I see it. That's a copy/paste mistake on my part. My bad. ;)

dmehus: Should be done now. :)

Universal-Omega (approved):

PR #3895, opened 2021-05-13
Add Wikimedia Meta to import sources

BukkitIsEnchanted:
Addition of Meta Wikimedia for imports


dmehus (approved):
:+1: LGTM. I was actually going to add to this to the `default` import sources, as, similar to MediaWiki.org and English Wikipedia, Meta Wikimedia wiki is a widely used source of templates (arguably better templates than English Wikipedia)

PR #3897, opened 2021-05-13
No longer manually set wmgUseExtension in LocalSettings

Universal-Omega:
Decided to revert #3878 and do it this way instead as I was unsure how performance would be effected afterall, however went ahead and made this PR but do not intend on merging it just yet, so Reception123 can review it, and make sure we are not complicating matters further.


JohnFLewis:
The reduction in lines by this method I wouldn’t advise. Either the full system should stay in config or be entirely moved to WikiInitialise which is what the tags commit I did months ago was designed for with further changes.

My advice is don’t merge for a cheap reduction in file size as you’ll just cause problems later on rather than solutions and easing of the process.

Universal-Omega:
> The reduction in lines by this method I wouldn’t advise. Either the full system should stay in config or be entirely moved to WikiInitialise which is what the tags commit I did months ago was designed for with further changes.
> 
> My advice is don’t merge for a cheap reduction in file size as you’ll just cause problems later on rather than solutions and easing of the process.

Thank you, and you make a great point. Will close.

PR #3906, opened 2021-05-15
T7315: Move GlobalWatchlist to loginwiki

Universal-Omega:


dmehus:
Yes, thank you for https://github.com/miraheze/mw-config/compare/306cb254279e...d783a2d8eb9c, @Universal-Omega. I wondered whether it was necessary to enable on `metawiki`. As it's not a global extension, it seems like something wikis should decide whether to enable individually. :+1:

Universal-Omega:
> Yes, thank you for [306cb25...d783a2d](https://github.com/miraheze/mw-config/compare/306cb254279e...d783a2d8eb9c), @Universal-Omega. I wondered whether it was necessary to enable on `metawiki`. As it's not a global extension, it seems like something wikis should decide whether to enable individually. 👍

This extension can only be enabled on one wiki. So for now at least is loginwiki.

dmehus:
> > Yes, thank you for [306cb25...d783a2d](https://github.com/miraheze/mw-config/compare/306cb254279e...d783a2d8eb9c), @Universal-Omega. I wondered whether it was necessary to enable on `metawiki`. As it's not a global extension, it seems like something wikis should decide whether to enable individually. 👍
> 
> This extension can only be enabled on one wiki. So for now at least is loginwiki.

Yeah, that's true, too.

PR #3926, opened 2021-05-28
Set WebChat to default to Libera; remove freenodeChat

Universal-Omega:


R4356th:
Not everyone is moving to Libera. I feel like you are forcing people to use a platform SRE thinks is best for them, not one _they_ find the best.

Universal-Omega:
> Not everyone is moving to Libera. I feel like you are forcing people to use a platform SRE thinks is best for them, not one _they_ find the best.

This just changes the default, Freenode may be removed entirely from both this and CSP. The community gave their input on Freenode, and Freenode has demonstrated their recent untrustworthyness. It's starting to become a privacy risk even. On the May 25th, they changed their policy and hours later acted upon their changes, taking control of countless channels (including some private ones). This type of action demonstrated their willingness to do what they want and change policy very suddenly and act upon it very soon after. I'm not convinced that wouldn't apply to something like their privacy policy either.

dmehus:
> > Not everyone is moving to Libera. I feel like you are forcing people to use a platform SRE thinks is best for them, not one _they_ find the best.
> 
> This just changes the default, Freenode may be removed entirely from both this and CSP. The community gave their input on Freenode, and Freenode has demonstrated their recent untrustworthyness. It's starting to become a privacy risk even. On the May 25th, they changed their policy and hours later acted upon their changes, taking control of countless channels (including some private ones). This type of action demonstrated their willingness to do what they want and change policy very suddenly and act upon it very soon after. I'm not convinced that wouldn't apply to something like their privacy policy either.

I disagree here. The community did enforce moving to Libera.Chat, and for those wikis that want to use Freenode as their webchat, a local override can be set, either in `LocalSettings.php` or ManageWiki.

dmehus:
> > Not everyone is moving to Libera. I feel like you are forcing people to use a platform SRE thinks is best for them, not one _they_ find the best.
> 
> This just changes the default, Freenode may be removed entirely from both this and CSP. The community gave their input on Freenode, and Freenode has demonstrated their recent untrustworthyness. It's starting to become a privacy risk even. On the May 25th, they changed their policy and hours later acted upon their changes, taking control of countless channels (including some private ones). This type of action demonstrated their willingness to do what they want and change policy very suddenly and act upon it very soon after. I'm not convinced that wouldn't apply to something like their privacy policy either.

That's interesting, though, about the Content Security Policy. I could definitely see that happening, and hadn't thought of that.

dmehus:
On the other hand, shouldn't we request this be changed as the default upstream? I'm sure the extension's developers would support this.

R4356th:
The community endorsed moving Miraheze's channels not the thousands of local wiki communities' channels. SRE [promised](https://meta.miraheze.org/w/index.php?title=IRC/RfC&diff=prev&oldid=178131&diffmode=source) to discuss these kind of changes with the community in the future but it looks like it was only made to be forgotten.

Universal-Omega:
> The community endorsed moving Miraheze's channels not the thousands of local wiki communities' channels. SRE [promised](https://meta.miraheze.org/w/index.php?title=IRC/RfC&diff=prev&oldid=178131&diffmode=source) to discuss these kind of changes with the community in the future but it looks like it was only made to be forgotten.

That is true. We probably should discuss before completely dropping FN (unless it does truly become a privacy issue), but not as needed before changing the default. (in my mind)

Reception123:
I understand your frustration, but indeed, we are only changing the **default** we're not forcing anyone to choose Libera. I'd also like to mention that the characterization of '_thousands_ of local wiki communities' channels' isn't correct as in practice very few wikis do actually use WebChat. We have not at all forgotten our statement, but it is up to each wiki to decide which network they use, all we are really doing is setting the default which is easily changeable.

dmehus:
> I understand your frustration, but indeed, we are only changing the **default** we're not forcing anyone to choose Libera. I'd also like to mention that the characterization of '_thousands_ of local wiki communities' channels' isn't correct as in practice very few wikis do actually use WebChat. We have not at all forgotten our statement, but it is up to each wiki to decide which network they use, all we are really doing is setting the default which is easily changeable.

Yeah, we're definitely not prohibiting using Freenode. At present, there's no need to remove `freenode.net` from the Content Security Policy, though if there were a security- or privacy-related reason for doing so, then that would be SRE's decision alone.

R4356th:
> 
> We have not at all forgotten our statement, but it is up to each wiki to decide which network they use, all we are really doing is setting the default which is easily changeable.

Thank you. The issue is that not all wikis have active admins and the active users may not know whom to contact to change these. Maybe you could set up a sitenotice for those wikis which are using this extension?

dmehus:
> > We have not at all forgotten our statement, but it is up to each wiki to decide which network they use, all we are really doing is setting the default which is easily changeable.
> 
> Thank you. The issue is that not all wikis have active admins and the active users may not know whom to contact to change these. Maybe you could set up a sitenotice for those wikis which are using this extension?

I'd definitely support a specific wiki sitenotice. I think that could be quite useful. What about also adding this variable to Special:ManageWiki/settings?

Universal-Omega:
Marked as draft since this is waiting on CSP.

dmehus:
LGTM now. Up to you if you want to still do a specific wiki sitenotice, but I can't imagine we have many, if _any_, **active** wikis still using Freenode as their chat network with this extension.

Universal-Omega:
> LGTM now. Up to you if you want to still do a specific wiki sitenotice, but I can't imagine we have many, if _any_, **active** wikis still using Freenode as their chat network with this extension.

We just removed the CSP so it doesn't work anymore.

dmehus:
> > LGTM now. Up to you if you want to still do a specific wiki sitenotice, but I can't imagine we have many, if _any_, **active** wikis still using Freenode as their chat network with this extension.
> 
> We just removed the CSP so it doesn't work anymore.

True, and they can always file a Phabricator task to request that a _new_ IRC chat network, being `*.freenode.net` be added to the Content Security Policy whitelist, subject to SRE's approval process therein.

JohnFLewis:
Where was this communicated to the communities using the extension/configuration?

Universal-Omega:
> Where was this communicated to the communities using the extension/configuration?

It was not. I was not even going to do this until we communicated this with communities, but I was asked to change my puppet PR to remove the FN whitelist as it has become a privacy risk, so did, then did this to provide a working extension.

JohnFLewis:
No communication for a break change - extremely poor conduct from SRE. The conduct SRE are exhibiting recently is dangerous in that there is a consistent disregard of the community in decision making with SRE acting like they own the service rather than support the service.

dmehus:
I do agree here that the continuing events from Freenode, which continue to rapidly evolve, have necessitated a "remove first, then communicate." Communication can, and should, be done, but there's nothing saying it can't be communicated _after_ implementation, where the circumstances warrant it, no?

JohnFLewis:
The circumstances do not warrant it here. Users opting to use a service, optionally where transmission of data is opt in. Miraheze aren’t forcing users to use Freenode in this regard.

dmehus:
> The circumstances do not warrant it here. Users opting to use a service, optionally where transmission of data is opt in. Miraheze aren’t forcing users to use Freenode in this regard.

Well, I suppose we could've left `*.freenode.net` as part of the Content Security Policy, but @paladox said he preferred to _replace_, rather than _add to_, the existing CSP. Do you suggest we add `*.freenode.net` as a new link then?

JohnFLewis:
I suggest the community be involved.

I will consider a RfC for a vote of no confidence in SRE in the coming days as SRE have given numerous assurances and all assurances have been disregarded.

dmehus:
> I suggest the community be involved.
> 
> I will consider a RfC for a vote of no confidence in SRE in the coming days as SRE have given numerous assurances and all assurances have been disregarded.

I mean, I _was_ contemplating starting a second IRC-related RfC on the matter to suggest this change. We could still revert the changes and start an RfC, or we could just do an RfC endorsing/ratifying this change. If it fails, the changes would be reverted?

JohnFLewis:
If that’s the route you’d want to take, feel free. I’m taking my route as this is consistent and SRE don’t seem to taking any concerns raised seriously

Southparkfan:
I am aware most projects have moved to Libera Chat, but let's not remove the ability to use freenode chat please.

PR #3928, opened 2021-05-28
Enable MP3 for my site using TimedMediaHandler

Njsmooth:
As the title says. Also, this option has not been implemented into managewiki (mp4 has), so this is the only way for me to enable this setting. Maybe consider adding it to managewiki?

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4115,6 +4115,10 @@ $wi->config->settings += [
 	'wgTmhEnableMp4Uploads' => [
 		'default' => false,
 	],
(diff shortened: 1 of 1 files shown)


Universal-Omega:
Going to go ahead and add this to ManageWiki probably tomorrow.

Njsmooth:
@Universal-Omega I made a mistake! Apparently MP3 is enabled by default, I just goofed and didn't realize it was. I'll make another pull request removing that setting I added, so don't worry about editing managewiki. My apologies!

PR #3929, opened 2021-05-29
Blacklist centralauth-createlocal right

Universal-Omega:


dmehus:
:+1: from me as I actually talked to @Universal-Omega and @Reception123 about this. As we can still add this user right to the global `steward` group in Special:GlobalGroupPermissions, this should be fine (since the special page is used locally, not on Meta).

dmehus (commented):
I assume this is needed because it's defaulted to `sysop`? If so, that's fine. Why is the `user` change needed, though?

dmehus (approved):

dmehus (commented):
Actually, now that I think about it, I'd rather remove `centralauth-createlocal` from the local `steward` user group, and have us just add it to the global `steward` group. It's not necessary for us to have to add local `steward` just to use this special page, though it's kind of harmless being in here also, so I won't request changes.

LocalSettings.php
```diff
@@ -2659,15 +2659,19 @@
...
-				'user' => true,
-			],
 			'steward' => [
+				'centralauth-createlocal' => true,
 				'centralauth-usermerge' => true,
 				'usermerge' => true,
 				'userrights' => true,
 			],
+			'sysop' => [
+				'centralauth-createlocal' => false,
```
JohnFLewis, line 2669: => false == do nothing at all

Universal-Omega: Oh. Thanks! I'll have to use WgRevokePermissions then? Or wgGroupPermissions directly instead of this?

JohnFLewis: Neither? What are you trying to achieve?

Universal-Omega: Revoke it from sysop since the extension adds it to sysop by default, blacklisting it won't matter since sysop would still have it.

PR #3937, opened 2021-06-02
Creation of SECURITY.md

BukkitIsEnchanted:
Although the policy is listed in the README.md, I believe SECURITY.md should be added


Universal-Omega:
I don't necessarily see the need for this with it being in README.

BukkitIsEnchanted:
> I don't necessarily see the need for this with it being in README.

True, I'll close this.

PR #3938, opened 2021-06-03
Configure wgRottenLinksExcludeWebsites for snapwikiwiki

R4356th:
`smerge.imp.fu-berlin.de` and `www.hinkler.com.au` are way too slow in responding to requests.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3830,6 +3830,7 @@ $wi->config->settings += [
 		'snapwikiwiki' => [
 			'smerge.imp.fu-berlin.de',
 			'www.hinkler.com.au',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3827,6 +3827,10 @@
 			'localhost',
 			'127.0.0.1',
 		],
+		'snapwikiwiki' => [
```
Universal-Omega, line 3830: Add `+` to include `localhost` and `127.0.0.1`

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #3942, opened 2021-06-08
Add CountDownClock

R4356th:
Per T7426.

### DIFF
--- a/extension-list
+++ b/extension-list
@@ -267,6 +267,7 @@ $IP/extensions/WikidataPageBanner/extension.json
 $IP/extensions/YouTube/extension.json
 $IP/extensions/cldr/extension.json
 $IP/extensions/wikihiero/extension.json
+$IP/extensions/CountDownClock/extension.json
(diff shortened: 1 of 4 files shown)


extension-list
```diff
@@ -267,6 +267,7 @@ $IP/extensions/WikidataPageBanner/extension.json
 $IP/extensions/YouTube/extension.json
 $IP/extensions/cldr/extension.json
 $IP/extensions/wikihiero/extension.json
+$IP/extensions/CountDownClock/extension.json
```
Universal-Omega, line 270: Should be done alphabetically

Universal-Omega (changes requested):

extension-list
R4356th: The extensions above it are not alphabetically though. The whole file needs work.

Universal-Omega: Yes they are. This is also case alphabetical. Capital then lowercase at the bottom.

PR #3945, opened 2021-06-09
Set wgMiserMode to true for opt-in wikis

Universal-Omega:


JohnFLewis:
How was metawiki opted in out of interest? I thought SPF said this could be done with community agreement only? As a bureaucrat I wasn’t aware this was happening and neither was the community to my knowledge.

Universal-Omega:
> How was metawiki opted in out of interest? I thought SPF said this could be done with community agreement only? As a bureaucrat I wasn’t aware this was happening and neither was the community to my knowledge.

I apologise that you were not aware of this. That was an oversight on my part. This was opted in through approval from dmehus as Steward and Reception123 as meta bureaucrat. Again, I apologise you were not aware of this prior as you should've been. Additionally I do not recall SPF saying this needed community census to do this. However if I missed that or am mistaken then once again I take responsibility for that mistake and apologise. Should metawiki be removed from the opt-in? We were currently planning to do these opt in wikis and then after a week or two if performance seems better it will be made global, if it seems less or the same it'll be completely reverted. That was the idea SPF brought up.

JohnFLewis:
Doug doesn’t have the ability to approve metawiki changes at all as a Steward.

it’s fine, but this is yet again another example of SRE taking little regard for the communities independence. This seriously needs to start happening otherwise SRE will lose faith of the community if it continually takes actions to oppress the independence by forcing changes on them.

dmehus:
> Doug doesn’t have the ability to approve metawiki changes at all as a Steward.
> 
> it’s fine, but this is yet again another example of SRE taking little regard for the communities independence. This seriously needs to start happening otherwise SRE will lose faith of the community if it continually takes actions to oppress the independence by forcing changes on them.

Meta is definitely a bit confusing jurisdictionally. Best would've been for SRE to have obtained your approval as you're both a Meta bureaucrat and a steward. I did suggest that they have SPF //and// Reception123's approval //or// your approval, and also consult Stewards vis-a-vis Meta's role as a central coordination and global wiki.

I really didn't see the rush to approve this and was a bit surprised it was merged so quickly after the MediaWiki upgrade was done.

I agree it's likely fine and minor, so probably not necessary to revert, but SRE should use this as a further example in their review of their workflows.

Reception123:
I confirm that I did indeed give my approval for this change as I thought it was relatively minor and didn't really affect the workflow of the wiki in a way that would require a discussion, and was more of a 'technical change' rather than a 'user facing change' given its ultimate goal. In the future if you would prefer I'd be happy to also consult another Meta bureaucrat before approving such changes.

As SRE I would add that I don't feel SRE was wrong here since they did get a perceived approval from me who was a Meta bureaucrat. If someone went wrong in this issue it would be me in my capacity as bureaucrat rather than SRE/Universal Omega .

dmehus:
> I confirm that I did indeed give my approval for this change as I thought it was relatively minor and didn't really affect the workflow of the wiki in a way that would require a discussion, and was more of a 'technical change' rather than a 'user facing change' given its ultimate goal. In the future if you would prefer I'd be happy to also consult another Meta bureaucrat before approving such changes.
> 
> As SRE I would add that I don't feel SRE was wrong here since they did get a perceived approval from me who was a Meta bureaucrat. If someone went wrong in this issue it would be me in my capacity as bureaucrat rather than SRE/Universal Omega .

I definitely agree this change wouldn't require a community discussion or RfC, per the reason(s) you articulated and also because it's such a highly technical change that most of the community would not even fully understand. I admit myself that I don't even fully understand it as well. I also think Stewards should be consulted, though, as Meta is a shared jurisdiction between Meta bureaucrats and Stewards as a result of the pan-Miraheze role Meta plays, with Meta bureaucrats responsible for determining consensus of local Meta discussions and permissions requests, and Stewards responsible for effecting certain technical changes to which Meta bureaucrats aren't capable of effecting because of the pan-Miraheze role Meta plays. In this case, I agree it's largely Meta bureaucrats that should be consulted, to determine if a discussion is required, as I'm not aware what, if any, impacts that setting change would have on counter-vandalism-related extensions enabled, or potentially enabled, on Meta Wiki.

JohnFLewis:
SPF said this should only have wikis opted in where the community are aware and have given consent - it’s clear the Meta community wasn’t aware this change was done as I only found out about it via a commit after deployment.

SRE have certainly done wrong in that they consulted the wrong group for a community change (a steward for meta, when Stewards are global not local) and failed to notify/inform the community about a trial which the DSRE said must be done.

dmehus:
> SPF said this should only have wikis opted in where the community are aware and have given consent - it’s clear the Meta community wasn’t aware this change was done as I only found out about it via a commit after deployment.

That's a fair point. While the community may not need to be consulted collectively via a broad community discussion, in this case, given that Reception123 is also SRE, perhaps there was a bit of a role conflict in Universal Omega asking for his approval and, thus, it would've been preferable for you to provide the determination on whether this was non-controversial enough to not warrant a community discussion?

> If SPF didn’t say this, why is this an opt in and not a global change? And what was the criteria for opt in?

I think this was meant to be a _trial_, with the view that it would be implemented globally (hence another reason why I feel Stewards should've been consulted, too, since this was only an _interim_ step).

> SRE have certainly done wrong in that they consulted the wrong group for a community change (a steward for meta, when Stewards are global not local) and failed to notify/inform the community about a trial which the DSRE said must be done.

That's a bit fair, I think, particularly when you consider my reply to your first point in this message.

PR #3948, opened 2021-06-10
1.36 sitenotice

MacFan4000:
per discussed day and time in -sre


dmehus (commented):
:+1:

R4356th (commented):
You did not increment the sitenotice ID.

PR #3956, opened 2021-06-12
DT sitenotice

MacFan4000:


dmehus:
@MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?

MacFan4000:
> @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?

I honestly don't see a problem having two sitenotices, plus it's only on certain wikis

dmehus:
> > @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?
> 
> I honestly don't see a problem having two sitenotices, plus it's only on certain wikis

They're very ugly and it creates visual clutter. It's really better to shorten it and combine it with sitenotices, or just not do this sitenotice at all.

ugochimobi (approved):

ugochimobi:
> > @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?
> 
> I honestly don't see a problem having two sitenotices, plus it's only on certain wikis
> 
> And yes, I saw the conversation

> There's actually no problem in having 2 sitenotice but when it go live it cause or probably occupies too many visual spaces on wikis. I think you would've seen a situation where you have your own sitenotice enabled and Miraheze has the Central notice enabled as well.

MacFan4000:
> > > @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?
> 
> > There's actually no problem in having 2 sitenotice but when it go live it cause or probably occupies too many visual spaces on wikis. I think you would've seen a situation where you have your own sitenotice enabled and Miraheze has the Central notice enabled as well.

Central notice would be hard as you would have to manually build a list of wikis with the extension enabled.

dmehus:
> > > > @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?
> > 
> > 
> > > There's actually no problem in having 2 sitenotice but when it go live it cause or probably occupies too many visual spaces on wikis. I think you would've seen a situation where you have your own sitenotice enabled and Miraheze has the Central notice enabled as well.
> 
> Central notice would be hard as you would have to manually build a list of wikis with the extension enabled.

Yeah, good point. I did think about a central notice actually, but that was one of the reasons I didn't suggest it. Not ruling it out, but it'd just be quite tedious

ugochimobi:
> > > > @MacFan4000 @Reception123 and I discussed in `#miraheze-sre` that we could just shorten this a bit and combine it with the existing global sitenotice, as otherwise wikis will have two sitenotices concurrently, which is sub-optimal. Can we close this and instead update the existing sitenotice?
> > 
> > 
> > > There's actually no problem in having 2 sitenotice but when it go live it cause or probably occupies too many visual spaces on wikis. I think you would've seen a situation where you have your own sitenotice enabled and Miraheze has the Central notice enabled as well.
> 
> Central notice would be hard as you would have to manually build a list of wikis with the extension enabled.

That's a fact though. plus the point #dmehus gave.

DarkMatterMan4500:
I never would've imagined this update would take much longer than I would've hoped for.

ugochimobi:
> I never would've imagined this update would take much longer than I would've hoped for.

Fact. But doesn't really matter the time, so long the upgrade is successful, then we're good.

DarkMatterMan4500:
Well, technically speaking, yes.

PR #3962, opened 2021-06-14
Revert "Set WebChat to default to Libera; remove freenodeChat"

Reception123:
Reverts miraheze/mw-config#3926

Per the comments on the original task it is best to revert this for the time being until there is a discussion. Before doing this again please ask me.


Universal-Omega:
```sql
UPDATE mw_settings SET s_settings = REPLACE(s_settings,'LiberaChat', 'freenodeChat');
```
Needed to be ran for this to be done.

PR #3965, opened 2021-06-16
Enable some CheckUser configs

Universal-Omega:


LocalSettings.php
```diff
@@ -381,6 +381,28 @@
 	'wgCheckUserEnableSpecialInvestigate' => [
 		'default' => true,
 	],
+	'wgCheckUserLogLogins' => [
+		'default' => true,
+	],
+	'wgCheckUserCAtoollink' => [
```
The-Voidwalker, line 387: For all three of these options, the wiki should be metawiki, and the group should be steward. The wiki value is used only to generate a link to the special page on that wiki, and all our global (b)locks are done through metawiki. For the group, it tries to check the global group, as long as CentralAuth is installed.

The-Voidwalker (changes requested):

LocalSettings.php
Universal-Omega: Thanks! Done.

The-Voidwalker (approved):

dmehus:
LGTM. :+1:

PR #3966, opened 2021-06-16
Add reception wiki as Wiki Purpose

BukkitIsEnchanted:
A large portion of Miraheze wikis is Reception wikis, so I think it would be appropriate to add this as a wiki purpose in CW.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -748,6 +748,7 @@ $wi->config->settings += [
 			'Organization (coordination) wiki' => 'Organization (coordination) wiki',
 			'Political simulation wiki' => 'Political simulation wiki',
 			'Roleplaying game wiki' => 'Roleplaying game wiki',
+			'Reception wiki' => 'Reception wiki',
 			'Video game (specified video game) information wiki' => 'Video game (specified video game) information wiki',
 			'Video game (broad genre or video game series) information wiki' => 'Video game (broad genre or video game series) information wiki',
 			'None of the above' => 'None of the above',


dmehus:
@Reception123 asked for my input. If you'd asked me this last year, I would've supported adding this. In fact, as he may recall, I even _suggested_ it. However, nowadays, as circumstances have changed and we're more cautious with approving wikis, I'm not sure this is a good idea. Sure, we can potentially use this category to more easily monitor Reception wikis, but it'd be _voluntary_, meaning it'd be _optional_ so I'm not sure this would even be helpful here. Worse, in my view, is that this may serve as a "crutch" with wiki creators seeing it categorised and approving without knowing what the _content_ shall be. Finally, I _do_ also feel this would add duplication with the Fandom category and even other categories as well. So, my preference would be to decline this for now. Nevertheless, I appreciate @BukkitIsEnchanted's enthusiasm.

Reception123:
I would agree with Doug and would rather not have this, at least not for the time being.

BukkitIsEnchanted:
> @Reception123 asked for my input. If you'd asked me this last year, I would've supported adding this. In fact, as he may recall, I even _suggested_ it. However, nowadays, as circumstances have changed and we're more cautious with approving wikis, I'm not sure this is a good idea. Sure, we can potentially use this category to more easily monitor Reception wikis, but it'd be _voluntary_, meaning it'd be _optional_ so I'm not sure this would even be helpful here. Worse, in my view, is that this may serve as a "crutch" with wiki creators seeing it categorised and approving without knowing what the _content_ shall be. Finally, I _do_ also feel this would add duplication with the Fandom category and even other categories as well. So, my preference would be to decline this for now. Nevertheless, I appreciate @BukkitIsEnchanted's enthusiasm.
> 
> **Edit:** Re-read what's being proposed. I see it's for the _purpose_, but in this case, my argument about re: a Reception wiki standalone category stands and is even stronger

I get what you’re saying.

DarkMatterMan4500:
Sorry I'm late, but I'd definitely approve this action.

PR #3970, opened 2021-06-18
T7336: Install BlueSky skin

R4356th:
Per  T7336.


ManageWikiExtensions.php
```diff
@@ -1747,6 +1747,14 @@
 		'requires' => [],
 		'section' => 'skins',
 	],
+	'bluesky' => [
+		'name' => 'BlueSky',
+		'linkPage' => 'https://www.mediawiki.org/wiki/Skin:BlueSky',
```
Universal-Omega, line 1752: Special:MyLanguage

extension-list
```diff
@@ -293,3 +293,4 @@ $IP/skins/erudite/skin.json
 $IP/skins/foreground/skin.json
 $IP/skins/liberty/skin.json
 $IP/skins/pivot/skin.json
+$IP/skins/BlueSky/skin.json
```
Universal-Omega, line 296: Should be before "Citizen"

Universal-Omega (changes requested):

ManageWikiExtensions.php
R4356th: That is not necessary as that page has not been marked for translation.

RhinosF1: I'd rather we got it marked for translation

Universal-Omega: Still it's preferable to do that so it wouldn't need to be changed later if it were.

extension-list
R4356th: Unrelated to this PR.

Universal-Omega: How so?

ManageWikiExtensions.php
R4356th: Anyone volunteering?

extension-list
R4356th: Look at the lines above that. The whole file needs work and it does not seem worth the effort.

Universal-Omega: No everything else is in order.

R4356th: See 7dd66a3.

Universal-Omega: That's not how it should be ordered though.

Universal-Omega: Capital comes before lowercase in this file.

So Z would come before a in this instance.

R4356th: Done, I believe.

Universal-Omega: Yes, thank you!

ManageWikiExtensions.php

Universal-Omega: Thank you!

PR #3973, opened 2021-06-21
Let otf files recognise application/vnd.ms-opentype mime type

Universal-Omega:
Per [T7506](https://phabricator.miraheze.org/T7506)

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -224,6 +224,12 @@ if ( !preg_match( '/^(.*).miraheze.org$/', $wi->hostname ) ) {
 		array_merge( $wgUrlShortenerAllowedDomains, [ preg_quote( str_replace( 'https://', '', $wgServer ) ) ] );
 }
 
+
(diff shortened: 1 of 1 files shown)


RhinosF1 (changes requested):
Please just handle upstream. Patch is at https://gerrit.wikimedia.org/r/c/mediawiki/core/+/700632

Universal-Omega:
> Please just handle upstream. Patch is at https://gerrit.wikimedia.org/r/c/mediawiki/core/+/700632

Yeah but then we have to wait until at least 1.36.1 before we can get it to work, I think.

RhinosF1:
> > Please just handle upstream. Patch is at https://gerrit.wikimedia.org/r/c/mediawiki/core/+/700632
> 
> Yeah but then we have to wait until at least 1.36.1 before we can get it to work, I think.

Why can't we cherry pick?

RhinosF1:
Picked from core directly

PR #3986, opened 2021-06-27
Update wgLinterSubmitterWhitelist to mw* and test3 and set wgParsoidSettings

Universal-Omega:
To hopefully fix [T7416](https://phabricator.miraheze.org/T7416).


RhinosF1 (approved):
I trust you can copy/paste so not checked the IPs but Thanks!

LocalSettings.php
```diff
@@ -2607,12 +2607,24 @@
 	// Linter
 	'wgLinterSubmitterWhitelist' => [
 		'wmgUseLinter' => [
```
The-Voidwalker, line 2609: Why is this `wmgUseLinter` and not `default`?

Universal-Omega: Because that makes it only be used when linter is enabled.

The-Voidwalker: Hmm, I'm unfamiliar with this syntax for wgConf, but it does appear that it works.

The-Voidwalker (approved):

LocalSettings.php
Universal-Omega: It's something that was added not very long ago.

Reception123 (approved):

PR #3987, opened 2021-06-28
T6471: Set wgGitInfoCacheDirectory

Universal-Omega:


Universal-Omega:
Setting this doesn't do any harm, so merging and just the core hack needs reviewed.

PR #3992, opened 2021-06-29
make Libera default for wgWebChatClient

The-Voidwalker:
remove freenode

See https://meta.miraheze.org/wiki/IRC/Replacement_of_default_chat_platform_in_WebChat_extension


Universal-Omega (changes requested):
Don't forget to change it in https://github.com/miraheze/mw-config/blob/a12401/LocalSettings.php#L4458 also, and change it in the DB for existing ones.

PR #3994, opened 2021-07-04
Adding PageImages Namespaces to my Wiki

Njsmooth:
Added namespaces for PageImages to display images in page previews for more namespaces in my wiki. The setting wasn't already there, so I tried to add it in. I may have done it incorrectly, though...

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3438,7 +3438,9 @@ $wi->config->settings += [
 
 	// Page Images
 	'wgPageImagesNamespaces' => [
-		'default' => NS_MAIN,
+		'default' => [
+			NS_MAIN,
+		],
 		'vgportdbwiki' => [
 			NS_MAIN,
 			3000,
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3439,7 +3439,12 @@ $wi->config->settings += [
 	// Page Images
 	'wgPageImagesNamespaces' => [
 		'default' => NS_MAIN,
-		'vgportdbwiki' => [NS_MAIN, 3000, 3004, 3006],
+		'vgportdbwiki' => [
+			NS_MAIN,
+			3000,
+			3004,
+			3006,
+		],
 	],
 	
 	// Pagelang
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3436,6 +3436,12 @@ $wi->config->settings += [
 		'default' => true,
 	],
 
+	// Page Images
+	'wgPageImagesNamespaces' = [
+		'default' => NS_MAIN,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3436,6 +3436,12 @@
 		'default' => true,
 	],
 
+	// Page Images
+	'wgPageImagesNamespaces' = [
+		'default' => NS_MAIN,
```
Universal-Omega, line 3441: This should be an array.

LocalSettings.php
```diff
@@ -3436,6 +3436,12 @@
...
+		'vgportdbwiki' => [NS_MAIN, NS_PLATFORM, NS_SERIES, NS_COMPANY],
```
Universal-Omega, line 3442: This should be reformatted so each constant is on a new line. Also, as for custom namespaces, I don't think the constants will work and need to be ID directly though not positive.

Universal-Omega (dismissed):

Universal-Omega:
Also syntax error.

LocalSettings.php
```diff
@@ -3436,6 +3436,12 @@
 		'default' => true,
 	],
 
+	// Page Images
+	'wgPageImagesNamespaces' = [
```
The-Voidwalker, line 3440:
```suggestion
	'wgPageImagesNamespaces' => [
```

The-Voidwalker (changes requested):

LocalSettings.php
Njsmooth: I'm not even entirely sure what the constants are (if they even have one that is). Do you know where I'd be able to check?

Universal-Omega: > I'm not even entirely sure what the constants are (if they even have one that is). Do you know where I'd be able to check?

If you see Special:ManageWiki/namespaces and select your namespace and select edit, the final subpage in the address bar is the namespace ID. (Custom ones are always above 3000)

Njsmooth: Oh, I know about the ID, but I wasn't sure about the constant. Also, id I were to try using the ID's, how would I type that in the code?

Universal-Omega: Just replace the constants with the ID integer.

Universal-Omega:
It should look like:
```
	// Page Images
	'wgPageImagesNamespaces' => [
		'default' => [
			NS_MAIN,
		],
		'vgportdbwiki' => [
			NS_MAIN,
			3000,
			3004,
			3006,
		],
	],
```

LocalSettings.php
Universal-Omega: I'll merge this once this is done.

Njsmooth: Did I fix it?

PR #3995, opened 2021-07-08
Added a setting for MultiBoilerplate

Njsmooth:
I added in a configuration option for MultiBoilerplate that wasn't already there (maybe add it to managewiki?). Also, there seemed to be duplicate configurations as well as some formatting oddities, so I changed them to be a bit more uniform.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3380,7 +3380,7 @@ $wi->config->settings += [
 
 	// MultiBoilerplate
 	'wgMultiBoilerplateDisplaySpecialPage' => [
-		'default' => true,
+		'wmgUseMultiBoilerplate' => true,
 	],
 	'wgMultiBoilerplateOptions' => [
 		'default' => false,
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3380,13 +3380,14 @@ $wi->config->settings += [
 
 	// MultiBoilerplate
 	'wgMultiBoilerplateDisplaySpecialPage' => [
-		'wmgUseMultiBoilerplate' => true,
+		'wmgUseMultiBoilerplate' => false,
+		'vgportdbwiki' => true,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3385,10 +3380,14 @@
 
 	// MultiBoilerplate
 	'wgMultiBoilerplateDisplaySpecialPage' => [
-		'wmgUseMultiBoilerplate' => true,
```
Universal-Omega, line 3388: This should still be wmgUseMultiBoilerplate

Universal-Omega (changes requested):

LocalSettings.php
Njsmooth: Did the change I make do it, or do I need to change all of them?

LocalSettings.php
```diff
@@ -3388,7 +3383,11 @@
 		'wmgUseMultiBoilerplate' => true,
 	],
 	'wgMultiBoilerplateOptions' => [
-		'wmgUseMultiBoilerplate' => false,
```
Universal-Omega, line 3391: This should still be wmgUseMultiBoilerplate also.

LocalSettings.php
```diff
@@ -3388,7 +3383,11 @@
...
+		'default' => false,
+	],
+	'wgMultiBoilerplateOverwrite' => [
+		'default' => false,
```
Universal-Omega, line 3389: wmgUseMultiBoilerplate

LocalSettings.php
```diff
@@ -3388,7 +3383,11 @@
 		'wmgUseMultiBoilerplate' => true,
```
Universal-Omega, line 3383: Should default to false.

Universal-Omega (dismissed):

PR #3998, opened 2021-07-14
T7570: Install SkinPerPage

R4356th:
Per T7570.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -862,6 +862,14 @@ $wgManageWikiExtensions = [
 		'requires' => [],
 		'section' => 'parserhooks',
 	],
+	'skinperpage' => [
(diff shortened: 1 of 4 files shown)


ManageWikiExtensions.php
```diff
@@ -862,6 +862,14 @@
 		'requires' => [],
 		'section' => 'parserhooks',
 	],
+	'skinperpage' => [
+		'name' => 'SkinPerPage',
```
Universal-Omega, line 866: "Skin per page" this should now match ExtensionRegistry name, once my ManageWiki and CreateWiki changes are done, so I'd prefer to do that now for this.

Universal-Omega (changes requested):

ManageWikiExtensions.php

Universal-Omega (approved):

PR #4004, opened 2021-07-19
Re-enable GlobalNewFiles

Universal-Omega:


RhinosF1 (approved):
I'm happy with the mitigations in place.

There's more work to do but I think the risk is more acceptable.

Please ensure the puppet patch goes first though so we can monitor jobs.

Universal-Omega:
> I'm happy with the mitigations in place.
> 
> There's more work to do but I think the risk is more acceptable.
> 
> Please ensure the puppet patch goes first though so we can monitor jobs.

Yep, that is what this is waiting on. Once miraheze/puppet#1815 is merged we can re-enable it.

PR #4017, opened 2021-07-29
LS: Set MaximalPasswordLength to 128

RhinosF1:


Reception123:
Closing per comments made in T7713 against lowering the max.

PR #4023, opened 2021-07-30
test3: test a theory

RhinosF1:
to see if we can deploy mediawiki patches locally for sec issues and potentially save time on git clones, try put stuff in mediawiki-staging.

If I can figure a process, we might need to pick one of the mw* and then start with the new sync system.

Most of it could naturally run by a post-deploy bash file.


RhinosF1:
not needed, has --outdir

PR #4048, opened 2021-08-11
Add mw12 and mw13 to wgLinterSubmitterWhitelist

Universal-Omega:


Universal-Omega:
Waiting on mw12 and mw13 to be fully operational.

PR #4066, opened 2021-09-04
Setting image repository for polandballwikisongcontestwiki

AgentIsai:
I was unable to enable shared image repository via ManageWiki for `polandballwikisongcontestwiki` to use `polcomwiki` as an image repo as `$wmgSharedUploadClientDBname` was already filled on `polcomwiki` by another wiki thus making it impossible to allow multiple wikis to use a single shared repository. I am proposing this manual change to allow the wiki in question to use the `polcomwiki` as an image repository.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -321,7 +321,7 @@ if ( $wgDBname === 'polandballwikisongcontestwiki' ) {
 		'descriptionCacheExpiry' => 86400 * 7,
 		'thumbScriptUrl' => false,
 		'transformVia404' => !$wgGenerateThumbnailOnParse,
-		'hasSharedCache' => 'polcomwiki',
+		'hasSharedCache' => false,
 		'wiki' => 'polcomwiki',
 		'descBaseUrl' => 'https://commons.polandballwiki.com/wiki/File:',
 	];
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -309,7 +309,23 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+		'class' => 'ForeignDBViaLBRepo',
+		'name' => 'shared-polcomwiki',
+		'directory' => '/mnt/mediawiki-static/polcomwiki',
+		'url' => 'https://static.miraheze.org/polcomwiki',
+		'hashLevels' => $wgHashedSharedUploadDirectory ? 2 : 0,
+		'fetchDescription' => true,
+		'descriptionCacheExpiry' => 86400 * 7,
+		'thumbScriptUrl' => false,
+		'transformVia404' => !$wgGenerateThumbnailOnParse,
+		'hasSharedCache' => 'polcomwiki',
```
Universal-Omega, line 324: This is boolean, should be false

Universal-Omega (changes requested):

LocalWiki.php
AgentIsai: This has been fixed, please review when you have the time, thanks!

Universal-Omega (approved):
I want to add a multi-repository support to ManageWiki, but in the meantime this looks good to me.

PR #4081, opened 2021-09-07
Set $wgParserCacheType to cache in the DB

paladox:
We have limited amount of memory so lets use the db where we have more space to store the cache longer.


paladox:
by the looks of things we can do similar to https://github.com/wikimedia/operations-mediawiki-config/blob/d7227d010ce88ae8c4711a7d3e23d4445b3fc27a/wmf-config/CommonSettings.php#L535 with a fallback to mysql? Unless I'm misreading.

PR #4087, opened 2021-09-08
Set Scribunto configs

Universal-Omega:


paladox:
Need to also remove from LS.

PR #4102, opened 2021-09-18
Adding wikis to image repository

AgentIsai:
Requested on Discord; added 2 wikis to polcomwiki's image repo

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -52,7 +52,7 @@ if ( $wgDBname === 'pokemundowiki' ) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -52,7 +52,10 @@ function onBeforePageDisplay( OutputPage $out ) {
 	}
 }
 
-if ( $wgDBname === 'polandballwikisongcontestwiki' ) {
+if ( $wgDBname === 'polandballwikisongcontestwiki' ||
```
Universal-Omega, line 55: This can be simplified to `in_array( $wgDBname, [ ... ] )` it'd probably be better that way, but not a blocker for deployment of this. If you don't want to I will merge right now though.

PR #4103, opened 2021-09-18
Add Evelution settings to ManageWiki

AgentIsai:
I added all the settings for Evelution to ManageWiki settings excluding '$wgEvelutionCustomFont' as it is unclear on whether it loads a font from a URL or changes font styles.


ManageWikiSettings.php
```diff
@@ -2408,6 +2408,65 @@
...
+		'requires' => [],
+	],
+	'wgEvelutionStickyRail' => [
+		'name' => 'Disable Evelution Sticky Rail',
+		'from' => 'evelution',
+		'type' => 'check',
+		'overridedefault' => 'true',
+		'section' => 'styling',
+		'help' => 'Removes the sticky behavior of Right Rail on larger screens with 700 pixels or more.',
+		'requires' => [
```
Universal-Omega, line 2463: Indentation

Universal-Omega (changes requested):

ManageWikiSettings.php
```diff
@@ -2408,6 +2408,65 @@
...
+			'settings' => [
+				'setting' => 'wgEvelutionDisableRightRail',
```
Universal-Omega, line 2465: * This currently is not ideal implementation in ManageWiki. I prefer not to use it at the moment, I'd just remove it, it's. Ot absolutely dependant on that setting. We can add this if miraheze/ManageWiki#298 ends up working after I test it.
* Secondly, throughout this PR all boolean values (`false` and `true` should be removed from the quotations.

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #4104, opened 2021-09-18
Adding wgDiscordNotificationAfterImportPage to settings

AgentIsai:
Requested on Discord; adding $wgDiscordNotificationAfterImportPage to ManageWiki Settings which controls whether a notification is generated when a page is imported or not

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -1440,7 +1440,7 @@ $wgManageWikiSettings = [
 		'requires' => [],
 	],
 	'wgDiscordNotificationAfterImportPage' => [
-		'name' => 'Discord Notification Import Article',
+		'name' => 'Discord Notification After Article Import',
 		'from' => 'mediawiki',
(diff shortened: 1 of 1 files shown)


ManageWikiSettings.php
```diff
@@ -1439,6 +1439,15 @@
 		'help' => 'Notify when an article is protected?',
 		'requires' => [],
 	],
+	'wgDiscordNotificationAfterImportPage' => [
+		'name' => 'Discord Notification Import Article',
+		'form' => 'mediawiki',
```
Universal-Omega, line 1444: Typo: should be `from`

Universal-Omega (changes requested):

ManageWikiSettings.php
Universal-Omega: Also the name is missing the word `After` which makes it a bit unclear.

Universal-Omega (approved):

PR #4120, opened 2021-09-30
Propose "Excessive requests" canned response

AgentIsai:
Recently, we've had an influx of excessive wiki requests made by particular users. I believe it would be good if we added a canned response for excessive requests so that we can decline them more efficiently as writing/copy and pasting a decline reason across all of these requests is a bit burdensome.


RhinosF1 (approved):

RhinosF1:
Can we get a community discussion?

AgentIsai:
@RhinosF1 There is an on-going RfC regarding proposing a mandatory cooldown period between wiki requests. While current consensus seems to be against proposing a mandatory time frame, there seems to be support for using current discretion to decline excessive wiki requests. This I feel wouldn't be granting wiki creators any additional right or discretion but instead would simply result in less typing.

Universal-Omega (approved):

Universal-Omega:
> Can we get a community discussion?

I don't think community consensus is necessarily required here, since this is basically wiki creator's decision whether to decline or not. And I do think this is a useful reason.

RhinosF1:
I wasn't expecting full consensus, more just an FYI with no objections raised in a day or 2

AgentIsai:
I add a "[Request for feedback](https://meta.miraheze.org/wiki/Requests_for_Comment/Window_between_wiki_requests#Request_for_feedback)" section to the RfC 2 days ago and so far has 2 supports (including myself, as proposer) and no opposes.

RhinosF1:
No objections

PR #4121, opened 2021-10-01
set $wgDplSettings['functionalRichness'] = 0; temporarily

Reception123:


Defines.php
```diff
@@ -13,6 +13,10 @@
 	wfLoadExtension( 'ElectronPdfService' );
 }
 
+if ( $wmgUseDynamicPageList ) {
```
Universal-Omega, line 16: DynamicPageList3

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #4130, opened 2021-10-06
Raise DPL requirement to 40,000 pages

Universal-Omega:
DPL3 at least should not have as much performance issues anymore. This can probably be further increased later on, but 40,000 is enough for now to gradually increase and see if issues continue to happen.


RhinosF1:
Let's just give it like a few hours to ensure nothing happens. If we're fine at 10pm UK time then go for it. As discussed, up to local wikis to re-enable.

PR #4208, opened 2021-11-07
Set default rights for CommentStreams and SpriteSheet

Kelvs599:
If I'm not mistaken, it seems that commenting needs cs-comment to work. The other additions are what the defaults are.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -281,7 +281,7 @@ $wgManageWikiExtensions = [
 				'cs137temp' => '/srv/mediawiki/config/commentstreams_temp.sql',
 			],
 			'permissions' => [
-				'*' => [
+				'user' => [
 					'permissions' => [
 						'cs-comment',
 					],
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -280,6 +280,26 @@ $wgManageWikiExtensions = [
 				'watch' => "$IP/extensions/CommentStreams/sql/watch.sql",
 				'cs137temp' => '/srv/mediawiki/config/commentstreams_temp.sql',
 			],
+			'permissions' => [
+				'*' => [
+					'permissions' => [
+						'cs-comment',
+					],
+				],
+				'csmoderator' => [
+					'permissions' => [
+						'cs-moderator-delete',
+					],
+				],
+				'bureaucrat' => [
+					'addgroups' => [
+						'csmoderator',
+					],
+					'removegroups' => [
+						'csmoderator',
+					],
+				],
+			],
 		],
 		'section' => 'parserhooks',
 	],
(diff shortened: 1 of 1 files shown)


ManageWikiExtensions.php
```diff
@@ -280,6 +280,26 @@
 				'watch' => "$IP/extensions/CommentStreams/sql/watch.sql",
 				'cs137temp' => '/srv/mediawiki/config/commentstreams_temp.sql',
 			],
+			'permissions' => [
+				'*' => [
+					'permissions' => [
+						'cs-comment',
```
Universal-Omega, line 286: Maybe should be user or even autoconfirmed by default, not *?

Kelvs599: Well, my thought process was that * has edit by default, so giving cs-comment to * would presumably be fine, but I suppose we can set it to user instead if that's more appropriate.

dmehus: What's the extension programmed to assign as a default right? Note that for Comments, there's a php variable to enable/disable anonymous comments, so it may be automatically added to `*` by default when that variable is set to `true`; otherwise, it may be added to `user` by default. Something to consider.

Universal-Omega: I think setting this to user would be best  if wanted anonymous to be able to can manually add it I think.

dmehus: > I think setting this to user would be best if wanted anonymous to be able to can manually add it I think.

Yeah, I'd tend to agree with that line of thinking generally. I also just noticed [this](https://www.mediawiki.org/wiki/Extension:CommentStreams#Comment_right) where it says that if _no_ group has the `cs-comment` user right, any user with the `edit` right has it implicitly. So I'm just wondering, is this even needed? Note, too, that below, there's an option to set anonymous comments to `true` or `false`. Are those variables added to ManageWiki? If _not_, I'd say we should add them. If _so_, I'm not sure we need to do anything here, now that I think about it.

Kelvs599: dmehus, I'm adding cs-comment because I recall various users in Discord noticing that CommentStreams apparently doesn't work without cs-comment being assigned (you can search for CommentStreams in the Discord server if you want to see the actual messages). As for the anonymous comments option, I assume you're talking about the $wgGroupPermissions thing. The equivalent to that on this wikifarm is just Special:ManageWiki/permissions.

Kelvs599: Anyway, do we agree on setting this to user?

AgentIsai: Per my testing, cs-comment is a required right and is not automatically implied by `edit`. 

Regardless, I think it *should* be changed to a right assignable to the `user` group.

Kelvs599: Alright, I've changed it to set cs-comment to user instead of *.

dmehus: @Kelvs599 Okay, yeah, I understand that, and @Agent-Isai, hrm, that's interesting. Perhaps the MediaWiki.org documentation page needs to be updated then? Why are setting this as a default user right in `LocalSettings.php`? Can this not be done in ManageWiki, or perhaps via the ManageWiki configuration file?

Universal-Omega: We are setting this a default right through ManageWiki.... LocalSettings.php is not touched here.

dmehus: Ah, thanks. It's not obvious on this PR //which// file is being edited, so I didn't see that.

Universal-Omega: The title of this conversation is "ManageWikiExtensions.php"; and you could see which file is being edited from the files tab.

dmehus: @Universal-Omega Ah, true. Fair enough then. That being said, I wish GitHub would add the file being edited to the top, like this:
`User A wants to merge 2 commits in filename into repo:branch from repo:branch`

Universal-Omega: For alot of files that would be a very large message...

Universal-Omega (approved):

PR #4219, opened 2021-11-14
TE and XCON protection levels on GP

ugochimobi:
Please review codes and merge appropriately, ;)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3405,6 +3405,10 @@ $wi->config->settings += [
 			'editextendedconfirmedprotected',
 			'edittemplateprotected',
 		],
+		'gratispaideiawiki' => [
+			'editextendedconfirmedprotected',
+			'edittemplateprotected',
+		],
 		'hypopediawiki' => [
 			'editextendedconfirmedprotected',
 		],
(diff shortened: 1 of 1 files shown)


dmehus (commented):
This looks fine to me, but I believe will also need to these user rights to the additional ManageWiki rights, so they can be assigned in ManageWiki/permissions.

ugochimobi:
> This looks fine to me, but I believe will also need to these user rights to the additional ManageWiki rights, so they can be assigned in ManageWiki/permissions.

TBH, that wouldn't be a bad idea

Universal-Omega:
> > This looks fine to me, but I believe will also need to these user rights to the additional ManageWiki rights, so they can be assigned in ManageWiki/permissions.
> 
> TBH, that wouldn't be a bad idea

Rights need added to $wgAvailableRights for this.

dmehus:
> > > This looks fine to me, but I believe will also need to these user rights to the additional ManageWiki rights, so they can be assigned in ManageWiki/permissions.
> > 
> > 
> > TBH, that wouldn't be a bad idea
> 
> Rights need added to $wgAvailableRights for this.

Thanks, I couldn't think of the exact variable for that. :)

ugochimobi:
Okay, I just did it.

ugochimobi:
Okay @Universal-Omega I just did it here.

PR #4230, opened 2021-11-22
Configure `$wgRemovePIIAllowedWikis`

Universal-Omega:
Only allows Special:RemovePII to be used from `metawiki` and `test3wiki`, but it still runs jobs on all wikis.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3305,6 +3305,13 @@ $wi->config->settings += [
 	'wgRemovePIIHashPrefix' => [
 		'default' => 'MirahezeGDPR_',
 	],
+	'wgRemovePIIAllowedWikis' => [
+		'default' => [
+			'loginwiki',
+			'metawiki',
+			'test3wiki',
+		],
+	],
 
 	// ReplaceText
 	'wgReplaceTextResultsLimit' => [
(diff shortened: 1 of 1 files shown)


dmehus:
This seems fine to me. I mean, you've verified that no local user groups on any wiki have the `handle-pii` user right and that user right is blacklisted from ManageWiki/permissions. It seems like a redundant safety check, I gather? Assuming that's the case, I have no objections to this personally, with or without my Trust and Safety hat on. I suspect @OwenBaines would also have no objections to this, but pinging him here nonetheless, as he is the Director of Trust and Safety and this is certainly a PR that would likely be of interest to him.

Universal-Omega:
> This seems fine to me. I mean, you've verified that no local user groups on any wiki have the `handle-pii` user right and that user right is blacklisted from ManageWiki/permissions. It seems like a redundant safety check, I gather? Assuming that's the case, I have no objections to this personally, with or without my Trust and Safety hat on. I suspect @OwenBaines would also have no objections to this, but pinging him here nonetheless, as he is the Director of Trust and Safety and this is certainly a PR that would likely be of interest to him.

Yeah this is just to provide absolute guarantee it won't be usable through other wikis.

PR #4249, opened 2021-12-03
LS: blacklist .*betawiki in prep for future usage

RhinosF1:
> select wiki_dbname from cw_wikis where wiki_dbname LIKE '%betawiki';
Query OK, 0 row(s) affected

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -805,6 +805,7 @@ $wi->config->settings += [
 			'mwtask\d{0,2}',
 			'security',
 			'deployment',
+			'.*beta',
 		],
 	],
 	'wgCreateWikiCannedResponses' => [


RhinosF1:
Deployed on test3

Universal-Omega (commented):
if the intention is to blacklist \*beta.miraheze.org this is valid. This would not blacklist beta\*.miraheze.org at all, I don't believe though I could be wrong.

Universal-Omega (approved):

JohnFLewis:
This seems like a too far ranging blacklist entry for me. It would prevent legitimate database names from being used by the community.

RhinosF1:
> This seems like a too far ranging blacklist entry for me. It would prevent legitimate database names from being used by the community.

The aim is so in the eventual beta cluster we can have all wiki dbnames end with a specific suffix and only have the beta wikis use credentials to access wikis with that suffix but avoid production potentially conflicting with dbnames while still using the same db hosts.

JohnFLewis:
What about legitimate use cases for wiki names ending in betawiki?

RhinosF1:
> What about legitimate use cases for wiki names ending in betawiki?

I checked that there was no current ones. The other option I had was to extend CreateWiki to support changing the suffix.

JohnFLewis:
Just because there are none now, does not mean there will never be anyone. A legitimate use case would be a wiki documenting a bleeding edge beta release branch of software, so <software>betawiki would be a legitimate use case that I don't think we should prevent/ban/blacklist from being used.

RhinosF1:
Then I guess I need to implement
> The other option I had was to extend CreateWiki to support changing the suffix.

JohnFLewis:
Alternatively, identifying the relevant names to use would be sufficient - rather than investing time to develop for an edge case here. As you've identified, there are no existing names - so if you choose the names now, there would be no clashes to be met.

Universal-Omega:
> Then I guess I need to implement
> 
> > The other option I had was to extend CreateWiki to support changing the suffix.

That should already be possible I think by changing `$wi->setVariables()`.

RhinosF1:
That's also an option.

There's probably only going to be about 4/5 permanent wikis. (Testing CreateWiki could involve temp wikis). It would only be open to trusted users so we could just trust they won't split brain anything by creating a conflicting name.

Reception123:
I think beta.miraheze.org should be blacklisted but it wouldn't be fair to blacklist every single use of the word 'beta' per the comments above.

RhinosF1:
Closing this while I assess the other 2 solutions

PR #4263, opened 2021-12-05
Fix config naming for 1.37

Universal-Omega:
This will need ran:
```sql
UPDATE mw_settings SET s_settings = REPLACE(s_settings,'wgRelatedArticlesFooterWhitelistedSkins', 'wgRelatedArticlesFooterAllowedSkins');
```


RhinosF1:
Can we not just set $wgNew = $wgOld to reduce during upgrade work (to remove post upgrade but when calm)?

PR #4270, opened 2021-12-07
Experimental: update for 1.37

RhinosF1:
1) remove JQuery, no one raised issues so this can go to prod
2) <s>Enable new media structure on all wikis but with cache so parser cache doesn't need killing</s>
3) Use new image metadata config on experimental

### DIFF
--- a/Defines.php
+++ b/Defines.php
@@ -107,17 +107,10 @@ if ( !$cwPrivate ) {
 
 // Experimental Wikis
 if ( $cwExperimental ) {
-	$wgIncludejQueryMigrate = false;
-	if ( version_compare( MW_VERSION, '1.37', '>=' ) ) {
-		$wgParserEnableLegacyMediaDOM = false;
-	} else {
-		$wgUseNewMediaStructure = true;
-	}
+	$wgParserEnableLegacyMediaDOM = false;
 } else {
-	$wgIncludejQueryMigrate = true;
-	if ( version_compare( MW_VERSION, '1.37', '>=' ) ) {
-		$wgParserEnableLegacyMediaDOM = true;
-	}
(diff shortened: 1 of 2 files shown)


Defines.php
```diff
@@ -107,17 +107,27 @@
...
+		'useJsonMetadata' => true,
+		'useSplitMetadata' => true,
+	];
 } else {
-	$wgIncludejQueryMigrate = true;
-	if ( version_compare( MW_VERSION, '1.37', '>=' ) ) {
-		$wgParserEnableLegacyMediaDOM = true;
-	}
+	$wgParserEnableLegacyMediaDOM = false;
+	$wgUseContentMediaStyles = true;
```
Universal-Omega, line 130: This does not exist until 1.38.

Defines.php
```diff
@@ -107,17 +107,27 @@
...
+		'reserializeMetadata' => $wgUpdateCompatibleMetadata,
+		'useJsonMetadata' => true,
+		'useSplitMetadata' => true,
+	];
 } else {
-	$wgIncludejQueryMigrate = true;
-	if ( version_compare( MW_VERSION, '1.37', '>=' ) ) {
-		$wgParserEnableLegacyMediaDOM = true;
-	}
+	$wgParserEnableLegacyMediaDOM = false;
```
Universal-Omega, line 129: Should this not be added to LS, if it is set to false for regardless of experimental?

Universal-Omega (changes requested):

Defines.php
RhinosF1: We'll need legacy on everywhere then expect experimental. I don't fancy a full PC purge.

PR #4273, opened 2021-12-08
Profile: Profile mw8

RhinosF1:


PhpAutoPrepend.php
```diff
@@ -20,6 +20,19 @@ function wfSetupProfiler() {
 			'output' => 'text',
 		];
 	}
+	$host = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : '';
```
Universal-Omega, line 23: $_SERVER['HTTP_HOST'] ?? ''

Universal-Omega (changes requested):

Universal-Omega:
Actually this entire thing can be simplified by just doing || host === 'mw8...' in the first one, I think.

RhinosF1:
> Actually this entire thing can be simplified by just doing || host === 'mw8...' in the first one, I think.

the config is different

Universal-Omega:
> > Actually this entire thing can be simplified by just doing || host === 'mw8...' in the first one, I think.
> 
> the config is different

I realised that but a few additional checks can be added to make it different for mw8, this is a lot of duplicated code.

PhpAutoPrepend.php
```diff
@@ -20,6 +20,19 @@ function wfSetupProfiler() {
 			'output' => 'text',
 		];
 	}
+	$host = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] ?? '';
```
Universal-Omega, line 23: $host = $_SERVER['HTTP_HOST'] ?? '';

PhpAutoPrepend.php
```diff
@@ -20,6 +20,19 @@ function wfSetupProfiler() {
...
+	if ( $host = 'mw8.miraheze.org' ) {
```
Universal-Omega, line 24: $host ===

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #4291, opened 2021-12-19
LS: setup betawiki

RhinosF1:
Bug: T8399


Universal-Omega (commented):
Can test3wiki not be removed if we are deleting it soon?

RhinosF1:
> Can test3wiki not be removed if we are deleting it soon?

Once it's been deleted yes, it's not immediately going and things are pointed specifically at test3 to avoid touching prod so I'd rather we didn't reintroduce any links.

Universal-Omega (approved):
Ignore this comment, for some reason it tried to submit the same message I posted above.

PR #4292, opened 2021-12-20
LS: move realm specific config to separate files

RhinosF1:


BetaSettings.php
```diff
@@ -0,0 +1,45 @@
+/**
```
supertassu, line 1: missing opening php header

ProductionSettings.php
```diff
@@ -0,0 +1,45 @@
+/**
```
supertassu, line 1: this too

supertassu (changes requested):

BetaSettings.php
RhinosF1: done

ProductionSettings.php
RhinosF1: done

JohnFLewis:
Surely this can just be done on LS as it was before? This seems like a step back in terms of config, especially as LS is managed by ManageWiki, and this new system wouldn’t be.

Wikis can be tagged in this format, it would seem sensible to do that, rather than split config into 3 distinct streams for no reason.

Universal-Omega:
To me this kinda just makes a mess of things. I personally don't see the point in moving out of LS at this time. It seems to be cleaner in LS and 2 additional files don't help with organisation much, and only cause things to be a bit of a mess. Just my opinion though.

PR #4293, opened 2021-12-20
Configure betaheze realm for SiteConfiguration tag

Universal-Omega:
* miraheze/CreateWiki#268 should add support for this.


RhinosF1:
I can handle the config changes seperately

PR #4294, opened 2021-12-21
PageForms: disable

RhinosF1:


LocalSettings.php
```diff
@@ -4416,7 +4416,7 @@
 
 // ManageWiki settings
 require_once __DIR__ . '/ManageWikiExtensions.php';
-$wi->disabledExtensions = [];
+$wi->disabledExtensions = [ 'pageforms' ];
```
Universal-Omega, line 4419: This won't disable. Needs to be PageForms, the same value as the 'name' field in ManageWiki I believe.

Universal-Omega (changes requested):

PR #4310, opened 2021-12-29
SCSVG: setup

RhinosF1:
- [x] memcache
- [x] wfHostname
- [x] DB Servers
- [x] Linter allowlist
- [x] populate IPs


RhinosF1:
smoke test on test3 looked fine. I logged in to beta & made an edit then reverted. No errors seen.

Universal-Omega:
This kinda makes a mess of things. This should be temporary if we do it like this. Therefore I don't think $wmgCacheSettings should really be completely removed.

RhinosF1:
> This kinda makes a mess of things. This should be temporary if we do it like this. Therefore I don't think $wmgCacheSettings should really be completely removed.

Is it not easier to find the information if it's it one file?

But the actual arrays will be temporary and anything old DC related once we're on SCSVG fully

RhinosF1:
Explained on IRC better

Universal-Omega (approved):

PR #4312, opened 2021-12-29
Redo CreateWiki subdomain blacklist

Universal-Omega:
* Adds new servers (es and bastion)
* Switches hostname regexes to `\d+` or `(\d+)?` to blacklist all possible hostnames
* Removes duplicate `mw` and `misc` entries from blacklist
* Removes `\d{0,2}` for entires that don't need that
* Combines entries where possible
* Blacklists `miraheze(.*)` rather than `miraheze\d{0,2}`
* Blacklists `betaheze(.*)`
* Blacklists `elasticsearch(\d+)?` to match the fact all current *and* former hostnames are blacklisted
* Adds `(fs)?` to gluster entry in blacklist, in order to blacklist `glusterfs#`
* Blacklists m, donate, blog, hostmaster, list, localhost, wildcard, mailman, static, matomo, grafana, icinga, phabricator, webmail, mail, and graylog


RhinosF1 (approved):

RhinosF1:
@The-Voidwalker: I think you might have come up with some of these and know regex well. FYI and any review appreciated.

RhinosF1 (approved):

dmehus:
:+1:. Good to request @The-Voidwalker's input, to potentially improve the regex syntax some more.

Universal-Omega:
> 👍. Good to request @The-Voidwalker's input, to potentially improve the regex syntax some more.

Tbh, I doubt the regex syntax can be improved much. I'm pretty sure this'll work, as I know regex pretty well now, since I've had to use it so much lately. I just want to be sure that the changes in this PR, to blacklist additional stuff etc... is definitely alright to do before deployment.

The-Voidwalker (approved):
Looks good to me. We could possibly combine a few entries, but then it would be harder to understand what is actually going on (which we should probably avoid).

Universal-Omega:
> Looks good to me. We could possibly combine a few entries, but then it would be harder to understand what is actually going on (which we should probably avoid).

Yeah, that's why I didn't combine more than I did. I just did a few very similar/simple ones combined. Thanks!

PR #4313, opened 2021-12-30
Add sitenotice about Comments being unavailable 

AgentIsai:


Universal-Omega (changes requested):
Needs to be uncommented and the id bumped maybe?

AgentIsai:
Maybe, just maybe...

Universal-Omega (approved):

PR #4328, opened 2022-01-02
Split `$wgCreateWikiCannedResponses` into sections

Universal-Omega:


dmehus (commented):
This looks fine to me. Will these be in one, or separate, drop-down boxes? I'd definitely prefer the former, with things like "Approval reasons," "Decline reasons," and "On hold reasons" being greyed our section headers

Universal-Omega:
> This looks fine to me. Will these be in one, or separate, drop-down boxes? I'd definitely prefer the former, with things like "Approval reasons," "Decline reasons," and "On hold reasons" being greyed our section headers

That's exactly what they should be.

PR #4340, opened 2022-01-05
Use MemcachedPeclBagOStuff on new mediawiki infra

paladox:


Universal-Omega:
What about for memcached-mem-2?

PR #4349, opened 2022-01-07
Shell: cap memoryLimit

RhinosF1:
Hopefully prevent some pages failing to load


RhinosF1:
I've set the shell cap early on so it affects everything loaded at any point.

Reception123 (approved):

PR #4353, opened 2022-01-07
[19:45] DC-Switch: Stop uploads

RhinosF1:


LocalSettings.php
```diff
@@ -4571,7 +4571,7 @@ function extractGlobals() {
 	global $wgConf, $wgDBname, $wgAPIModules;
```
Universal-Omega, line 4571: global `$wgRevokePermissions` though would this not be better to overwrite ManageWiki's $wgEnableUploads instead? If it is done after the wgConf extract it should work I think though can't guarantee that.

Universal-Omega (changes requested):

LocalSettings.php
```diff
@@ -4568,10 +4568,12 @@
 $wgHooks['MediaWikiServices'][] = 'extractGlobals';
 
 function extractGlobals() {
-	global $wgConf, $wgDBname, $wgAPIModules;
+	global $wgConf, $wgDBname, $wgAPIModules, $wgRevokePermissions;
```
Universal-Omega, line 4571: +$wgUploadMaintenance

Universal-Omega (changes requested):

LocalSettings.php
```diff
@@ -4568,10 +4568,12 @@
...
 
 	$wgConf->extractAllGlobals( $wgDBname );
-
+	$wgRevokePermissions['*']['upload'] = true;
+		$wgRevokePermissions['*']['movefile'] = true;
```
Universal-Omega, line 4575: Formatting

Universal-Omega (changes requested):

PR #4354, opened 2022-01-07
[22:15] DC-Switch: Disable wiki creation (trust Stews for rename)

RhinosF1:


dmehus (changes requested):
You told me you were going to use `$wgRevokePermissions` rather than revoking the actual user right, which I would prefer

RhinosF1:
> You told me you were going to use `$wgRevokePermissions` rather than revoking the actual user right, which I would prefer

I said that for upload. There's no technical difference in impact.

dmehus:
> > You told me you were going to use `$wgRevokePermissions` rather than revoking the actual user right, which I would prefer
> 
> I said that for upload. There's no technical difference in impact.

Okay, but I also asked you if you'd use that for the other instances, and you indicated you would or would have no objections to that, so since I'd definitely prefer that, and arguably `$wgRevokePermissions` is the way to go anyway, since the way you described it, that nullifies any permissions added by those rights (i.e., cannot be re-added).

RhinosF1:
> > > You told me you were going to use `$wgRevokePermissions` rather than revoking the actual user right, which I would prefer
> > 
> > 
> > I said that for upload. There's no technical difference in impact.
> 
> Okay, but I also asked you if you'd use that for the other instances, and you indicated you would or would have no objections to that, so since I'd definitely prefer that, and arguably `$wgRevokePermissions` is the way to go anyway, since the way you described it, that nullifies any permissions added by those rights (i.e., cannot be re-added).

We don't need that for here. The only people that can add these rights to others have access to edit global permissions which according to Majavah isn't impacted anyway by that variable.

dmehus:
> > > > You told me you were going to use `$wgRevokePermissions` rather than revoking the actual user right, which I would prefer
> > > 
> > > 
> > > I said that for upload. There's no technical difference in impact.
> > 
> > 
> > Okay, but I also asked you if you'd use that for the other instances, and you indicated you would or would have no objections to that, so since I'd definitely prefer that, and arguably `$wgRevokePermissions` is the way to go anyway, since the way you described it, that nullifies any permissions added by those rights (i.e., cannot be re-added).
> 
> We don't need that for here. The only people that can add these rights to others have access to edit global permissions which according to Majavah isn't impacted anyway by that variable.

Right, but that goes to my point. If the point is to restrict Stewards from technically performing a global rename, this would not do that, as Stewards could still add the user right via Special:GlobalGroupPermissions. I'm not suggesting a Steward would do this, but it goes to whether this is //needed// or not. Either you trust Stewards not to perform a global rename or not.

LocalSettings.php
```diff
@@ -2397,7 +2397,7 @@
 				'requestwiki' => true,
 			],
 			'wikicreator' => [
-				'createwiki' => true,
+				'read' => true,
```
Universal-Omega, line 2400: Can't we just comment this with `//` so it can be very easily uncommented later, rather than replacing it with read?

dmehus: > Can't we just comment this with `//` so it can be very easily uncommented later, rather than replacing it with read?

That's not a bad idea, actually. I like this.

PR #4355, opened 2022-01-07
[22:45] DC-Switch: All wikis read only

RhinosF1:


Database.php
```diff
@@ -56,23 +56,13 @@
 		],
 	],
 	'readOnlyBySection' => [
-		// 'DEFAULT' => 'Maintenance ongoing on the database server.',
-		//'c1' => 'Maintenance ongoing on the database server.',
+		'DEFAULT' => 'DC Switchover in progress. Please try again in a few minutes.',
```
Universal-Omega, line 59: this only sets db11 read only, why not use $wgReadOnly?

PR #4365, opened 2022-01-08
Comment out `createwiki` user right

dmehus:
Per suggestion of @Universal-Omega, https://github.com/miraheze/mw-config/pull/4354#discussion_r780708886


dmehus:
Cancelled on second thought and per discussion with @RhinosF1

PR #4368, opened 2022-01-09
Sitenotice for feedback on Category Browser removal

AgentIsai:
This adds a sitenotice for all wikis that have $wgUseCategoryBrowser telling them to let us know what they think and to comment on the CN thread.


Universal-Omega:
I think this causes fatals on wikis that use it. 2 sitenotices can't exist, using the same named function.

PR #4407, opened 2022-01-18
Enable more caching

Universal-Omega:
* set `$wgResourceLoaderMaxage` to default (by removing it being declared here)
* set `$wgRevisionCacheExpiry` to 2 days
* set `$wgPreprocessorCacheThreshold` to 2000

### DIFF
--- a/GlobalCache.php
+++ b/GlobalCache.php
@@ -12,7 +12,7 @@ $wgLanguageConverterCacheType = CACHE_ACCEL;
 $wgParserCacheExpireTime = 86400 * 10;
 $wgDLPQueryCacheTime = 120;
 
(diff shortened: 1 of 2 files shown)


RhinosF1:
We can't enable sidebar cache, it's not compatible with ManageWiki

Universal-Omega:
> We can't enable sidebar cache, it's not compatible with ManageWiki

Why not?

RhinosF1:
> > We can't enable sidebar cache, it's not compatible with ManageWiki
> 
> Why not?

The sidebar varies based on user preferences but it caches the wrong state. It was tried by @Southparkfan but reverted.

PR #4424, opened 2022-01-21
Configure QuickInstantCommons

Universal-Omega:


GlobalExtensions.php
```diff
@@ -33,6 +33,7 @@
 	'OATHAuth',
 	'OAuth',
 	'ParserFunctions',
+	'QuickInstantCommons',
```
RhinosF1, line 36: Is it global?

It only needs to be on certain wikis albeit most.

Universal-Omega: It is a config that disables main functionality, in order to use for miraheze Commons, and shared uploads, the extension itself must be global.

PR #4436, opened 2022-01-24
Betawiki: configure guarded ES cluster

RhinosF1:
In case of issues; do not enable anywhere else.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -3,10 +3,10 @@
 # Per-wiki settings that are incompatible with LocalSettings.php
 switch ( $wi->dbname ) {
 	case 'betawiki':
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -3,7 +3,18 @@
 # Per-wiki settings that are incompatible with LocalSettings.php
 switch ( $wi->dbname ) {
 	case 'betawiki':
-		wfLoadExtension( 'GlobalWatchlist' );
+		wfLoadExtension( [ 'GlobalWatchlist',
```
Universal-Omega, line 6: Formatting

Universal-Omega (changes requested):

PR #4437, opened 2022-01-25
Sitenotice: add notice about ES

RhinosF1:


Universal-Omega:
I don't think that this is necessary and if it is, and this is kept permanently (during ES migrations) then it should be done differently, rather than sitenotice.php in my opinion.

RhinosF1:
> I don't think that this is necessary and if it is, and this is kept permanently (during ES migrations) then it should be done differently, rather than sitenotice.php in my opinion.

This is only set to show up while search is disabled. My plan is to do it in batches so it's not going to be on every wiki at every point.

Universal-Omega:
> > I don't think that this is necessary and if it is, and this is kept permanently (during ES migrations) then it should be done differently, rather than sitenotice.php in my opinion.
> 
> This is only set to show up while search is disabled. My plan is to do it in batches so it's not going to be on every wiki at every point.

I know, but I don't like making this a permanent change, or even a few months. Search being disabled during the enabling of CS is always going to be necessary I believe.

RhinosF1:
Where's a few months coming from?

Universal-Omega:
Since this can't be done right now, closing.

PR #4439, opened 2022-01-25
LS: disable usermerge

RhinosF1:


LocalSettings.php
```diff
@@ -430,9 +430,9 @@
 		'betawiki' => true,
 	],
 	'wgCentralAuthEnableUserMerge' => [
```
Universal-Omega, line 432: The extension default seems to be false so this config can just be removed.

Universal-Omega (changes requested):

LocalSettings.php
29TB, line 432: If we removed the extension, there is no point for that config to be there.

29TB (changes requested):

LocalSettings.php
RhinosF1: We aren't removing CentralAuth

RhinosF1: Nor does this PR remove it, it just disables it.

Universal-Omega: I already mentioned this also in my comment above. The config should probably be removed since its disabled by default in REL1_37 and completely removed in master (1.38)

PR #4447, opened 2022-01-27
Set `$wgCreateWikiDisallowedSubdomains`

Universal-Omega:
`$wgCreateWikiBlacklistedSubdomains` will be replaced with `$wgCreateWikiDisallowedSubdomains`. So set it separately to prevent a minor window of no blacklist existing during updating.


Naleksuh:
@Universal-Omega Wouldn't it make more sense to just set disallowed and then set `$wgCreateWikiBlacklistedSubdomains = $wgCreateWikiDisallowedSubdomains;`

Universal-Omega:
> @Universal-Omega Wouldn't it make more sense to just set disallowed and then set `$wgCreateWikiBlacklistedSubdomains = $wgCreateWikiDisallowedSubdomains;`

Would be, except it won't work due to SiteConfiguration ($wi->config->settings).

PR #4448, opened 2022-01-27
Update LocalSettings.php

Naleksuh:


Universal-Omega:
This should not be done. This will enable Global site CSS/JS, meaning it will make metawiki MediaWiki:Global.css/MediaWiki:Global.js, effect all wikis, from my understanding.

Naleksuh:
> This should not be done. This will enable Global site CSS/JS, meaning it will make metawiki MediaWiki:Global.css/MediaWiki:Global.js, effect all wikis, from my understanding.

Yes, that's the purpose of the page. Doesn't explain why it "should not be done", that explains why it _should_

Universal-Omega:
> > This should not be done. This will enable Global site CSS/JS, meaning it will make metawiki MediaWiki:Global.css/MediaWiki:Global.js, effect all wikis, from my understanding.
> 
> Yes, that's the purpose of the page. Doesn't explain why it "should not be done", that explains why it _should_

We won't interfere with global community wiki site-wide CSS/JS. Doing this will likely need to go through the community.

Naleksuh:
Right now, Global.js is empty but Global.css has some CSS on it. Not sure what it does but it probably belongs there. This needs to be future proofed instead of requiring multiple steps to edit a page.

Universal-Omega:
> Right now, Global.js is empty but Global.css has some CSS on it. Not sure what it does but it probably belongs there. This needs to be future proofed instead of requiring multiple steps to edit a page.

Doing this would allow interface-admins to make global changes also. As for why it has CSS in it, it was added by Dmehus after a request a SN, likely not realizing that we don't use it as we should not make Global style changes. We shouldn't enable this without it at least going through an RfC first. It effects the global community and the level of interference with local communities that can be done from metawiki.

PR #4468, opened 2022-01-31
Move SiteConfiguration setup

Universal-Omega:


initialise/MirahezeFunctions.php
```diff
@@ -28,7 +28,18 @@ class MirahezeFunctions {
 		'wikibeta' => 'betaheze.org',
 	];
 
+	public function __construct() {
+		// Safeguard LocalSettings from being accessed
+		if ( !defined( 'MEDIAWIKI' ) ) {
+			die( 'Not an entry point.' );
+		}
+
+		$this->initialise();
```
lens0021, line 37: Because the name of MirahezeFunctions is ambiguous, I can't expect the creation of an instance will result in some meaningful jobs. Would it be improved?

Universal-Omega: I'm not exactly sure what you mean?

lens0021: With `$wi->initialise();`, we can easily know that it is for initalising something. But with `new MirahezeFunctions;`, we have no clues so we can't figure out what it is for. Maybe the call in this line should moved to LocalSettings.php, or the name of the class should be changed to more meaningful one.

Universal-Omega: Ah yes, this is actually the reason I didn't merge this into the original PR yet, because I plan to move initialisation back to LocalSettings.php.

PR #4471, opened 2022-02-07
GlobalBlocking: disable on non-global wikis

RhinosF1:
Required for 1.38


RhinosF1:
It might be worth doing this for the other 'global' extensions that do cross wiki stuff

Defines.php
```diff
@@ -8,6 +8,7 @@
 	wfLoadExtensions( [
 		'CentralAuth',
 		'GlobalPreferences',
+				'GlobalBlocking',
```
Reception123, line 11: tabbing

PR #4472, opened 2022-02-08
disable sidebarcache for solarawiki per T8732

Reception123:


RhinosF1 (changes requested):
Why not add to the if statement setting it

GlobalCache.php
```diff
@@ -17,7 +17,8 @@
 // This should ideally be patched upstream, converting the hook used
 // to SidebarBeforeOutput rather than SkinBuildSidebar, which is
 // more appropriate for this extension.
-if ( !$wi->config->get( 'wmgUseGroupsSidebar', $wi->dbname ) ) {
+// Disabling Sidebarcahce per wiki as a temporary solution (see T8732)
```
Universal-Omega, line 20: A solution to what this doesn't explain much. Also there is a typo. Ideally I'd rather this be an entirely different if statement for code standards, but I'm not to big on it, so up to you.

Universal-Omega (changes requested):

GlobalCache.php
Reception123: Well that's interesting, one MW Engineer wants it separate and one wants it together.

Universal-Omega: Oh I didn't realise there was something else. To me it just makes more sense to be separate for better explanation of reasoning. The comments don't go good combined like that in my opinion.

GlobalCache.php
```diff
@@ -17,7 +17,8 @@
...
+if ( !$wi->config->get( 'wmgUseGroupsSidebar', $wi->dbname ) || $wgDBname === 'solarawiki' ) {
```
RhinosF1, line 21: This should be and not or

Reception123: Why and? Wouldn't that mean that only solarawiki will have it enabled? Since none of the other wikis would meet the requirement of having GroupsSidebar AND being 'solarawiki'.

Universal-Omega: No it should be or

Universal-Omega: Wait it should be and !==

RhinosF1: Yeah I missed that it was === too.

It should be and not

Universal-Omega:
```suggestion
if ( !$wi->config->get( 'wmgUseGroupsSidebar', $wi->dbname ) && $wgDBname !== 'solarawiki' ) {
```

RhinosF1 (approved):
@Reception123: Can this be merged

Universal-Omega (approved):

PR #4484, opened 2022-02-27
T8846: Update Wikibase.php regarding Benpedia's {{WBREPO}} name

ugochimobi:


RhinosF1 (approved):
Looks ok but Could we do this on a more generic basis at some point so any wiki with WikiBase sets this to the name of the repo wiki?

@Agent-Isai: thoughts?

Reception123 (approved):

Reception123:
> Looks ok but Could we do this on a more generic basis at some point so any wiki with WikiBase sets this to the name of the repo wiki?
> 
> @Agent-Isai: thoughts?

I'm not sure what you mean exactly, but this PR can probably be merged and that can be done after if we want.

RhinosF1:
Yeah no objections to merging, that's why I approved.

PR #4487, opened 2022-03-03
add ArticlePlaceholder settings to ManageWiki per T8818

Reception123:
(done in the Wikibase tab since AP is a Wikibase-related extension.


Universal-Omega (approved):

Universal-Omega (changes requested):
If these aren't already, they actually need added to LS also.

Universal-Omega (approved):
Fixed issues and corrected prior mistaken defaults

PR #4498, opened 2022-03-09
T8710: entityDataCachePaths configurations

ugochimobi:
This might be wrong; I mean, I might have not done it properly, please look and correct appropriately


RhinosF1 (changes requested):
Syntax

ugochimobi:
tbh, I don't know what the correct syntax is, please if you can, check the syntax yourself

Wikibase.php
```diff
@@ -108,6 +108,20 @@
 		// hardcoded internal handling
 		'html',
 	];
+	$wgWBRepoSettings['entityDataCachePaths'] = function() [
```
RhinosF1, line 111: Should be a { instead of [ at the end. Same to close

ugochimobi: I so much thought of that, I was just starring at it lol

RhinosF1 (approved):

Wikibase.php
```diff
@@ -108,6 +108,20 @@
...
+				str_replace( '$1', 'Special:EntityData/{entity_id}.json', $GLOBALS['wgArticlePath'] ),
+				'revision={revision_id}'
+			),
+			// Turtle from Query Service updater
+			wfAppendQuery(
+				str_replace( '$1', 'Special:EntityData/{entity_id}.ttl', $GLOBALS['wgArticlePath'] ),
+				'flavor=dump&revision={revision_id}'
+			),
+		];
+	};
```
RhinosF1, line 124: Not sure the ; is needed

ugochimobi: okay

ugochimobi: welp, as it stands, I think it is needed

Wikibase.php
```diff
@@ -108,6 +108,14 @@
...
 		'html',
 	];
+	$wgWBRepoSettings['entityDataCachePaths'] = [
+		// // JSON from entity page JS, compare wikibase.entityPage.entityLoaded.js
+		'/wiki/Special:EntityData/{entity_id}.json?revision={revision_id}',
+		// Turtle from Query Service updater, compare WikibaseRepository.java
+		'/wiki/Special:EntityData/{entity_id}.ttl?flavor=dump&revision={revision_id}',
+		// third pattern with high volume of requests in Hive, source unknown
+		'/wiki/Special:EntityData?id={entity_id}&revision={revision_id}&format=json',
+	];
```
ugochimobi, line 118: This is how simple it should look, instead of the former.

Universal-Omega (changes requested):
This is equivalent to the default, and unnecessary from what I can see.

ugochimobi:
@Universal-Omega you should probably see this https://phabricator.miraheze.org/T8710

Universal-Omega (commented):
This particular config value is just copied and pasted from Wikimedia config. Can you please say how it'll fix the mentioned task, and why it's necessary for us?

Universal-Omega:
This wouldn't fix the issue, as its basically the same as the default, plus the mentioned task is already resolved.

PR #4501, opened 2022-03-10
ULS settings for GP

ugochimobi:


LocalSettings.php
```diff
@@ -4110,6 +4110,13 @@
 		'wmgUseTranslate' => false,
 		'wmgUseUniversalLanguageSelector' => false,
 	],
+	'wgULSIMEEnabled' => [
+		'default' => true,
+		'gratispaideiawiki' => false,
+	],
+	'wgULSWebfontsEnabled' => [
```
Reception123, line 4117: setting the default only again...

Reception123 (changes requested):

LocalSettings.php
Reception123: This isn't the default value: https://github.com/wikimedia/mediawiki-extensions-UniversalLanguageSelector/blob/master/extension.json#L101

Reception123 (changes requested):

LocalSettings.php
ugochimobi: Done

Reception123 (approved):

Universal-Omega:
The default for `$wgULSWebfontsEnabled` is wrong here. It is false, meaning it doesn't even need to be set here...

ugochimobi:
> The default for `$wgULSWebfontsEnabled` is wrong here. It is false, meaning it doesn't even need to be set here...

The default in extension.json is true? https://github.com/wikimedia/mediawiki-extensions-UniversalLanguageSelector/blob/master/extension.json#L101 
So it isn't wrong

Universal-Omega:
> > The default for `$wgULSWebfontsEnabled` is wrong here. It is false, meaning it doesn't even need to be set here...
> 
> The default in extension.json is true? https://github.com/wikimedia/mediawiki-extensions-UniversalLanguageSelector/blob/master/extension.json#L101 So it isn't wrong

Sorry I had misread and was looking at `$wgULSMobileWebfontsEnabled`.

ugochimobi:
I was only wondering though

PR #4502, opened 2022-03-10
Wikibase: Adding missing setups in favour of GD & GP

ugochimobi:


RhinosF1 (dismissed):

LocalSettings.php
```diff
@@ -4256,8 +4258,118 @@
...
 	],
+	'wmgWikibaseAllowLocalShortDesc' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseForceLocalShortDesc' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseEnableData' => [
```
Reception123, line 4269: Again, why is it necessary to set the default here unless 1) the setting is in ManageWiki 2) you wish to set something different for your wiki?

LocalSettings.php
```diff
@@ -4256,8 +4258,118 @@
...
+		'default' => [],
+		'gratisdatawiki' => [
+			NS_CATEGORY,
+			NS_PROJECT,
+			NS_TEMPLATE,
+			NS_HELP,
+			828, // NS_MODULE
+		],
+	],
+	'wmgWBRepoIdGenerator' => [
```
Reception123, line 4290: same question as above

LocalSettings.php
```diff
@@ -4256,8 +4258,118 @@
...
+		'default' => [],
+		'gratispaideiawiki' => [
+			'url' => 'https://static.miraheze.org/commonswiki/a/a4/GDechoIcon.svg',
+		],
+	],
+	'wmgEchoEnablePush' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	// Only enable this ('wmgWikibaseClientTrackLuaFunctionCallsPerWiki') conservatively, having this on to many
```
Reception123, line 4307: this comment seems to be WMF specific

LocalSettings.php
```diff
@@ -4256,8 +4258,118 @@
...
+	// wikis will take *a lot* of Graphite storage space.
+	'wmgWikibaseClientTrackLuaFunctionCallsPerWiki' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseClientTrackLuaFunctionCallsPerSiteGroup' => [
```
Reception123, line 4313: same question as above

Reception123 (changes requested):

LocalSettings.php
ugochimobi: @Reception123 some of the configs are different for my wiki?

LocalSettings.php
ugochimobi: WMF specific? if I got you right, My reply would be, because I saw the comment somewhere on one example config file, _can't remember where again_

LocalSettings.php
RhinosF1: Given in WikiBase.php we just set $wg = $wmg, why not just set the wg version here?

ugochimobi: Tbh, I think the wg versions should better be set in our Wikibase.php file, tbqh

LocalSettings.php
```diff
@@ -4256,8 +4258,122 @@
...
+	'wgWBQualityConstraintsEnableConstraintsCheckJobsRatio' => [
+		'default' => null,
+		'gratisdatawiki' => 100,
+	],
+	'wgWBQualityConstraintsFormatCheckerShellboxRatio' => [
+		'default' => null,
+		'gratisdatawiki' => 1,
+	],
+	'wgWBQualityConstraintsTypeCheckMaxEntities' => [
+		'default' => null,
```
Reception123, line 4374: Not default: https://github.com/wikimedia/mediawiki-extensions-WikibaseQualityConstraints/blob/master/extension.json#L190

LocalSettings.php
```diff
@@ -4256,8 +4258,122 @@
...
+			'P531',
+			'P327',
+			'P532',
+			'P533',
+		],
+	],
 
 	// WikibaseQualityConstraints
+	'wgWBQualityConstraintsEnableConstraintsCheckJobs' => [
+		'default' => null,
```
Reception123, line 4362: not correct default, can't be 'null', https://github.com/wikimedia/mediawiki-extensions-WikibaseQualityConstraints/blob/master/extension.json#L220

LocalSettings.php
```diff
@@ -4256,8 +4258,122 @@
...
+	'wmgWikibaseClientTrackLuaFunctionCallsPerWiki' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseClientTrackLuaFunctionCallsPerSiteGroup' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseClientAddEntityUsagesBatchSize' => [
+		'default' => null,
```
Reception123, line 4320: not the default

LocalSettings.php
Reception123: the default is already set to true so this is not needed

LocalSettings.php
Reception123: What I mean is that we don't have Graphite, meaning that the comment isn't useful/relevant to Miraheze

LocalSettings.php
```diff
@@ -4256,8 +4258,122 @@
...
+	'wmgWBRepoIdGenerator' => [
+		'default' => '',
+		'gratisdatawiki' => 'mysql-upsert',
+	],
+	'wmgWBRepoIdGeneratorInErrorPingLimiter' => [
+		'default' => null,
+		'gratisdatawiki' => 9,
+	],
+	'wmgWikibaseClientEchoIcon' => [
+		'default' => [],
```
Reception123, line 4300: not default (afaik)

LocalSettings.php
```diff
@@ -4256,8 +4258,122 @@
...
+	'wmgWikibaseForceLocalShortDesc' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseEnableData' => [
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wmgWikibaseEntityAccessLimit' => [
+		'default' => null,
```
Reception123, line 4274: not the default

Reception123 (changes requested):

ugochimobi:
This is fixed and ready now.

Universal-Omega (changes requested):
Almost everything you are setting here either don't exist as configuration variables, or are already in ManageWiki and setting them here will have no effect.

ugochimobi:
> Almost everything you are setting here either don't exist as configuration variables, or are already in ManageWiki and setting them here will have no effect.

That's not correct, I agree that are in ManageWiki, but I intentionally wanted to set them here, but if it doesn't make any difference then I propose removal leaving the rest. But every single other variable exists big time.

Universal-Omega:
> > Almost everything you are setting here either don't exist as configuration variables, or are already in ManageWiki and setting them here will have no effect.
> 
> That's not correct, I agree that are in ManageWiki, but I intentionally wanted to set them here, but if it doesn't make any difference then I propose removal leaving the rest. But every single other variable exists big time.

No they don't, all. The `wmg`  variables never exist, those are always custom variables. Setting the ManageWiki ones from here will have no effect at all.

ugochimobi:
> > > Almost everything you are setting here either don't exist as configuration variables, or are already in ManageWiki and setting them here will have no effect.
> > 
> > 
> > That's not correct, I agree that are in ManageWiki, but I intentionally wanted to set them here, but if it doesn't make any difference then I propose removal leaving the rest. But every single other variable exists big time.
> 
> No they don't, all. The `wmg` variables never exist, those are always custom variables. Setting the ManageWiki ones from here will have no effect at all.

The `wmgWikibaseClientEchoIcon`? The `wmgWBRepoPreferredGeoDataProperties` ? Even the `wmgEchoEnablePush` and `wmgWBRepoPreferredPageImagesProperties`? Nah, that's not correct, you can even see them at Wikimedia initialise.php.

Universal-Omega:
> > > > Almost everything you are setting here either don't exist as configuration variables, or are already in ManageWiki and setting them here will have no effect.
> > > 
> > > 
> > > That's not correct, I agree that are in ManageWiki, but I intentionally wanted to set them here, but if it doesn't make any difference then I propose removal leaving the rest. But every single other variable exists big time.
> > 
> > 
> > No they don't, all. The `wmg` variables never exist, those are always custom variables. Setting the ManageWiki ones from here will have no effect at all.
> 
> The `wmgWikibaseClientEchoIcon`? The `wmgWBRepoPreferredGeoDataProperties` ? Even the `wmgEchoEnablePush` and `wmgWBRepoPreferredPageImagesProperties`? Nah, that's not correct, you can even see them at Wikimedia initialise.php.

We are not wikimedia. We don't have those variables existing on Miraheze.

PR #4504, opened 2022-03-11
Update Wikibase.php

ugochimobi:


RhinosF1 (changes requested):
Are any of these $wmg* set

ugochimobi:
> Are any of these $wmg* set

Yes, they're set in LocalSettings.php

ugochimobi:
But the PR hasn't been merged yet #4502

RhinosF1:
See #4502

This isn't needed.

ugochimobi:
I thought as much anyway

ugochimobi:
Although, there are some configs that are now wmg, that needs to be set here....

PR #4505, opened 2022-03-12
Update with GeoData configs

ugochimobi:


ugochimobi:
Please Run `updateIndexGranularity.php` after merging

LocalSettings.php
```diff
@@ -1507,6 +1507,72 @@
...
+		'default' => 1000,
+		'gratispaideiawiki' => 2000,
+		'gratidatawiki' => 2000,
+	],
+	// Controls what GeoData should do when it encounters some problem
+	// Reaction type:
+	// 'track' - Add tracking category
+	// 'fail' - Consider the tag invalid, display message and add tracking category
+	// 'none' - Do nothing
+	'wgGeoDataWarningLevel' => [
```
Universal-Omega, line 1532: Why are we only.setting default? If it isn't custom, then this shouldn't be added, and left at extension default.

Universal-Omega (changes requested):
Not extensively reviewed, and defaults not checked. But one thing I noticed.

LocalSettings.php
ugochimobi: @Universal-Omega this should be ready now

RhinosF1 (changes requested):
We don't support http traffic

ugochimobi:
This should do @RhinosF1

LocalSettings.php
```diff
@@ -1507,6 +1507,40 @@
 			'mode' => 'packed',
 		],
 	],
+	
+	// GeoData
+	'wgGlobes' => [
+		'default' => [],
+		'gratisdatawiki' => [
+			'https://gratisdata.miraheze.org/entity/Q476' => 'earth',
```
Universal-Omega, line 1515: Where did you find documentation for this? I can't, and I'm not sure it supports Wikibase entities without first knowing that.

ugochimobi:
Wait, I think I got it wrong, I was supposed to set the globes here and set
the gratisdata entities at Wikibase.php.

The docs? No, I know that it is possible to set these configs at least from
my understanding and plus the mediawiki extension page.

Universal-Omega:
Sure the config exists, but I can't find any examples in Wikimedia config, in the extension documentation, or in extension code that makes me believe it supports wikibase entities like this.

ugochimobi:
No, I just realized it doesn't support wikibase entities like this, just
the arrays.
The wikibase entities should instead be set from Wikibase.php, you see how
it looks like from Wikimedia wikibase.php config (currently on mobile so
can't link to it).

Universal-Omega:
Per the comments, the entirety of this PR won't work.

PR #4506, opened 2022-03-12
T8924: setting wgPopupsReferencePreviewsBetaFeature to false

ugochimobi:


Reception123:
Setting exists in ManageWiki (https://github.com/miraheze/mw-config/blob/9e29df798b5455d935a01be588dbdf0196335adc/ManageWikiSettings.php#L303). Please check before opening a PR.

PR #4507, opened 2022-03-13
T8925: setting Tweeki's subnav to false on obeymewiki

ugochimobi:


Reception123 (approved):

Universal-Omega:
This is the wrong format for config... default should be the key used, not subnav as first key...

ugochimobi:
I think I resolved it at #4514

PR #4508, opened 2022-03-14
necessary wikibase settings for gp/gd

ugochimobi:


Wikibase.php
```diff
@@ -155,6 +155,20 @@
 		],
 	];
 	$wgWBRepoSettings['allowEntityImport'] = false;
+	$wgWBRepoSettings['pagePropertiesRdf'] = [
```
Universal-Omega, line 158: This is unnecessary, seems to be equal to the default.

Universal-Omega (changes requested):

ugochimobi:
Fixed that now

Universal-Omega (approved):

PR #4509, opened 2022-03-14
Popups config for gratispaideiawiki

ugochimobi:


Reception123:
You forgot to add the default for all the settings you are adding.

LocalSettings.php
```diff
@@ -3311,6 +3311,18 @@
 	'wgPopupsOptInDefaultState' => [
 		'default' => 0,
 	],
+	'wgPopupsConflictingNavPopupsGadgetName' => [
+                'default' => '',
```
Reception123, line 3315: https://github.com/wikimedia/mediawiki-extensions-Popups/blob/master/extension.json#L67

LocalSettings.php
```diff
@@ -3311,6 +3311,18 @@
...
+		'gratispaideiawiki' => 'Navigation_popups',
+	],
+	'wgPopupsConflictingRefTooltipsGadgetName' => [
+                'default' => '',
```
Reception123, line 3319: https://github.com/wikimedia/mediawiki-extensions-Popups/blob/master/extension.json#L71

LocalSettings.php
```diff
@@ -3311,6 +3311,18 @@
...
+		'gratispaideiawiki' => 'ReferenceTooltips',
+	],
+	'wgPopupsOptInStateForNewAccounts' => [
+                'default' => null,
```
Reception123, line 3323: Already the default setting; no need for this (https://github.com/wikimedia/mediawiki-extensions-Popups/blob/master/extension.json#L63)

Reception123 (changes requested):

ugochimobi:
@Reception123 This should be done now

Reception123:
@ugochimobi Sorry you keep getting confused about this. Let me try to explain again.

The default value is the one that exists in the extensions extension.json, and unless it is changed that is what it will be for all Miraheze wikis. If a wiki wants to have a different value than the one that is default that will have to be added, but the default value will also have to be.

So you need to do both the default (as it's shown in extension.json) AND the value you need for your wiki.

Reception123:
no, just like you did https://github.com/miraheze/mw-config/commit/ab883efb06146b2a9d1c28fa187f1acf6e4cec8e

ugochimobi:
okay, but if want I want is also the default, then I should add just default right?

Reception123:
> okay, but if want I want is also the default, then I should add just default right?

No, because the default already applies because of extension.json, it doesn't need to be added separately to LS

ugochimobi:
Ah, I don't know why I am doubting that, because there are so many things that are supposed to be happening (apart from this PR) which aren't and it's causing me to doubt

LocalSettings.php
```diff
@@ -3311,6 +3311,15 @@
...
 		'default' => 0,
 	],
+	'wgPopupsConflictingNavPopupsGadgetName' => [
+                'default' => 'Navigation_popups',
+	],
+	'wgPopupsConflictingRefTooltipsGadgetName' => [
+                'default' => 'ReferenceTooltips',
+	],
+	'wgPopupsOptInStateForNewAccounts' => [
+                'default' => 1,
```
Universal-Omega, line 3321: This should be a string.

Universal-Omega (changes requested):
+this is pointless, we are adding configs, setting only defaults that would already be done by the extension.

PR #4510, opened 2022-03-14
T8932: ApprovedRevs configs

ugochimobi:
Should I set the defaults? (I didn't because it's not available from managewiki) or It is better this way (setting it per wiki)?


Reception123 (approved):

Universal-Omega:
Defaults are wrong here....

PR #4515, opened 2022-03-18
Update LocalSettings.php with MFE configs

ugochimobi:


LocalSettings.php
```diff
@@ -2935,8 +2936,9 @@
 	],
 	'wgMFNoindexPages' => [
 		'wmgUseMobileFrontend' => false,
-		'gratisdatawiki' => true,
-		'gratispaideiawiki' => true,
+		'default' => true,
```
Universal-Omega, line 2939: No. This should not be done..the default is already false, by the wmgUseMobileFrontend part of this. Which makes it only set config if MobileFrontend is enabled.

Universal-Omega (changes requested):
+conflict. Not extensively reviewed.

ugochimobi:
@Universal_Omega, the 'default' should be removed now?

LocalSettings.php
ugochimobi: The default is supposed to be true according to extension.json https://github.com/wikimedia/mediawiki-extensions-MobileFrontend/blob/master/extension.json#L870
So what I want is false for my wikis.

ugochimobi:
@Universal-Omega I don't know how to resolve the other conflict

LocalSettings.php
RhinosF1: This does the opposite

ugochimobi: @RhinosF1 how do you mean?

ugochimobi:
Well, think I'd have to open separate PRs for each config as I actually don't know what causes the conflict.

PR #4521, opened 2022-03-20
MF configs for gratispaideiawiki and gratidatawiki

ugochimobi:


LocalSettings.php
```diff
@@ -2975,6 +2975,81 @@
...
+			'tagline' => false,
+		],
+		'gratisdatawiki' => [
+			'search' => true,
+			'nearby' => true,
+			'watchlist' => true,
+			'tagline' => false,
+		],
+	],
+	'wgMFNearby' => [
```
Universal-Omega, line 3043: I don't think we support this functionality. Even if we do, I'd be somewhat concerned with it as it geo locates users, so I have some privacy concerns here as well. Unless I'm not understanding it.

Universal-Omega (changes requested):
I have not completely reviewed this. But please provide a reasoning/explanation you need each config enabled. This is just to verify you understand what you are requesting, as some of these configs are complex usage, as well as it complicates our configuration in a way that if the upstream defaults are changed in any way, we have to verify ours are as well.

ugochimobi:
I don't think you got it right, firstly, this Nearby is used in Wikimedia
wikis, (that's not a criteria, but) Secondly, when a user wants to use the
nearby feature to look for articles close to them, the system sends them a
prompt to grant nearby their location, so it's only possible to be used
when a user grants the feature, then it can't work.

So this is a plus.

Universal-Omega:
It would like confirmation from someone else that such functionality is allowed, and just because Wikimedia supports it doesn't mean we do, they have way more resources to support more things.

ugochimobi:
It exists, I have used it on Gratispaideia before, not until the config was
reverted by only God knows who.
And I don't think it uses resources per se, it's a normal config like every
other config that comes along with the MobileFrontend.

Universal-Omega:
I was the one who reverted it last time, due to wgNearby usage without confirmation we support it, as well as the fact that PR caused fatals on MF on every wiki using it.

ugochimobi:
Ah, but I believe when you go to English Wikipedia you'd find it there
(like I earlier said, it's not a criteria that we should use too) but you
could do your own research, because AFAIK it cause no harm at all.

Universal-Omega:
Also, all nearby functionality is being removed from MobileFrontend, so I'm not we should do that at all anyway. https://phabricator.wikimedia.org/T246494

ugochimobi:
Welp this is interesting if it happens. Welp, I think nearby config should
be removed then. Leaving the remaining to be merged.

LocalSettings.php
```diff
@@ -2975,6 +2975,71 @@
...
+		'gratisdatawiki' => [
+			'wbetterms' => 'label',
+		],
+	],
+	'wgMFSearchGenerator' => [
+		'default' => [
+			'name' => 'prefixsearch',
+			'prefix' => 'ps',
+		],
+		'gratisdatawiki' => [
```
Universal-Omega, line 2999: I think this should use `+` to inherit the default as well?

LocalSettings.php
```diff
@@ -2975,6 +2975,71 @@
...
+		],
+		'gratisdatawiki' => [ 
+			'entityterms',
+		],
+	],
+	'wgMFSearchAPIParams' => [
+		'default' => [
+			'ppprop' => 'displaytitle',
+		],
+		'gratisdatawiki' => [
```
Universal-Omega, line 2990: I think this should use `+` to inherit the default as well?

LocalSettings.php
```diff
@@ -2975,6 +2975,71 @@
 			'amc' => true,
 		],
 	],
+	'wgMFQueryPropModules' => [
+		'default' => [ 
+			'pageprops', 
+		],
+		'gratisdatawiki' => [
```
Universal-Omega, line 2982: I think this should use `+` to inherit the default as well?

LocalSettings.php
```diff
@@ -2975,6 +2975,71 @@
...
+		'default' => [
+			'name' => 'prefixsearch',
+			'prefix' => 'ps',
+		],
+		'gratisdatawiki' => [
+			'name' => 'wbsearch',
+			'prefix' => 'wbs',
+		],
+	],
+	'wgMFVaryOnUA' => [
```
Universal-Omega, line 3004: Can you please provide an explanation for your reasoning for this one?

Universal-Omega (changes requested):

Universal-Omega:
Alot of these could also probably be added to ManageWiki.

LocalSettings.php
Universal-Omega: > CAUTION: Setting this to true in combination with a (frontend)caching layer (such as Varnish) can have a huge impact on how your caching works, as it now caches every single page multiple times for any possible/different User Agent string!

And we do use Varnish...

ugochimobi:
> Alot of these could also probably be added to ManageWiki.

Well, per Reception123, a lot of configs shouldn't be added to ManageWiki since the use case is very few, and mostly because it's might affects the performances of ManageWiki in general, which seems understandable.

LocalSettings.php
ugochimobi: > > WARNING: You should set this to true only, if you know what you're doing!
> 
> Can you please provide an explanation for your reasoning for this one?
And this is the reason why I'm enabling it on Gratispaideia, the statement you quoted below.

LocalSettings.php
ugochimobi: > I think this should use `+` to inherit the default as well?

I don't think so, but I would, later on when what I expect on gratisdatawiki is not met.

LocalSettings.php
Universal-Omega: It says that on README docs. As for the other thing with Varnish, we really shouldn't do that. I think that wouldn't be great either.

LocalSettings.php
```diff
@@ -2975,6 +2975,71 @@
...
+		'default' => false,
+		'gratispaideiawiki' => true,
+	],
+	'wgMFUseWikibase' => [
```
Universal-Omega, line 3008: It seems like this was actually removed years ago (2018), but never from extension.json or any documentation? Seems to have been removed in https://github.com/wikimedia/mediawiki-extensions-MobileFrontend/commit/304c0b0923a263386c54f2a484b8a10a26cb094c.

Universal-Omega (changes requested):

LocalSettings.php
ugochimobi: > It says that on README docs. As for the other thing with Varnish, we really shouldn't do that. I think that wouldn't be great either.

Setting this true isn't a problem at all as far as I have read, plus, caching every single page multiple times isn't alarming at all.

Universal-Omega: There is multiple warnings about this variable on docs. So i'm not comfortable deploying it if you can't provide a good explanation for needing it.

ugochimobi: > There is multiple warnings about this variable on docs. So i'm not comfortable deploying it if you can't provide a good explanation for needing it.

The fact that it will improve the seo of the wiki is the reason it should be deployed, as a matter of fact, that's one aim of an Encyclopedia (doing well in seo)

LocalSettings.php
ugochimobi: > It seems like this was actually removed years ago (2018), but never from extension.json or any documentation? Seems to have been removed in [wikimedia/mediawiki-extensions-MobileFrontend@304c0b0](https://github.com/wikimedia/mediawiki-extensions-MobileFrontend/commit/304c0b0923a263386c54f2a484b8a10a26cb094c).

No, this wasn't removed at all. A few was removed which I can't mention now cuz I'm mobile but not this.

Universal-Omega: Usage of this variable literally does not exist in the extension anymore.

ugochimobi: > Usage of this variable literally does not exist in the extension anymore.

Okay, remove it then.

LocalSettings.php
Universal-Omega: This should be removed from this PR, if you wish it to be merged in a timely manner. I'm not comfortable with it with all the warnings, and as such won't be comfortable with deploying this at all, at this time, with this setting.

ugochimobi:
Would you mind removing it, I'm currently mobile.

Universal-Omega (approved):

PR #4522, opened 2022-03-25
Disallow `centralauth-suppress` right in ManageWiki

Universal-Omega:
Renamed from centralauth-oversight in 1.38, this can be done whenever, even before 1.38.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2572,6 +2572,7 @@ $wi->config->settings += [
 				'centralauth-createlocal',
 				'centralauth-lock',
 				'centralauth-oversight',
+				'centralauth-suppress',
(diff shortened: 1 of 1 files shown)


RhinosF1:
Is there anywhere else oversight is used?

Universal-Omega:
> Is there anywhere else oversight is used?

Yes, in assigning it, but that shouldn't be done until after 1.38, but this just preemptively blacklists it.

> Is there anywhere else oversight is used?

Yes, in assigning it, but that can't be done until after 1.38, so this just blacklists it as a precaution ahead of time.

RhinosF1:
Ack. I'll add a note later to the tracking task.

PR #4523, opened 2022-03-26
Add testwikiwiki to loginwiki's import sources

dmehus:
Required for transwiki imports on Loginwiki

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1803,6 +1803,9 @@ $wi->config->settings += [
 			'wmincubator',
 			'wikiaincubatorplus',
 		],
+		'+loginwiki' => [
+			'testwikiwiki',
+		],
 		'+memedatawiki' => [
 			'd',
 			'fd',


RhinosF1 (commented):
Has anyone requested this for a specific purpose?

dmehus:
> Has anyone requested this for a specific purpose?

Me.

RhinosF1:
Fair enough.

Do we know if testwikiwiki has been privacy reviewed? (Not that I don't trust @MacFan4000 )

I'm not entirely sure how this works with cross-wiki interaction and whether it would allow access to PII when it imports.

dmehus:
> Fair enough.
> 
> Do we know if testwikiwiki has been privacy reviewed? (Not that I don't trust @MacFan4000 )
> 
> I'm not entirely sure how this works with cross-wiki interaction and whether it would allow access to PII when it imports.

I don't see why the privacy policy would come in to play. That's never been a requirement to check for adding transwiki import sources. I'm not requesting that `testwiki.wiki` be added to the CSP whitelist; this is //just// for importing templates as needed.

RhinosF1 (approved):
Logic in https://github.com/wikimedia/mediawiki/blob/5357695270161bce1f6dbaccc96645791f17a013/includes/import/ImportStreamSource.php#L114 definately doesn't pass any user info when doing the import so \o/

Universal-Omega (approved):

PR #4527, opened 2022-03-27
Add bureaucrat and consul protection/restriction level user rights to…

dmehus:
… testwiki

So Consuls can grant these protection/restriction level user rights to subordinate user groups as required

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2504,15 +2504,6 @@ $wi->config->settings += [
 				'editor' => true,
 			],
 		],
-		'+testwiki' => [
-			'consul' => [
-				'consul' => true,
-				'bureaucrat' => true,
-			],
-			'bureaucrat' => [
-				'bureaucrat' => true,
-			],
-		],
 		'+thesciencearchiveswiki' => [
 			'sysop' => [
 				'templateeditor' => true,
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3661,6 +3661,10 @@ $wi->config->settings += [
 			'editfragment',
 			'edittemplate',
 		],
+		'testwiki' => [
+			'bureaucrat',
+			'consul',
+		],
 		'+wmgUseSocialProfile' => [
 			'updatepoints',
 		],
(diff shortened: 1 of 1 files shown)


RhinosF1:
Consensus?

Universal-Omega:
They should be removed from $wgManageWikiPermissionsAdditionalRights if they are being added to $wgAvailableRights. The only difference is that with this, now they can be assigned via ManageWiki, whereas the ManageWiki config ones can't be.

dmehus:
> Consensus?

Not needed. TestWiki operates on the basis of community discussions and Consul discretion.

dmehus:
> They should be removed from $wgManageWikiPermissionsAdditionalRights if they are being added to $wgAvailableRights. The only difference is that with this, now they can be assigned via ManageWiki, whereas the ManageWiki config ones can't be.

That's fair. I didn't know that. I will add that. Thank you!

RhinosF1:
> 

Is there any discussion on testwiki or indication of why now?

I don't see what subordinate group needs it.

dmehus:
> > 
> 
> Is there any discussion on testwiki or indication of why now?
> 
> I don't see what subordinate group needs it.

`bot`, potentially, to allow approved bots to edit Consul-protected pages within their approved parameters. But the key is, Consuls would still be deciding which user groups should have it

RhinosF1:
Is there a bot that needs it?

Universal-Omega:
> > 
> 
> Is there any discussion on testwiki or indication of why now?
> 
> I don't see what subordinate group needs it.

This doesn't need discussion, I wouldn't think, it has no functional changes, as the restriction levels already exist, this only makes the rights available to assign in ManageWiki rather than a particular forced group, though now that I think about it, why? These rights won't be assigned to any groups besides the ones of the same name I think, which they already are.

dmehus:
> Is there a bot that needs it?

In this case, yes, https://publictestwiki.com/wiki/TestWiki:Bot_approval_requests (@Agent-Isai's `bot`). This wouldn't grant any Consul user rights, other than the ability to edit Consul-protected pages, with the mandate approved by a Consul.

RhinosF1 (approved):
approved with my sysadmin hat and endorsed with my consul hat. (Some of those questions had a bit of both hats on)

Universal-Omega (approved):

PR #4532, opened 2022-03-29
Update Wikibase.php

ugochimobi:
Please merge appropriately after reviewing. Find my explanation for each variable below per @Universal-Omega

### preferredPageImagesProperties
The property IDs listed here are those that should be considered for the page property `page_image`.
### preferredGeoDataProperties
The only difference is that the properties listed here are taken into account when determining primary coordinates for the GeoData extension on an entity.
### sendEchoNotification
As I set this to true, users on gratispaideia and/or any other client wiki will be notified when a page they created is linked to a repository item.
### echoIcon
This one is related to the `sendEchoNotification`; it only has an effect when it is set to true, otherwise, it is ineffectual. It is the icon that will be visible along with the message when the alert is sent to the user.
### allowLocalShortDesc
Setting this to true will enable local override of the central description with the `{{SHORTDESC:}}` parser function. Just for a note; the central description is the description that is added to the wikibase item. So Instead of displaying that setting this to true will make it display the short description of the article.
### forceLocalShortDesc
This one simply forces it, when it's refusing by any chance

### DIFF
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -196,7 +196,7 @@ if ( $wgDBname === 'gratispaideiawiki' ) {
 	];
 	$wgWBClientSettings['allowLocalShortDesc'] = true;
 	$wgWBClientSettings['forceLocalShortDesc'] = true;
-	$wgWBClientSettings['allowDataTransclusion'] = true;
+	$wgWBClientSettings['showExternalRecentChanges'] = true;
 }
 if ( $wgDBname === 'benpediawiki' ) {
 	$wgWBClientSettings['repoSiteName'] = 'Gratisdata';
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -196,7 +196,6 @@ if ( $wgDBname === 'gratispaideiawiki' ) {
 	];
 	$wgWBClientSettings['allowLocalShortDesc'] = true;
 	$wgWBClientSettings['forceLocalShortDesc'] = true;
-	$wgWBClientSettings['showExternalRecentChanges'] = true;
 }
 if ( $wgDBname === 'benpediawiki' ) {
 	$wgWBClientSettings['repoSiteName'] = 'Gratisdata';
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -140,6 +140,36 @@ if ( $wgDBname === 'gratisdatawiki' ) {
 		],
 	];
 	$wgWBRepoSettings['allowEntityImport'] = false;
+	$wgWBRepoSettings['preferredPageImagesProperties'] = [
+		// Photos
+		'P386',
+		'P520',
+		'P521',
+		'P522',
+		'P523',
+		'P524',
+		// Complex graphics
+		'P135',
+		'P136',
+		'P387',
+		'P525',
+		// Simple graphics
+		'P526',
+		'P527',
+		'P528',
+		'P470',
+		'P529',
+		// Multi page content
+		'P530',
+		// Maps
+		'P531',
+		'P327',
+		'P532',
+		'P533',
+	];
+	$wgWBRepoSettings['preferredGeoDataProperties'] = [
+		'P134',
+	];
 }
 
 if ( $wgDBname === 'gratispaideiawiki' ) {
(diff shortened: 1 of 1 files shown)


Wikibase.php
```diff
@@ -160,6 +190,13 @@
 	$wgWBClientSettings['linkItemTags'] = [
 		'client-linkitem-change'
 	];
+	$wgWBClientSettings['sendEchoNotification'] = true;
+	$wgWBClientSettings['echoIcon'] = [
+		'url' => 'https://static.miraheze.org/commonswiki/a/a4/GDechoIcon.svg',
+	];
+	$wgWBClientSettings['allowLocalShortDesc'] = true;
+	$wgWBClientSettings['forceLocalShortDesc'] = true;
+	$wgWBClientSettings['showExternalRecentChanges'] = true;
```
Universal-Omega, line 199: This is the default.

Universal-Omega (changes requested):

Wikibase.php
ugochimobi: Ah, no wonder I could see it from recent changes, but was still doubting.

Universal-Omega:
Note that the shortdesc configs will also possibly conflict with some of the other description extensions (maybe Extension:ShortDescription at least) so should confirm its disabled first before this is merged.

Universal-Omega (approved):

ugochimobi:
> Note that the shortdesc configs will also possibly conflict with some of the other description extensions (maybe Extension:ShortDescription at least) so should confirm its disabled first before this is merged.

Wait, what? Please help me suggest a commit removing the short description variable. I can't afford to disable the short description extension, although it can be added later because I'm not sure it will really conflict, according to the way it works. But just suggest a commit removing it.

ugochimobi:
> > > Note that the shortdesc configs will also possibly conflict with some of the other description extensions (maybe Extension:ShortDescription at least) so should confirm its disabled first before this is merged.
> > 
> > 
> > Wait, what? Please help me suggest a commit removing the short description variable. I can't afford to disable the short description extension, although it can be added later because I'm not sure it will really conflict, according to the way it works. But just suggest a commit removing it.
> 
> The ShortDescription extension is designed to do this behaviour, that this configuration does I believe.

Actually, no, that's not what this config does.
The config is a little self explanatory per it's name; it simply enables local override of the central description with {{SHORTDESC:}}. That it, if the central description of `Article A` is `this is article` and we want it to be in something else in only one clientwiki, then enabling this will make it possible to use {{SHORTDESC:}} to locally change it to `this is an article` whereas the central description still exists.

ugochimobi:
> > > Note that the shortdesc configs will also possibly conflict with some of the other description extensions (maybe Extension:ShortDescription at least) so should confirm its disabled first before this is merged.
> > 
> > 
> > Wait, what? Please help me suggest a commit removing the short description variable. I can't afford to disable the short description extension, although it can be added later because I'm not sure it will really conflict, according to the way it works. But just suggest a commit removing it.
> 
> The ShortDescription extension is designed to do this behaviour, that this configuration does I believe. So to keep that probably remove `allowLocalShortDesc` and `forceLocalShortDesc` from this I think.

Thank you for clarifying, lol, so please help me do a suggestion commit because I'm xurr

Universal-Omega:
> Plus, the {{SHORTDESC:}} parser function comes along with the short description extension.

That's the conflict... because it does, using both is not possible, I don't think.

Universal-Omega:
> > > Plus, the {{SHORTDESC:}} parser function comes along with the short description extension.
> > 
> > 
> > That's the conflict... because it does, using both is not possible, I don't think.
> 
> Okay, see my comment above

The only options are disable the ShortDescription extension, or remove setting those 2 configs I believe.

ugochimobi:
> > > > Plus, the {{SHORTDESC:}} parser function comes along with the short description extension.
> > > 
> > > 
> > > That's the conflict... because it does, using both is not possible, I don't think.
> > 
> > 
> > Okay, see my comment above
> 
> The only options are disable the ShortDescription extension, or remove setting those 2 configs I believe.

Yeah, I can not afford to disable the short description extension, lol, so let's remove the last two variables.

ugochimobi:
Will you make the suggestion commit?

PR #4534, opened 2022-03-29
GeoData config for gratisdata

ugochimobi:
This is a follow-up of #4505. This PR gives us the proper way it should look like.

This is a lengthy array of each globe's parameters; they are the geographical entities on which the coordinate from GeoData is located. Having these arrays set and having #4535 set too will make them more effectual when using the {{#coordinates:}} parser function. Full explanation can be found on the extension page.


RhinosF1:
@ugochimobi how does this meet @Universal-Omega's advice to explain your actions?

ugochimobi:
> @ugochimobi how does this meet @Universal-Omega's advice to explain your actions?

Oops, I think I missed this one, let me update the original PR text field now.

Universal-Omega:
Does `updateIndexGranularity.php` need ran if this is done still?

Universal-Omega:
Does `updateIndexGranularity.php` need ran if this is done still?

ugochimobi:
> Does `updateIndexGranularity.php` need ran if this is done still?

The extension page says that

Universal-Omega:
> > Does `updateIndexGranularity.php` need ran if this is done still?
> 
> The extension page says that

This isn't changing `$wgGeoDataIndexGranularity` though, which is the only one the extension page says it needs ran for.

ugochimobi:
> > > Does `updateIndexGranularity.php` need ran if this is done still?
> > 
> > 
> > The extension page says that
> 
> This isn't changing `$wgGeoDataIndexGranularity` though, which is the only one the extension page says it needs ran for.

That's true

PR #4535, opened 2022-03-29
Set globe for coordinates added to GeoData

ugochimobi:
This is a follow-up to #4534. It has a mapping from URIs to names that GeoData recognizes.


Wikibase.php
```diff
@@ -140,8 +140,36 @@
...
+		'http://gratisdata.miraheze.org/entity/Q2122' => 'miranda',
+		'http://gratisdata.miraheze.org/entity/Q2117' => 'ariel',
+		'http://gratisdata.miraheze.org/entity/Q2129' => 'umbriel',
+		'http://gratisdata.miraheze.org/entity/Q2128' => 'titania',
+		'http://gratisdata.miraheze.org/entity/Q2125' => 'oberon',
+		'http://gratisdata.miraheze.org/entity/Q982' => 'triton',
+		'http://gratisdata.miraheze.org/entity/Q2123' => 'pluto',
+	];
 }
-
```
Universal-Omega, line 144: Why remove this line?

Universal-Omega (dismissed):

Wikibase.php
ugochimobi: Guess that was a mistake I didn't took note of, let me add

Universal-Omega (approved):

PR #4537, opened 2022-03-30
T8874: Add the otherProjectsLink on gratispaideiawiki

ugochimobi:
Per my comment [here](https://phabricator.miraheze.org/T8874#182273), an empty or non-existent otherProjectsLinks variable suppresses the Other projects section on the sidebar.
If this works as intended then we have to tweak it in a way that it will be able to show all site IDs in the `miraheze` `siteLinkGroups ` without needing to list a thousand wikis.

### DIFF
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -160,6 +160,12 @@ if ( $wgDBname === 'gratispaideiawiki' ) {
 	$wgWBClientSettings['linkItemTags'] = [
 		'client-linkitem-change'
 	];
+	$wgWBClientSettings['otherProjectsLinks'] = [
+		'gratisdatawiki',
+		'metawiki',
+		'commonswiki',
+		'benpediawiki',
+	];
 }
 if ( $wgDBname === 'benpediawiki' ) {
 	$wgWBClientSettings['repoSiteName'] = 'Gratisdata';


Universal-Omega:
I would like an explanation here as to why you want metawiki and commonswiki, and later "all ids in the miraheze site group". As it seems unnecessary here to me. In addition metawiki and commonswiki might not even be in the Miraheze site group if they aren't Wikibase wikis. Though I could be wrong if I'm not completely understanding how this works.

Universal-Omega:
I would like an explanation here as to why you want metawiki and commonswiki, and later "all ids in the miraheze site group". As it seems unnecessary here to me. In addition metawiki and commonswiki might not even be in the Miraheze site group if they aren't Wikibase wikis. Though I could be wrong if I'm not completely understanding how this works.

ugochimobi:
I was actually replying to this, I don't know what happened, but I'm gonna reply to it asap.

ugochimobi:
So as I was saying, using metawiki et al doesn't make them connected to gratisdatawiki as a client, in fact, this is a choice of the repo wiki. And, currently, our one and only siteLinkGroup on gratisdatawiki contains every single wiki hosted Miraheze, so there's no way metawiki et al would not be included.

metawiki et al doesn't need to be a wikibase wiki for them to show up on the `miraheze` siteLinkGroup, that's our current situation. So you're almost entirely wrong. All wikis can be linked from the miraheze siteLinkGroup but because they're link on just site group doesn't make it connected as a client wiki.

Seems clear enough.

PR #4539, opened 2022-03-31
Try setting parsoid url to localhost

Universal-Omega:


RhinosF1 (dismissed):
Only host needs checking, we can assume *.mh.o is valid just wrong host

Universal-Omega:
> Only host needs checking, we can assume *.mh.o is valid just wrong host

Technically setting either is supposed to disable both, this was just a formality. Localhost as no cert so doesn't need cert check, and wrong host so no host check. I don't know if this is even valid for VirtualRestConfig though works for other http requests some places.

RhinosF1:
And I haven't checked if the option does anything either

PR #4547, opened 2022-04-06
Alphabetise Cosmos settings in LocalSettings.php

AgentIsai:


LocalSettings.php
```diff
@@ -810,7 +748,103 @@
 		'batmanwiki' => true,
 		'snapwikiwiki' => true,
 	],
-
+	'wgCosmosFooterBackgroundColor' => [
+		'default' => '#c0c0c0',
+	],
+	'wgCosmosRailBlacklistedNamespaces' => [
+		'default' => [
+			'NS_SPECIAL',
```
Universal-Omega, line 756: Should without the string, just constants for this and all below. Though we can probably add this to ManageWiki namespaces also.

Universal-Omega (changes requested):

LocalSettings.php
AgentIsai: Removed in favor of https://github.com/miraheze/mw-config/pull/4548

LocalSettings.php
```diff
@@ -810,7 +748,76 @@
 		'batmanwiki' => true,
 		'snapwikiwiki' => true,
 	],
-
+	'wgCosmosFooterBackgroundColor' => [
+		'default' => '#c0c0c0',
+	],
+	'wgCosmosRailBlacklistedNamespaces' => [
```
Universal-Omega, line 754: Defaults are not set in LS for configs in ManageWikiNamespaces.

Universal-Omega (changes requested):

LocalSettings.php
AgentIsai: Removed, thanks!

Universal-Omega (approved):

PR #4548, opened 2022-04-06
T9025: Add wgCosmosRailBlacklistedNamespaces to ManageWiki/namespaces

AgentIsai:
Per suggestion at https://github.com/miraheze/mw-config/pull/4547#discussion_r843392254 which makes more sense

### DIFF
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -127,11 +127,13 @@ $wgManageWikiNamespacesAdditional = [
 		'type' => 'check',
 		'main' => true,
 		'talk' => true,
-		'excluded' => [
-			8,
-			9,
+		'excluded' => [],
+		'overridedefault' => [
+			-1 => true,
+			8 => true,
+			9 => true,
+			'default' => true
 		],
-		'overridedefault' => false,
 		'help' => '',
 		'requires' => [],
 	],
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -121,6 +121,20 @@ $wgManageWikiNamespacesAdditional = [
 		'help' => '',
 		'requires' => [],
 	],
+	'wgCosmosRailBlacklistedNamespaces' => [
+		'name' => 'Disable Cosmos side rail in this namespace.',
+		'from' => 'cosmos',
+		'type' => 'check',
+		'main' => true,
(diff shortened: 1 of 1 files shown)


ManageWikiNamespaces.php
```diff
@@ -121,6 +121,20 @@
 		'help' => '',
 		'requires' => [],
 	],
+	'wgCosmosRailBlacklistedNamespaces' => [
+		'name' => 'Disable Cosmos side rail in this namespace.',
+		'from' => 'cosmos',
+		'type' => 'check',
+		'main' => true,
+		'talk' => true,
+		'excluded' => [
```
Universal-Omega, line 130: Should be empty

ManageWikiNamespaces.php
```diff
@@ -121,6 +121,20 @@
...
+			8,
+			9,
+		],
+		'overridedefault' => false,
```
Universal-Omega, line 134:
```suggestion
		'overridedefault' => [
			NS_MEDIAWIKI => true,
			NS_MEDIAWIKI_TALK => true,
			'default' => false,
		],
```

... I believe

Universal-Omega (changes requested):

ManageWikiNamespaces.php
AgentIsai: Done

ManageWikiNamespaces.php
AgentIsai: Done, thanks.

ManageWikiNamespaces.php
```diff
@@ -121,6 +121,22 @@
...
 	],
+	'wgCosmosRailBlacklistedNamespaces' => [
+		'name' => 'Disable Cosmos side rail in this namespace.',
+		'from' => 'cosmos',
+		'type' => 'check',
+		'main' => true,
+		'talk' => true,
+		'excluded' => [],
+		'overridedefault' => [
+			-1 => true,
```
Universal-Omega, line 132: I'm actually not certain how this will work, since NS_SPECIAL isn't available to be modified via MWN. We should consider supporting it actually, while only allowing additional settings to be set for it. We could also consider adding this particular config to ManageWikiSettings, using the `namespaces` type, but I guess this can be tested first.

PR #4562, opened 2022-04-08
set recaptcha v3 minimum score to 0.3 from 0.5 [TRIAL]

Reception123:
Since there is clearly no progress on ReCaptcha enterprise or investigating into detail, I propose we set the score to 0.3 for 12 hours and observe what the effects are - are spambots getting in? are there no more form entries? etc. I feel that most users getting turned away get 0.3 while spambots get 0.1

@RhinosF1 @MacFan4000 @Universal-Omega thoughts?

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -369,6 +369,9 @@ $wi->config->settings += [
 	'wgReCaptchaVersion' => [
 		'default' => 'v3',
 	],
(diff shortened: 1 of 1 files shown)


Universal-Omega (approved):
Just to note enterprise is one of the things I plan to resume work on soon, but for now this seems fine to me.

PR #4545, opened 2022-04-04
Add more Evelution Configuration settings

HM100:
Does not add the internal ``$wgEvelutionServelutionRelease`` or ``$wgEvelutionDisableDynamicWidgets`` configuration settings as those are used for testing/debugging only

Also does not add any associative array configuration setting as the current state of ManageWiki does not support it

PR #4574, opened 2022-04-13
Add the Configuration setting that was initially intended to be added via PR #4545

HM100:
This is a follow up of #4545


ManageWikiSettings.php
```diff
@@ -2897,6 +2897,15 @@
 		'help' => 'Always forces the content area to its maximum width.',
 		'requires' => [],
 	],
+	'wgEvelutionDisableRightRailFromSpecificPages' => [
+		'name' => 'Evelution Disable Right Rail From Specific Pages',
+		'from' => 'evelution',
+		'type' => 'wikipages',
+		'overridedefault' => false,
```
Universal-Omega, line 2904: Must be array default. So `[]`.

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #4578, opened 2022-04-13
Update Wikibase.php with propertyOrderUrl

ugochimobi:
Please check if the syntax is correct.

### DIFF
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -208,7 +208,7 @@ if ( $wgDBname === 'gratisdatawiki' ) {
 		'http://gratisdata.miraheze.org/entity/Q982' => 'triton',
 		'http://gratisdata.miraheze.org/entity/Q2123' => 'pluto',
 	];
-	$wgWBRepoSettings['propertyOrderUrl'] = $wgServer. $wgScriptPath . '/index.php?title=MediaWiki:Wikibase-SortedProperties&action=raw&sp_ver=1';
+	$wgWBRepoSettings['propertyOrderUrl'] = 'https://gratisdata.miraheze.org/wiki/MediaWiki:Wikibase-SortedProperties?action=raw&sp_ver=1';
 }
 
 if ( $wgDBname === 'gratispaideiawiki' ) {
--- a/Wikibase.php
+++ b/Wikibase.php
@@ -208,6 +208,7 @@ if ( $wgDBname === 'gratisdatawiki' ) {
 		'http://gratisdata.miraheze.org/entity/Q982' => 'triton',
 		'http://gratisdata.miraheze.org/entity/Q2123' => 'pluto',
 	];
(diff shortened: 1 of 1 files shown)


Wikibase.php
```diff
@@ -208,6 +208,7 @@
 		'http://gratisdata.miraheze.org/entity/Q982' => 'triton',
 		'http://gratisdata.miraheze.org/entity/Q2123' => 'pluto',
 	];
+	$wgWBRepoSettings['propertyOrderUrl'] = $wgServer. $wgScriptPath . '/index.php?title=MediaWiki:Wikibase-SortedProperties&action=raw&sp_ver=1';
```
Universal-Omega, line 211: Since just the one wiki, you can probably just use this:
```suggestion
	$wgWBRepoSettings['propertyOrderUrl'] = 'https://gratisdata.miraheze.org/wiki/MediaWiki:Wikibase-SortedProperties?action=raw&sp_ver=1';
```

Although, this does appear to be a WikibaseClient, not a WikibaseRepo setting.

Universal-Omega (changes requested):

Wikibase.php
ugochimobi: It is a wikibase repo settings, we're ordering wikibase properties, and we only have wikibase properties on repo not client.

ugochimobi: > > It is a wikibase repo settings, we're ordering wikibase properties, and we only have wikibase properties on repo not client.
> 
> Not according to code, according to that, this is wgWBClientSettings, and I even checked WMF config, which they use this as client.

This is not going to have any effect on the client wiki. Like, what exactly is it now going to be doing in the client wiki?

Even without any docs, I know that this is a repo setting, and according to mediawiki.org's wikibase advanced configuration page, it's is a repo setting.

Universal-Omega: > > > It is a wikibase repo settings, we're ordering wikibase properties, and we only have wikibase properties on repo not client.
> > 
> > 
> > Not according to code, according to that, this is wgWBClientSettings, and I even checked WMF config, which they use this as client.
> 
> This is not going to have any effect on the client wiki. Like, what exactly is it now going to be doing in the client wiki?
> 
> Even without any docs, I know that this is a repo setting, and according to mediawiki.org's wikibase advanced configuration page, it's is a repo setting.

I am telling you, there is no $wgWBRepoSettings['propertyOrderUrl'] in code, only $wgWBClientSettings['propertyOrderUrl']

ugochimobi:
Look, we don't have MediaWiki:Wikibase-SortedProperties in the client wiki, that is only present in the repo.

RhinosF1:
Please can those participating in technical areas be reminded of the Code of Conduct.

Universal-Omega:
This variable does not exist in code. It will not be merged like this. I am sorry.

PR #4594, opened 2022-04-16
[#T9071](https://phabricator.miraheze.org/T9071

dmehus:
Migrate user rights that should be done centrally out of an abundance of caution and prudence. Setting in //draft// merging pending approval from Owen and an EM

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2489,12 +2489,20 @@ $wgConf->settings += [
 			'sysadmin' => [
 				'globalgroupmembership' => true,
 				'globalgrouppermissions' => true,
+				'centralauth-lock' => true,
+				'centralauth-rename' => true,
+				'oathauth-api-all' => true,
+				'oathauth-disable-for-user' => true,
 			],
 			'trustandsafety' => [
 				'userrights' => true,
 				'globalgroupmembership' => true,
 				'globalgrouppermissions' => true,
 				'userrights-interwiki' => true,
+				'centralauth-lock' => true,
+				'centralauth-rename' => true,
+				'oathauth-api-all' => true,
+				'oathauth-disable-for-user' => true,
 			],
 			'sysop' => [
 				'interwiki' => true,


MacFan4000:
@dmehus a couple more rights changes were suggested.

LocalSettings.php
```diff
@@ -2489,12 +2489,18 @@
 			'sysadmin' => [
 				'globalgroupmembership' => true,
 				'globalgrouppermissions' => true,
+				'oathauth-api-all' => true,
```
RhinosF1, line 2492: Drop altogether please

LocalSettings.php
```diff
@@ -2489,12 +2489,18 @@
...
+				'oathauth-disable-for-user' => true,
 			],
 			'trustandsafety' => [
 				'userrights' => true,
 				'globalgroupmembership' => true,
 				'globalgrouppermissions' => true,
 				'userrights-interwiki' => true,
+				'centralauth-lock' => true,
+				'centralauth-rename' => true,
+				'oathauth-api-all' => true,
```
RhinosF1, line 2502: Do you actually need this?

LocalSettings.php
MacFan4000: Could also be replaced with oathauth-verify-user which allows the same thing to be done.

RhinosF1: Yep.

dmehus: I don't see any harm in leaving `oauth-api-all` local. It's a local user right for SRE and Trust and Safety online.  As for `oathauth-verify-user`, this exists in the global `sysadmin` and `trustandsafety` global groups, which I'd prefer to leave as global, should Meta Wiki be inaccessible in some way.

LocalSettings.php
dmehus: `centralauth-lock` and `centralauth-rename` would be _removed_ from the global `trustandsafety` group once this PR is merged. I suppose we could live without `oathauth-api-all` so long as `oauthauth-verify-user` is left in the global group per above.

RhinosF1: 2FA is global so you'd never need it off meta but it's logged and not the worst you could have.

You do not need the api right when you have the UI.

LocalSettings.php
RhinosF1: I am asking you as a member of SRE to remove it. I'd rather SRE only have verify-user on meta and not api-all at all.

MacFan4000:
@dmehus per the task handle-pii and globalblock should be moved as well. (Both assigned only to t&s)

LocalSettings.php
MacFan4000: Yes api is redundant when you can do the exact same via the UI.

LocalSettings.php
MacFan4000: If metawiki is not accessible then more then likely many others are inaccessible as well and the higher priority would be to restore access. For logging reasons it should never be used on other wikis anyway.

LocalSettings.php
```diff
@@ -2489,12 +2489,18 @@
...
 				'globalgrouppermissions' => true,
+				'oathauth-api-all' => true,
+				'oathauth-disable-for-user' => true,
 			],
 			'trustandsafety' => [
 				'userrights' => true,
 				'globalgroupmembership' => true,
 				'globalgrouppermissions' => true,
 				'userrights-interwiki' => true,
+				'centralauth-lock' => true,
```
MacFan4000, line 2500: If this is being moved for t&s then it should also be moved for sysadmin

LocalSettings.php
dmehus: > I am asking you as a member of SRE to remove it. I'd rather SRE only have verify-user on meta and not api-all at all. If metawiki is inaccessible and we need this urgently, we have a DB.

Oh, sorry, I thought you were replying about both groups. Yeah, I guess we don't really need `oathauth-api-all` and if you guys prefer `oathauth-verify-user` only Meta Wiki, that would be fine

RhinosF1: We would. Can you make the change here?

LocalSettings.php
dmehus: > 2FA is global so you'd never need to check it off meta but it's logged and not the worst you could have.
> 
> You do not need the api right when you have the UI.

Yeah, that's true. Plus I would just add that `trustandsafety` is a special global group as any log action or edit performed by a Trust and Safety team member is e-mailed to Owen as well.

We wouldn't be checking off of Meta Wiki, but it would just be a backup I'd feel more comfortable leaving `oathauth-verify-user` as global, as it's enquiry only user right and no suggestion it can be exploited in some way.

RhinosF1: Okay, can you still drop api-all if you're happy using the UI. -api-all is risky and not for humans anyway.

dmehus:
> @dmehus per the task handle-pii and globalblock should be moved as well. (Both assigned only to t&s)

Can GlobalBlocking be used off of Meta Wiki? Is there any suggestion it can be exploited in some way? Not saying we _would_ likely use it off of Meta Wiki, but I can see a potential use case where less visibility _may_ be preferred. I wouldn't be totally opposed, as we can always edit our global groups temporarily to re-add this user right.

`handle-pii` we would never use off of Meta Wiki anyway. My preferred preference would be maybe for @Universal-Omega to build in a configuration variable to the RemovePII extension that _only_ lets that special page be accessible from Meta Wiki; otherwise, a permission denied error is returned. But if that's too difficult, then yes, I definitely think this one should be made local with the others, mainly for transparency reasons and also potential unforeseen technical complications if performed off Meta.

LocalSettings.php
dmehus: > If this is being moved for t&s then it should also be moved for sysadmin

Does SRE still have `centralauth-lock` globally? If _so_, then my idea against moving it for `sysadmin` is because I don't think SRE has a need for this user right. As with `userrights` and `userrights-interwiki`, they have the capability to temporarily add this user right for any emergency actions they need to take. I would recommend @Reception123 remove it from the global group

RhinosF1:
> > @dmehus per the task handle-pii and globalblock should be moved as well. (Both assigned only to t&s)
> 
> Can GlobalBlocking be used off of Meta Wiki? Is there any suggestion it can be exploited in some way? Not saying we _would_ likely use it off of Meta Wiki, but I can see a potential use case where less visibility _may_ be preferred. I wouldn't be totally opposed, as we can always edit our global groups temporarily to re-add this user right.

it has an API so yes
> `handle-pii` we would never use off of Meta Wiki anyway. My preferred preference would be maybe for @Universal-Omega to build in a configuration variable to the RemovePII extension that _only_ lets that special page be accessible from Meta Wiki; otherwise, a permission denied error is returned. But if that's too difficult, then yes, I definitely think this one should be made local with the others, mainly for transparency reasons and also potential unforeseen technical complications if performed off Meta.

Why would we develop something that can you can do by revoking the permission.

LocalSettings.php
RhinosF1: I'm fine with that. The less rights we have by default is better.

dmehus:
> > `handle-pii` we would never use off of Meta Wiki anyway. My preferred preference would be maybe for @Universal-Omega to build in a configuration variable to the RemovePII extension that _only_ lets that special page be accessible from Meta Wiki; otherwise, a permission denied error is returned. But if that's too difficult, then yes, I definitely think this one should be made local with the others, mainly for transparency reasons and also potential unforeseen technical complications if performed off Meta.
> 
> Why would we develop something that can you can do by revoking the permission.

Because some extensions //only// work on a central wiki, whether that user right is assigned globally or not. I've seen it done for some things.

LocalSettings.php
dmehus: Fair. Doing now.

LocalSettings.php
dmehus: Yes, doing this now.

RhinosF1:
> > > `handle-pii` we would never use off of Meta Wiki anyway. My preferred preference would be maybe for @Universal-Omega to build in a configuration variable to the RemovePII extension that _only_ lets that special page be accessible from Meta Wiki; otherwise, a permission denied error is returned. But if that's too difficult, then yes, I definitely think this one should be made local with the others, mainly for transparency reasons and also potential unforeseen technical complications if performed off Meta.
> > 
> > 
> > Why would we develop something that can you can do by revoking the permission.
> 
> Because some extensions _only_ work on a central wiki, whether that user right is assigned globally or not. I've seen it done for some things.

I can't think of any

MacFan4000:
Still need to move `globalblock`

dmehus:
> Still need to move `globalblock`

Yeah, I was just in the process of doing that.

MacFan4000:
LGTM @Reception123 and @OwenBaines please review this.

RhinosF1 (approved):

OwenBaines (approved):

Reception123 (approved):

Reception123:
This seems fine. In case there is confusion I requested that SRE retains oauth-api-all because it's a requirement for members of SRE who have 'sysadmin' on-wiki user rights to have 2FA enabled, and I (or anyone else for that matter) should be able to easily check whether that is the case or not.

Universal-Omega (approved):

MacFan4000:
@dmehus This has been discussed now with @Reception123 and @RhinosF1, and it has been agreed that we don't want -api-all locally or at all for sysadmin.

dmehus:
> @dmehus This has been discussed now with @Reception123 and @RhinosF1, and it has been agreed that we don't want -api-all locally or at all for sysadmin as well we want the ui right locally.

Thanks. I'll make the necessary changes shortly.

MacFan4000:
> > @dmehus This has been discussed now with @Reception123 and @RhinosF1, and it has been agreed that we don't want -api-all locally or at all for sysadmin as well we want the ui right locally.
> 
> Thanks. I'll make the necessary changes shortly.

@dmehus reminding you to update this.

MacFan4000:
@dmehus why are we no longer moving globalblock to local?

Reception123:
I think that only rights that either are capable of deleting things permanently or exposing PII need to be local. Otherwise, if we are making globalblock local, we could argue that any other right that Stewards have can be abused and should be local.

RhinosF1:
So we allow people to globally block our entire user base with a bit of js

RhinosF1:
Anything that can have a serious impact and be interacted with via an api needs to go as minimum

Reception123:
What about nuke then? Couldn't that be used to nuke pages on multiple wikis?

dmehus:
I _do_ agree with @Reception123 here. There's no suggestion `globalblock`, `nuke`, `centralauth-lock`, or similar administrative user rights can be exploited in some way. They neither (a) add additional sensitive access nor do they (b) change the access level of a user's account by disabling their two-factor authentication status. Trust and Safety are trusted to perform global blocks and locks on Meta Wiki, but there may be those exceptionally rare events where less visibility is preferred, as I'm sure you can appreciated. If there's a suggestion that the API is _that_ compromised this badly, and have _evidence_ of that, that's actually a greater argument to restricting use of certain API user rights to only trusted users. However, I have not seen any evidence this is the case.

RhinosF1:
I've already said I'm more than happy to work with chrs to pentest your accounts.

I can't really at this point see a much better way of getting the point through.

RhinosF1:
And it's not the API that's the issue it's the loading arbitrary javascript for users

JohnFLewis:
Can we *stop* openly discussing security issues in public please?

Reception123:
Redone due to now urgent nature since revealed publicly. For future notice, please do not create PRs for security tasks unless they are private or it has been agreed to on the task, rather than having a public argument on the PR itself.

PR #4595, opened 2022-04-16
revert unauthorized changes made without consensus

Naleksuh:
This so-called "security change" has very little change to security and only sees to make some interface admins unable to do their job. In addition, there was not consensus for this change which absolutely must have been obtained first

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3173,12 +3173,6 @@ $wgConf->settings += [
 		'ldapwikiwiki' => 'ldapwikiwiki',
 		'betaheze' => 'testglobal',
 	],
-	'wgOATHExclusiveRights' => [
-		'metawiki' => [
-			'edituserjs',
-			'editsitejs',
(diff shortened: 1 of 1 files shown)


MacFan4000:
I don’t think we need to revert at this time. Feel free to make a new PR if there is consensus to revert.

dmehus:
While I would've _preferred_ a short discussion at https://meta.miraheze.org/wiki/Meta:Interface_administrators, as this is a minor security-related change affecting one wiki, being a sensitive wiki with global permissions, I don't see the need to procedurally revert what is a fait accompli, given that at least one (or both) Meta bureaucrats and a plurality of Stewards consented to the change.

JohnFLewis:
No clear case was originally put forward to do this change.

User affecting with no original decision made on threat or risk, rather an untoward desire to ‘enforce security’ when this change wouldn’t do that or improve security at all.

PR #4601, opened 2022-04-17
Re-enable CreateRedirect; Disable EditNotify

Universal-Omega:


RhinosF1 (changes requested):
Not without a full security review and consideration of alternatives.

dmehus:
I'd be satisfied if this has been thoroughly tested and confirmed that in no way are users able to bypass global blocks anymore. Hopefully their permissions checks have improved considerably.

PR #4612, opened 2022-04-20
override page HTML title with the wikibase label, if available, or else the id

ugochimobi:
This may or may not require an extensive review. Currently while sharing wikibase entities on social platforms, like Discord, Telegram, etc, the item shows the Qxxx instead of the item's label, so this is a possible fix.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -29,10 +29,10 @@ switch ( $wi->dbname ) {
 
 		break;
 	case 'gratisdatawiki':
-		$wgHooks['overridePageMetaTags'][] = 'overridePageMetaTags';
+		$wgHooks['overridePageMetaTags'][] = 'onoverridePageMetaTags';
 		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
 		// The title of wikibase entities will be the label, if available, or else the entity id (e.g. 'Q42').
-		function overridePageMetaTags( OutputPage $outputPage ) {
+		function onoverridePageMetaTags( OutputPage $outputPage ) {
 			$meta = $this->getOutput()->getProperty( 'wikibase-meta-tags' );
 			// This would remove description, social media tags, or any search engine optimization for diffs
 			if ( $this->isDiff() ) {
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -53,7 +53,6 @@ switch ( $wi->dbname ) {
 				
 		function onBeforePageDisplay( OutputPage $out ) {
 			$out->addMeta( 'og:image:width', '1200' );
-			$out->addMeta( 'revisit-after', '1 days' );
 		}
 		
 		break;
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -45,6 +45,7 @@ switch ( $wi->dbname ) {
 				return;
 			}
 			$out->addMeta( 'og:image', wfExpandUrl( $thumb->getUrl(), PROTO_CANONICAL ) );
+			$out->addMeta( 'revisit-after', '1 days' );
 		}
 		
 		break;
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -29,10 +29,10 @@ switch ( $wi->dbname ) {
 
 		break;
 	case 'gratisdatawiki':
-		$wgHooks['overridePageMetaTags'][] = 'onoverridePageMetaTags';
+		$wgHooks['OverridePageMetaTags'][] = 'onOverridePageMetaTags';
 		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
 		// The title of wikibase entities will be the label, if available, or else the entity id (e.g. 'Q42').
-		function onoverridePageMetaTags( OutputPage $outputPage ) {
+		function onOverridePageMetaTags( OutputPage $outputPage ) {
 			$meta = $this->getOutput()->getProperty( 'wikibase-meta-tags' );
 			// This would remove description, social media tags, or any search engine optimization for diffs
 			if ( $this->isDiff() ) {
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -28,6 +28,15 @@ switch ( $wi->dbname ) {
 		}
 
 		break;
+	case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $out, Skin $skin ) {
+			$meta = $out->getMetaTags();
+			
+			if ( !isset( $meta['og:type'] ) && !isset( $meta['twitter:card'] ) ) {
+		
+		break;
 	case 'ldapwikiwiki':
 		wfLoadExtension( 'LdapAuthentication' );
 
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -30,6 +30,7 @@ switch ( $wi->dbname ) {
 		break;
 	case 'gratisdatawiki':
 		$wgHooks['OutputPageParserOutput'][] = 'onOutputPageParserOutput';
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
 		
 		function onOutputPageParserOutput( OutputPage $out, ParserOutput $parserOutput ) {
 			$placeholders = $parserOutput->getExtensionData( 'wikibase-view-chunks' );
@@ -48,11 +49,8 @@ switch ( $wi->dbname ) {
 					$out->addLink( $link );
 				}
 			}
-			
-			return true;
 		}
-		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
-			
+				
 		function onBeforePageDisplay( OutputPage $out ) {
 			$out->addMeta( 'og:image:width', '1200' );
 			$out->addMeta( 'og:image', wfExpandUrl( $thumb->getUrl(), PROTO_CANONICAL ) );
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -35,6 +35,7 @@ switch ( $wi->dbname ) {
 			$meta = $out->getMetaTags();
 			
 			if ( !isset( $meta['og:type'] ) && !isset( $meta['twitter:card'] ) ) {
+				$out->addMeta( 'og:type', 'summary' );
 		
 		break;
 	case 'ldapwikiwiki':
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -29,21 +29,32 @@ switch ( $wi->dbname ) {
 
 		break;
 	case 'gratisdatawiki':
-		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		$wgHooks['OutputPageParserOutput'][] = 'onOutputPageParserOutput';
 		
-		function onBeforePageDisplay( OutputPage $out, Skin $skin ) {
-			$meta = $out->getMetaTags();
+		function onOutputPageParserOutput( OutputPage $out, ParserOutput $parserOutput ) {
+			$placeholders = $parserOutput->getExtensionData( 'wikibase-view-chunks' );
+			if ( $placeholders !== null ) {
+				$out->setProperty( 'wikibase-view-chunks', $placeholders );
+			}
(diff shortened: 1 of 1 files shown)


Universal-Omega:
This is a very advanced hook for ParserOutput usage usage. It could possibly cause warnings here, trying to change extension data from here isn't the best idea. I'm not sure we want to do it. However I'm willing to deploy this, and if it works great!, if it has no effect, or negative, revert, I'm willing to since it only effects one wiki. But I do want to extensively review things here first before I'm comfortable to.

LocalWiki.php
```diff
@@ -27,6 +27,38 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+			$meta = $parserOutput->getExtensionData( 'wikibase-meta-tags' );
+			$out->setProperty( 'wikibase-meta-tags', $meta );
+			$alternateLinks = $parserOutput->getExtensionData( 'wikibase-alternate-links' );
+			if ( $alternateLinks !== null ) {
+				foreach ( $alternateLinks as $link ) {
+					$out->addLink( $link );
+				}
+			}
+			
+			return true;
```
Universal-Omega, line 52: No return value necessary.

LocalWiki.php
```diff
@@ -27,6 +27,38 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+		}
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
```
Universal-Omega, line 54: Move to the top, so `$wgHooks` is kept together for nicer format.

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: Done

LocalWiki.php
ugochimobi: Done

ugochimobi:
> On second thought, maybe we should not deploy this. Most if not all the Parser stuff you used here is deprecated in 1.38 also I believe, and 1.39 switches to Parsoid, so would likely break then also.

I think maybe it should be deployed though, but you're the sysadmin here (one who has extensive knowledge than me) so you should know what's best at this point, but like you said, since it will affect just one wiki, we might be interested in seeing if it works at all.

LocalWiki.php
```diff
@@ -27,6 +27,36 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+			if ( $alternateLinks !== null ) {
+				foreach ( $alternateLinks as $link ) {
+					$out->addLink( $link );
+				}
+			}
+		}
+				
+		function onBeforePageDisplay( OutputPage $out ) {
+			$out->addMeta( 'og:image:width', '1200' );
+			$out->addMeta( 'og:image', wfExpandUrl( $thumb->getUrl(), PROTO_CANONICAL ) );
```
Universal-Omega, line 56: This will definitely cause exceptions, since `$thumb` doesn't exist. But this is the default added by PageImages when thumb does exist, so not necessary, since you have that enabled.

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: > This will definitely cause exceptions, since `$thumb` doesn't exist. But this is the default added by PageImages when thumb does exist, so not necessary, since you have that enabled.

that's true

LocalWiki.php
```diff
@@ -27,6 +27,35 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+			if ( $alternateLinks !== null ) {
+				foreach ( $alternateLinks as $link ) {
+					$out->addLink( $link );
+				}
+			}
+		}
+				
+		function onBeforePageDisplay( OutputPage $out ) {
+			$out->addMeta( 'og:image:width', '1200' );
+			$out->addMeta( 'revisit-after', '1 days' );
```
RhinosF1, line 56: This isn't used by any major browser anymore.

ugochimobi: > This isn't used by any major browser anymore.

It should go out then, I got the inspiration from metawiki's own, but since it's going out, then why not

RhinosF1: I have created #4613 for cleaning meta up

ugochimobi: Yeah, I saw

ugochimobi:
I realize that most of these parsers have been removed, some will soon be... Perhaps, I'd just update them with the new ones.

LocalWiki.php
```diff
@@ -27,6 +27,39 @@ function onBeforePageDisplay( OutputPage $out ) {
 			$out->addMeta( 'og:image:width', '1200' );
 		}
 
+		break;
+	case 'gratisdatawiki':
+		$wgHooks['overridePageMetaTags'][] = 'overridePageMetaTags';
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		// The title of wikibase entities will be the label, if available, or else the entity id (e.g. 'Q42').
+		function overridePageMetaTags( OutputPage $outputPage ) {
```
RhinosF1, line 35: Please call onX as standard for hooks

ugochimobi: > Please call onX as standard for hooks

I was almost asking a question about that, I thought it doesn't matter...

ugochimobi: @RhinosF1 like this?

RhinosF1: each word excluding the first should be a capital so onOverridePageMetaTags

ugochimobi: > each word excluding the first should be a capital so onOverridePageMetaTags

Okay 👍

LocalWiki.php
```diff
@@ -29,10 +29,10 @@ function onBeforePageDisplay( OutputPage $out ) {
 
 		break;
 	case 'gratisdatawiki':
-		$wgHooks['overridePageMetaTags'][] = 'onoverridePageMetaTags';
+		$wgHooks['OverridePageMetaTags'][] = 'onOverridePageMetaTags';
```
RhinosF1, line 32: No, that's not the first letter being lowercase

LocalWiki.php
ugochimobi: @RhinosF1 we good with this?

RhinosF1: Nope, see comment below

LocalWiki.php
ugochimobi: > No, that's not the first letter being lowercase

Oh, like the first one `overridePageMetaTags` then the remaining `OverridePageMetaTags`?

ugochimobi: > Yes

Okay, Done!

LocalWiki.php
```diff
@@ -27,6 +27,39 @@ function onBeforePageDisplay( OutputPage $out ) {
 			$out->addMeta( 'og:image:width', '1200' );
 		}
 
+		break;
+	case 'gratisdatawiki':
+		$wgHooks['overridePageMetaTags'][] = 'onOverridePageMetaTags';
```
Universal-Omega, line 32: Where does this hook come from? I can't find any hook with this name anywhere.

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: > Where does this hook come from? I can't find any hook with this name anywhere.

Let me link you https://github.com/wikimedia/Wikibase/blob/master/repo/includes/Actions/ViewEntityAction.php#L140

Universal-Omega: > 

There are no replacements for some stuff.

Universal-Omega: Um... that is not a hook.

ugochimobi: > Um... that is not a hook.

Yeah, that's a function, but I thought after seeing a function, knowing the hook shouldn't be hard.... (dumb me)

ugochimobi:
> > I realize that most of these parsers have been removed, some will soon be... Perhaps, I'd just update them with the new ones.
> 
> There are no replacements for some stuff.

Yeah, this is a new new to do the requested, I mean, the most recent way.

ugochimobi:
@Universal-Omega This is ready for merge, I'd love to see the production and know if there's any issue.

LocalWiki.php
```diff
@@ -32,6 +32,37 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+  case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$out->addMeta( 'og:image:width', '1200' );
+			
+			// The title of wikibase entities will be the label, if available, or else the entity id (e.g. 'Q42').
+			$meta = $this->getOutput()->getProperty( 'wikibase-meta-tags' );
+			// This would remove description, social media tags, or any search engine optimization for diffs
+			if ( $this->isDiff() ) {
```
Universal-Omega, line 45: Doesn't exist.

LocalWiki.php
```diff
@@ -32,6 +32,37 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				if ( isset( $meta['title'] ) ) {
+					$this->setDiffPageTitle( $outputPage, $meta['title'] );
```
Universal-Omega, line 47: Doesn't exist.

LocalWiki.php
```diff
@@ -32,6 +32,37 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				}
+			}
+			
+			if ( isset( $meta['title'] ) ) {
+				$this->setHTMLTitle( $outputPage, $meta['title'] );
```
Universal-Omega, line 52: Doesn't exist.

LocalWiki.php
```diff
@@ -32,6 +32,37 @@ function onBeforePageDisplay( OutputPage $out ) {
...
 		
+		break;
+  case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$out->addMeta( 'og:image:width', '1200' );
+			
+			// The title of wikibase entities will be the label, if available, or else the entity id (e.g. 'Q42').
+			$meta = $this->getOutput()->getProperty( 'wikibase-meta-tags' );
```
Universal-Omega, line 43: Doesn't exist.

LocalWiki.php
```diff
@@ -32,6 +32,37 @@ function onBeforePageDisplay( OutputPage $out ) {
 		$wgJsonConfigs['Map.JsonConfig']['store'] = true;
 		$wgJsonConfigs['Tabular.JsonConfig']['store'] = true;
 		
+		break;
+  case 'gratisdatawiki':
```
Universal-Omega, line 36: Tabbing.

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: > Doesn't exist.

@Universal-Omega See this https://github.com/wikimedia/Wikibase/blob/master/repo/includes/Actions/ViewEntityAction.php#L140

Universal-Omega: Yeah but not in this context.

Universal-Omega: LocalWiki.php is not $this in ViewEntityAction.php.

ugochimobi: > LocalWiki.php is not $this in ViewEntityAction.php.

Oh, what do we use then?

Universal-Omega:
Closing since this definitely won't work like this, and I don't think we even can do this the way you want.

ugochimobi:
The possible replacement I've seen so far is `$wgOut`

Universal-Omega:
There is no replacement for some. $this is only available in the initial file, in a local context, it isn't available anywhere else. So no this most definitely won't work, with 100% certainty.

ugochimobi:
and, if you go to this page https://gratisdata.miraheze.org/wiki/Q101 and Inspect it on your browser, you'd find out that there are two (2) `og:title` tags one of which has the right title which the other has the Qid name, is there a way to merge the both, possibly?

ugochimobi:
> There is no replacement for some. $this is only available in the initial file, in a local context, it isn't available anywhere else. So no this most definitely won't work, with 100% certainty.

Hmm

ugochimobi:
There are also two (2) `og:description`` tags, one of which has the correct value while the other has just a period (.) 
All these need to be fixed actually.

PR #4613, opened 2022-04-20
metawiki: drop long unused metatags

RhinosF1:
These have no benefit to SEO and haven't been used to much effect by any major browser in years


AgentIsai:
I’d actually like to see the description field also removed as it clashes with WikiSEO which doesn’t allow it to do it’s job properly. It also causes SEO penalties for reusing the same description across the entire wiki and is also sometimes even ignored by search engines as it’s repetitive. Pinging @Reception123, a Meta bureaucrat, for his thoughts.

PR #4622, opened 2022-04-22
set wgLexemeLanguageCodePropertyId and wgLexemeEnableDataTransclusion for gratisdatawiki

ugochimobi:

PR #4623, opened 2022-04-22
set wgPropertySuggesterMinProbability for gratisdatawiki

ugochimobi:
this sets how fast the suggestions should show up.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3534,6 +3534,10 @@ $wgConf->settings += [
 	'wgPropertySuggesterInitialSuggestions' => [
 		'default' => [],
 	],
+	'wgPropertySuggesterMinProbability' => [
+		'default' => 0.05,
+		'gratisdatawiki' => 0.069,
+	],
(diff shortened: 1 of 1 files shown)


Universal-Omega (changes requested):
Should be added to ManageWiki instead.

ugochimobi:
> Should be added to ManageWiki instead.

It's currently not in ManageWiki, yoild you add it now?

Universal-Omega:
> > Should be added to ManageWiki instead.
> 
> It's currently not in ManageWiki, yoild you add it now?

Yeah, what I'm saying is we could add it.

ugochimobi:
> > > Should be added to ManageWiki instead.
> > 
> > 
> > It's currently not in ManageWiki, yoild you add it now?
> 
> Yeah, what I'm saying is we could add it.

Yeh I got you from the reply at #4622 💯

Universal-Omega:
* #4665

PR #4624, opened 2022-04-22
add UploadWizard Flickr configs for gpcommonswiki

ugochimobi:
I doubt if this is the correct syntax though

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4269,6 +4269,10 @@ $wgConf->settings += [
 		'wmgUseUploadWizard' => [
 			'campaignExpensiveStatsEnabled' => false,
 		],
+		'+gpcommonswiki' => [
+			'flickrApiUrl' => 'https://api.flickr.com/services/rest/?',
+			'flickrApiKey' => 'c436a013c03d8a823a4333ace4771ccd',
+		],
 	],
 
 	// Uploads


LocalSettings.php
```diff
@@ -4269,6 +4269,10 @@
 		'wmgUseUploadWizard' => [
 			'campaignExpensiveStatsEnabled' => false,
 		],
+		'+gpcommonswiki' => [
```
ugochimobi, line 4272: I'd so much be inclined to setting this UploadWizard on a per-wiki basis on LocalWiki.php though

Universal-Omega: This should work how you have it here.

ugochimobi: > This should work how you have it here.

Ah, okay, I thought it wasn't going to though.

Universal-Omega: It is though we should probably add this to ManageWiki, or have someone at Miraheze generate a simple Flickr key to use globally.

ugochimobi: > It is though we should probably add this to ManageWiki, or have someone at Miraheze generate a simple Flickr key to use globally.

Yeah, it would be necessary at this point to have the configurations added to ManageWiki, as for the Flickr key, a global one is indeed a good idea, but anyone could use theirs if they wish.

LocalSettings.php
```diff
@@ -4269,6 +4269,10 @@
...
+			'flickrApiUrl' => 'https://api.flickr.com/services/rest/?',
```
Universal-Omega, line 4273: What's the reasoning for this? I think this might be the default but not 100% sure.

ugochimobi: > What's the reasoning for this? I think this might be the default but not 100% sure.

Tbh, I couldn't even find a single default for the extension on its .json.... So I can't really tell, but I tried using the Flickr stuff just now on UploadWizard and it responded fine, although I didn't try uploading using it
....

Universal-Omega: The defaults are set a different way for this extension.

ugochimobi: > The defaults are set a different way for this extension.

Yeah,ni observed that.

Universal-Omega: flickrApiUrl can be removed from here I think.

Universal-Omega (changes requested):

Universal-Omega:
#4649 will add this to ManageWiki.

PR #4626, opened 2022-04-23
set JsonConfig configuration for gpcommonswiki

ugochimobi:


LocalSettings.php
```diff
@@ -1949,16 +1949,39 @@
...
 	],
 	'wgJsonConfigModels' => [
 		'default' => [
 			'Map.JsonConfig' => JsonConfig\JCMapDataContent::class,
 			'Tabular.JsonConfig' => JsonConfig\JCTabularContent::class,
 		],
+		'gpcommonswiki' => [
+			'Tabular.JsonConfig' => 'JsonConfig\JCTabularContent',
+			'Map.JsonConfig' => 'JsonConfig\JCMapDataContent',
+		],
```
Universal-Omega, line 1984: This is the exact same as the above default.

Universal-Omega (changes requested):

LocalSettings.php
ugochimobi: > This is the exact same as the above default.

Oh, I'd remove that then, I thought there was a difference between both, like if they function differently.

Universal-Omega: Nope, its the exact same as the default.

ugochimobi: > Nope, its the exact same as the default.

Okay, I'd just remove it then.

Universal-Omega: Please do then.

Universal-Omega (approved):

PR #4631, opened 2022-04-24
Add essential special pages to `*`

dmehus:
Per [SN request](https://meta.miraheze.org/w/index.php?title=Stewards%27_noticeboard&diff=next&oldid=246492&diffmode=source#I_have_a_question). This has created the scenario where these rights where these rights were blacklisted from being removed from `*`, but on wikis that already removed them, they cannot be re-added, by local bureaucrats or Stewards. This was an unfortunate oversight as these _should've_ been made default `*` rights, locked down from being removed in ManageWiki

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2305,9 +2305,9 @@ $wgConf->settings += [
 				'oathauth-enable' => true,
 				'editmyprivateinfo' => true,
 				'viewmyprivateinfo' => true,
-				'editmyoptions',
-				'editmyprivateinfo',
-				'editmywatchlist',
+				'editmyoptions' => true,
+				'editmyprivateinfo' => true,
+				'editmywatchlist' => true,
 				'writeapi' => true,
 			],
 			'checkuser' => [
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2305,6 +2305,9 @@
 				'oathauth-enable' => true,
 				'editmyprivateinfo' => true,
 				'viewmyprivateinfo' => true,
+				'editmyoptions',
```
Universal-Omega, line 2308: This and below: => true needed.

Universal-Omega (changes requested):

LocalSettings.php
dmehus: Oh, my bad! Totally missed that.

Universal-Omega (approved):

LocalSettings.php
```diff
@@ -2305,6 +2305,9 @@
 				'oathauth-enable' => true,
 				'editmyprivateinfo' => true,
 				'viewmyprivateinfo' => true,
+				'editmyoptions' => true,
+				'editmyprivateinfo' => true,
```
Universal-Omega, line 2309: This is already present above.

Universal-Omega (changes requested):

LocalSettings.php
dmehus: Done

Universal-Omega (approved):

PR #4632, opened 2022-04-24
GlobalCache: don't use wg for redis ip

RhinosF1:
This is only used to vary config between beta/prod. It shouldn't be used outside of this file.

### DIFF
--- a/GlobalCache.php
+++ b/GlobalCache.php
@@ -5,7 +5,7 @@ $wgMemCachedServers = [
 	'127.0.0.1:11213',
 ];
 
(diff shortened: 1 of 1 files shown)


dmehus:
Why not create a custom variable for Betaheze, such as `wmgRedisServerIP`, or how come you can't specify a custom configuration for `betawiki` for this variable?

RhinosF1:
> Why not create a custom variable for Betaheze, such as `wmgRedisServerIP`, or how come you can't specify a custom configuration for `betawiki` for this variable?

Because it stops people accessing a variable that isn't the best way.

PR #4633, opened 2022-04-24
Set spamblacklist to true on all wikis

dmehus:
Set `default` to `true`, as this would be useless to us otherwise. Individual wikis would still be able to control who has access to the spam blacklist log by setting `spamblacklist` to a group other than `user`

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -191,14 +191,9 @@ $wgConf->settings += [
 	],
 	'wgLogSpamBlacklistHits' => [
 		'default' => true,
-		'metawiki' => true,
-		'betaheze' => true,
 	],
 	'wgTitleBlacklistLogHits' => [
 		'default' => true,
-		'loginwiki' => true,
-		'metawiki' => true,
-		'betaheze' => true,
 	],
 
 	// ApprovedRevs
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -190,12 +190,12 @@ $wgConf->settings += [
 		],
 	],
 	'wgLogSpamBlacklistHits' => [
-		'default' => false,
+		'default' => true,
 		'metawiki' => true,
 		'betaheze' => true,
 	],
 	'wgTitleBlacklistLogHits' => [
-		'default' => false,
+		'default' => true,
 		'loginwiki' => true,
 		'metawiki' => true,
 		'betaheze' => true,


RhinosF1 (changes requested):
This right is only applied to stewards via global group not user as stated. Can you clarify intent?

dmehus:
> This right is only applied to stewards via global group not user as stated. Can you clarify intent?

`spamblacklist`, as far as I am aware, is not restricted to Stewards. If it is, it shouldn't be. It's no different than us granting local wikis the ability to grant the title blacklist override rights to local bureaucrats. I _can_ see restricting it to `user` or even `sysop`

RhinosF1:
The right to access logs is @dmehus as far as I can check. I'm not sure whether it's actually restricted or just never set. I didn't look far.

dmehus:
> The right to access logs is @dmehus as far as I can check. I'm not sure whether it's actually restricted or just never set. I didn't look far.

Actually, it's not restricted. Originally there was a request from C.Syde65 to add the user right on `testwiki`, and I remember seeing it was not restricted. Indeed, confirming with https://csydes.miraheze.org/wiki/Special:ListGroupRights#user, it's not restricted, but as I say, it's useless since we're not logging hits :)

RhinosF1:
Cool, that means it just needs adding to default permissions then. I only checked whether it was listed on meta and a recently created one.

dmehus:
> Cool, that means it just needs adding to default permissions then. I only checked whether it was listed on meta and a recently created one.

I have no objections to adding that to the default permissions in ManageWikiDefaultPermissions, but this would not change existing wikis' configurations, only to _newly created_ wikis, and I'm not sure it's needed. I'd rather discuss this with other Stewards to see what group should have it by default.

As to setting this variable to `true`, I did check with John, who had no objections, and Agent, who was surprised it was not already set to `true`.

RhinosF1:
We could consider a community discussion and adding to existing if needed globally. Please discuss away :)

I have no objections as long as the situation is clear. I'd consider whether we want spammers to be able to see the log but they can see the lists anyway so it's not really a risk.

RhinosF1 (approved):
No issues once @dmehus has discussed with Stewards & Community who gets the rights and what to do about existing wikis. Also not really fussed if that doesn't happen on merge as long as commitment to do it.

dmehus:
> We could consider a community discussion and adding to existing if needed globally. Please discuss away :)
> 
> I have no objections as long as the situation is clear. I'd consider whether we want spammers to be able to see the log but they can see the lists anyway so it's not really a risk.

I'm not proposing to add this user right to the default permissions, and it's not clear this would even need a discussion anyway.

To your latter point, which is what this PR proposes, as you said, the local and global spam blacklists (and whitelists) are public, so any potential concern of spambots seeing the spam blacklist logs are mooted/non-existent by this fact. Additionally, all the spam blacklist log will tell anyone is the actor _who performed_ the action. The actual blocked external link is already public via the spam blacklist.

dmehus:
> All checks have passed

As stated, I'm not proposing to add this user right to the ManageWikiDefaultPermissions. That can be done separately, via a separate PR, if we feel it's needed. This is just to make the spam blacklist not useless. :)

JohnFLewis:
> > All checks have passed
> 
> As stated, I'm not proposing to add this user right to the ManageWikiDefaultPermissions. That can be done separately, via a separate PR, if we feel it's needed. This is just to make the spam blacklist not useless. :)

This is done on wiki by Stewards not SRE anyway

RhinosF1:
I'm proposing adding it to default permissions but I'm saying I want it to be discussed. I'm not bothered about blocking this PR on it.

RhinosF1:
> > > All checks have passed
> > 
> > 
> > As stated, I'm not proposing to add this user right to the ManageWikiDefaultPermissions. That can be done separately, via a separate PR, if we feel it's needed. This is just to make the spam blacklist not useless. :)
> 
> This is done on wiki by Stewards not SRE anyway

Yes but we can discuss work needed by stewards to get the best out of this PR.

dmehus:
> I'm proposing adding it to default permissions but I'm saying I want it to be discussed. I'm not bothered about blocking this PR on it.

Okay, great. And that can surely be discussed, but I also don't see the urgency in changing the default permissions. Wikis can freely add the spam blacklist log user right to their wikis as they see fit

PR #4636, opened 2022-04-26
fixing Tabular.JsonConfig and Map.JsonConfig for gpcommonswiki

ugochimobi:


ugochimobi:
This PR should't be merged until #4637 is merged, if that doesn't brings up the Data: namespace, then maybe this would.

ugochimobi:
Okay, luckily, #4637 brought the Data: namespace, so this is of course not going to be done

PR #4644, opened 2022-04-27
set wgPageLanguageUseDB to true for gpcommonswiki and gratisdatawiki

ugochimobi:


Universal-Omega:
This is in ManageWiki.

ugochimobi:
> This is in ManageWiki.

I couldn't find it

PR #4656, opened 2022-04-29
remove disabledDataTypes from gratispaideia and co

ugochimobi:
This config currently throws exceptions on some pages eg, special pages, and some other


ugochimobi:
or @Universal-Omega you could look the Graylog if you want, because they're not entirely down, just that accessing some pages throws errors, particularly special pages.

PR #4669, opened 2022-05-06
set wgInterwikiSortingSort to alphabetically on gratispaideiawiki

ugochimobi:


LocalSettings.php
```diff
@@ -1793,6 +1793,12 @@
 		],
 	],
 
+	//InterwikiSorting
```
Universal-Omega, line 1796:
```suggestion
	// InterwikiSorting
```

Universal-Omega (changes requested):

Universal-Omega:
This could probably be added to ManageWiki.

LocalSettings.php
```diff
@@ -1793,6 +1793,12 @@
 		],
 	],
 
+	// InterwikiSorting
+	'wgInterwikiSortingSort' => [
+		'default' => 'code',
+		'gratispaideiawiki' => 'alphabetic',
```
Universal-Omega, line 1799: We don't support this yet actually. This is not built in to the extension and only supported by the WMF.

Universal-Omega (changes requested):

LocalSettings.php
ugochimobi: > We don't support this yet actually. This is not built in to the extension and only supported by the WMF.

Did you add it to ManageWiki?

ugochimobi:
Per above.

PR #4671, opened 2022-05-07
Licenses: Add GFDL

RhinosF1:


Universal-Omega (approved):
Since already in $wmgWikiLicense in defines.php I guess this is OK?

PR #4677, opened 2022-05-10
Enable TextExtracts globally + SEO fixes

AgentIsai:
TextExtracts has been bundled with MediaWiki since MW 1.34 and provides an API module which is used by many popular extensions such as Popups and WikiSEO and allows for a very vital part of WikISEO to function, it's auto-generated descriptions. This change proposes TextExtracts be enabled globally as a default extension and that 2 non-user impacting configuration variables be set in order to help with SEO. 

These settings should also preferably be toggled globally if this PR is merged to enable this feature globally.


Universal-Omega:
Is there community consensus for this?

AgentIsai:
Before making this PR, I evaluated if a community discussion was required for this and ultimately I believe it is not for the following reasons: 

1) TextExtracts has been bundled with MediaWiki since MediaWiki 1.34. In keeping with our promise of providing a near-stock experience of MediaWiki, it is assumed that we also include all bundled extensions by default on all wikis which is mostly true as we do enable by default most bundled extensions except a handful which were bundled with MediaWiki just recently (CodeEditor, MultimediaViewer, PageImages, VisualEditor and this extension). 

2) These changes will not impact user experience. Unlike was the case for the RfF for Purge/WIkiSEO, this change will not be visible at all by the average user as TextExtracts simply adds a new API module and the two settings tweaked here merely extend and improve the functionality of WikiSEO using the newly enabled extension. The two previously mentioned extensions differ from this as Purge added a new menu to skins to allow for purging page caches and WikiSEO enabled a whole new system of changing meta tags in order to boost a wiki in it's search engine visibility while TextExtracts is just a simple API module and the two toggled settings take advantage of the extension and generate clean descriptions instead of having Google generate a description for them or penalize them for not having a description set.

3) The community has endorsed in the past tweaks to help Miraheze's SEO. As noted in the RfF for the addition of WikiSEO, users supported the addition of the extension as historically we have had very poor search engine visibility and users have long complained on Discord that their wikis have issues with search engines and that we rank poorly. The addition of WikiSEO has certainly helped us but I believe changing setting globally would benefit SEO a lot. The toggling of these two extensions would *not* have any impact on users and users who wish to manually set a description perfectly can or they can even disable WikiSEO as they please as it is a default (but not global) extension.

RhinosF1:
Note: What's the plan with existing wikis?

Universal-Omega (approved):

LocalSettings.php
```diff
@@ -4876,10 +4877,10 @@
 		'default' => false,
 	],
 	'wgWikiSeoEnableAutoDescription' => [
-		'default' => false,
+		'default' => true,
```
Universal-Omega, line 4880: overridedefault needs changed in ManageWikiSettings.php.

LocalSettings.php
```diff
@@ -4876,10 +4877,10 @@
...
 	],
 	'wgWikiSeoTryCleanAutoDescription' => [
-		'default' => false,
+		'default' => true,
```
Universal-Omega, line 4883: overridedefault needs changed in ManageWikiSettings.php.

Universal-Omega (changes requested):

LocalSettings.php
AgentIsai: Should this be changed back to 'false' then?

Universal-Omega: No, it should match ManageWikiSettings.php

AgentIsai: Changed then, thanks.

LocalSettings.php
AgentIsai: Changed, thanks.

Universal-Omega (approved):

PR #4689, opened 2022-05-11
Set meta tags on entity pages gratisdatawiki

ugochimobi:
For now, let's set this one.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -49,6 +49,7 @@ switch ( $wi->dbname ) {
 		function onBeforePageDisplay( OutputPage $outputPage ) {
 			$outputPage->addMeta( 'og:image:width', '1200' );
 			
+			$meta = $outputPage->getProperty( 'wikibase-meta-tags' );
 			if ( isset( $meta['title'] ) ) {
 				$outputPage->addMeta( 'og:title', $meta['title'] );
 			}
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -67,7 +67,7 @@ switch ( $wi->dbname ) {
 		];
 		$wgJsonConfigs['Map.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
-		];	
+		];
 
 		break;
 	case 'gratispaideiawiki':
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -44,12 +44,32 @@ switch ( $wi->dbname ) {
 
 		break;
 	case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$out->addMeta( 'og:image:width', '1200' );
+			
+			if ( isset( $meta['title'] ) ) {
+				$outputPage->addMeta( 'og:title', $meta['title'] );
+			}
+			
+			if ( isset( $meta['description'] ) ) {
+				$outputPage->addMeta( 'description', $meta['description'] );
+				$outputPage->addMeta( 'og:description', $meta['description'] );
+				
+				if ( isset( $meta['title'] ) ) {
+					$outputPage->addMeta( 'og:type', 'summary' );
+				}
+			}
+		}
+		
 		$wgJsonConfigs['Tabular.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
 		];
 		$wgJsonConfigs['Map.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
 		];
+		
 
 		break;
 	case 'gratispaideiawiki':
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -61,15 +61,13 @@ switch ( $wi->dbname ) {
 					$outputPage->addMeta( 'og:type', 'summary' );
 				}
 			}
-		}
-		
+		}	
 		$wgJsonConfigs['Tabular.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
 		];
 		$wgJsonConfigs['Map.JsonConfig']['remote'] = [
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -44,12 +44,30 @@ function onBeforePageDisplay( OutputPage $out ) {
 
 		break;
 	case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$out->addMeta( 'og:image:width', '1200' );
```
Universal-Omega, line 50:
```suggestion
			$outputPage->addMeta( 'og:image:width', '1200' );
```

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: I was going to ask you that though.

ugochimobi: I was going to ask you that though.

LocalWiki.php
```diff
@@ -44,12 +44,30 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				}
+			}
+		}	
 		$wgJsonConfigs['Tabular.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
 		];
 		$wgJsonConfigs['Map.JsonConfig']['remote'] = [
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
-		];
+		];
```
Universal-Omega, line 70: extra tabbing needs removed.

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: > extra tabbing needs removed.

Where exactly is the extra tabbing?

Universal-Omega: The only change on this line.

ugochimobi: > The only change on this line.

Done

LocalWiki.php
```diff
@@ -44,6 +44,24 @@ function onBeforePageDisplay( OutputPage $out ) {
 
 		break;
 	case 'gratisdatawiki':
+		$wgHooks['BeforePageDisplay'][] = 'onBeforePageDisplay';
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$outputPage->addMeta( 'og:image:width', '1200' );
+			
+			if ( isset( $meta['title'] ) ) {
```
Universal-Omega, line 52: this won't do anything, as $meta is never set anywhere within this hook, so the `if` always returns false, and nothing is executed/added here.

LocalWiki.php
```diff
@@ -44,6 +44,24 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				$outputPage->addMeta( 'og:title', $meta['title'] );
+			}
+			
+			if ( isset( $meta['description'] ) ) {
```
Universal-Omega, line 56: this won't do anything, as $meta is never set anywhere within this hook, so the `if` always returns false, and nothing is executed/added here.

LocalWiki.php
```diff
@@ -44,6 +44,24 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				$outputPage->addMeta( 'description', $meta['description'] );
+				$outputPage->addMeta( 'og:description', $meta['description'] );
+				
+				if ( isset( $meta['title'] ) ) {
```
Universal-Omega, line 60: this won't do anything, as $meta is never set anywhere within this hook, so the `if` always returns false, and nothing is executed/added here.

LocalWiki.php
```diff
@@ -44,6 +44,24 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+		
+		function onBeforePageDisplay( OutputPage $outputPage ) {
+			$outputPage->addMeta( 'og:image:width', '1200' );
+			
+			if ( isset( $meta['title'] ) ) {
+				$outputPage->addMeta( 'og:title', $meta['title'] );
+			}
+			
+			if ( isset( $meta['description'] ) ) {
+				$outputPage->addMeta( 'description', $meta['description'] );
```
Universal-Omega, line 57: $meta is not set

LocalWiki.php
```diff
@@ -44,6 +44,24 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+				$outputPage->addMeta( 'og:description', $meta['description'] );
```
Universal-Omega, line 58: $meta is not set

Universal-Omega (changes requested):

LocalWiki.php
ugochimobi: This is where `$meta = $this->getOutput()->getProperty( 'wikibase-meta-tags' );` comes in, but I don't know how we'd do `$this` for LocalWiki otherwise we'd have to do a direct calling, instead of the IFs.

Universal-Omega: this should work in this particular instance:
```php
$meta = $outputPage->getProperty( 'wikibase-meta-tags' );
```

ugochimobi: hmm, okay then, let me set $meta with that.

ugochimobi: Done

Universal-Omega (approved):

LocalWiki.php
ugochimobi: This actually took effect, but didn't help fixing the main og:title issue

Universal-Omega:
> This actually took effect, but didn't help fixing the main og:title issue

What exactly are you trying to do?

ugochimobi:
> > This actually took effect, but didn't help fixing the main og:title issue
> 
> What exactly are you trying to do?

For example, take the Q101 item I linked above, if you share the link to maybe Facebook or Twitter the description and title (in this case it is supposed to be the item label) doesn't appear. Although sharing the link on discord shows the description bit doesn't show the title (label).

So maybe because Twitter and Facebook requires the open graph tags and discord doesn't exclusively requires it.

When you inspect the page on your browser, you'd find two title and description tags also.

PR #4699, opened 2022-05-16
Add search-related Citizen config to ManageWikiSettings

alistair3149:
* `$wgCitizenEnableSearch` - Enable or disable Citizen search module. Useful for unbreaking and also wikis with custom search suggestions (e.g. Wikidata)
* `$wgCitizenSearchGateway` - Default API to use for search, some wikis might prefer REST API
* `$wgCitizenSearchDescriptionSource` - MH supports ShortDescription, Wikidata, Description2, and WikiSEO, so it is better to let the user pick where they get the search suggestion description from.
* `$wgCitizenMaxSearchResults` - Max search result in suggestions

Note that `$wgCitizenSearchDescriptionSource` has both prerequisites:
* `$wgCitizenEnableSearch` has to be `true`
* `$wgCitizenSearchGateway` has to be `mwActionApi`
But I am not sure whether `requires` support multiple items so I have only put `$wgCitizenSearchDescriptionSource`.

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -3296,7 +3296,7 @@ $wgManageWikiSettings = [
 		'help' => 'Source of description text on search suggestions',
 		'requires' => [
 			'settings' => [
-				'setting' => 'wgCitizenSearchDescriptionSource',
+				'setting' => 'wgCitizenSearchGateway',
 				'value' => 'mwActionApi',
 			],
 		],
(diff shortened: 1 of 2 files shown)


ManageWikiSettings.php
```diff
@@ -3255,6 +3255,68 @@
...
+			'TextExtracts' => 'textextracts',
+			'ShortDescription/Wikidata' => 'wikidata',
+			'Description2/WikiSEO' => 'pagedescription',
+		],
+		'overridedefault' => 'textextracts',
+		'section' => 'styling',
+		'help' => 'Source of description text on search suggestions',
+		'requires' => [
+			'settings' => [
+				'setting' => 'wgCitizenSearchDescriptionSource',
```
Universal-Omega, line 3299: wgCitizenSearchGateway

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #4700, opened 2022-05-20
T9250: revoke "read" right from "user" usergroup on feignwiki

ugochimobi:


RhinosF1 (changes requested):
No. No. This will make the wiki not usable.

ugochimobi:
> No. No. This will make the wiki not usable.
yeah, that's what it means

Universal-Omega (changes requested):
This will revoke it from everyone, including bureaucrats, and administrators I think. If the wiki needs privated, it should be done from ManageWiki.

PR #4718, opened 2022-05-29
Improve loading of extensions

Universal-Omega:
This should also hopefully resolve the need to run setupStore.php for SemanticMediaWiki on all MediaWiki appservers.

### DIFF
--- a/initialise/MirahezeFunctions.php
+++ b/initialise/MirahezeFunctions.php
@@ -333,9 +333,7 @@ class MirahezeFunctions {
 
 		// Assign extensions variables now
 		if ( isset( $this->cacheArray['extensions'] ) ) {
-			foreach ( $this->cacheArray['extensions'] as $var ) {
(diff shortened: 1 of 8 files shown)


initialise/MirahezeFunctions.php
```diff
@@ -262,15 +279,15 @@ public static function isMissing() {
 		return !in_array( self::getCurrentDatabase(), $wgConf->wikis );
 	}
 
-	public function getCacheArray() {
+	public function getCacheArray(): array {
 		// If we don't have a cache file, let us exit here
 		if ( !file_exists( self::CACHE_DIRECTORY . '/' . $this->dbname . '.json' ) ) {
 			return false;
```
lens0021, line 285: This doesn't seem to match the defined return type(`array`).

Universal-Omega: Oh, thanks!

PR #4728, opened 2022-06-03
T9328: set base of wgMinervaShowCategories to true for criticalrolewiki

ugochimobi:
defaults confirmed from skin.json

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2958,6 +2958,18 @@ $wgConf->settings += [
 			'amc' => true,
 		],
 	],
+	'wgMinervaShowCategories' => [
+		'default' => [
+			'base' => false,
+			'loggedin' => false,
+			'amc' => false,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2958,6 +2958,18 @@
 			'amc' => true,
 		],
 	],
+	'wgMinervaShowCategories' => [
+		'default' => [
+			'base' => false,
+			'loggedin' => false,
+			'amc' => false,
```
Universal-Omega, line 2965: This is enabled by default.

Universal-Omega (changes requested):

LocalSettings.php
Universal-Omega: Should also be true for override unless requested otherwise.

Reception123 (approved):

PR #4729, opened 2022-06-04
T9331: set up CC BY-SA 3.0 for sekatetwiki

ugochimobi:


Reception123:
I have checked contributions on the wiki and can confirm that only the user who requested the change has edits on that wiki

Reception123 (approved):

PR #4761, opened 2022-06-16
Post-upgrade sitenotice

AgentIsai:


Sitenotice.php
```diff
@@ -30,17 +30,17 @@ function onSiteNoticeAfter( &$siteNotice, $skin ) {
 // }
 
 // Specific wiki SiteNotice
-/* if ( $wi->isAnyOfExtensionsActive( 'AddThis', 'FancyBoxThumbs', 'Foreground', 'GettingStarted', 'HeaderFooter', 'MagicNumberedHeadings', 'Pivot' ) ) {
+if ( $wi->isAnyOfExtensionsActive( 'Vector' ) ) {
 	$wgHooks['SiteNoticeAfter'][] = 'onSiteNoticeAfter';
 
 	function onSiteNoticeAfter( &$siteNotice, $skin ) {
```
Universal-Omega, line 36: Needs to be another function name when both enabled or fatal.

Universal-Omega (changes requested):

Universal-Omega:
Wait, vector is global, so using that check will never show since now from ManageWiki

PR #4798, opened 2022-06-26
Add settings for Share extension

AgentIsai:
This PR adds settings for the Share extension


LocalSettings.php
```diff
@@ -3965,12 +3965,37 @@
...
+	'wgShareUsePlainLinks' => [
+		'default' => true,
+	],
+	
+	// Site notice opt out
 	'wmgSiteNoticeOptOut' => [
 		'default' => false,
 	],
-
+
```
Universal-Omega, line 3998: Tabbing

Universal-Omega (changes requested):

ManageWikiSettings.php
```diff
@@ -2429,6 +2429,78 @@
 			],
 		],
 	],
+	'wgShareUsePlainLinks' => [
+		'name' => 'Use plain links for sidebar share menu',
+		'from' => 'share',
+		'type' => 'check',
+		'overridedefault' => true,
+		'section' => 'socialtools',
```
Universal-Omega, line 2437: If we are adding this to "Social tools", we need to rename the section to "Social", as "Social tools" is the social tools extensions (IE SocialProfile), or configuration involving the social tools extensions.

Universal-Omega (changes requested):

ManageWikiSettings.php
AgentIsai: If you're fine with that then sure but I added it there as I really can't think of where else to put it.

LocalSettings.php
AgentIsai: Fixed, thanks

ManageWikiSettings.php
Universal-Omega: I will just do the section rename later. This can be merged as-is for now I suppose.

Universal-Omega (approved):

PR #4806, opened 2022-06-30
Disable `$wgCosmosSearchUseActionAPI` by default

Universal-Omega:


Universal-Omega:
After 1.39 this will be done.

PR #4831, opened 2022-07-30
add PageProperties ext. per T9432

Reception123:


Universal-Omega:
It needs sql installed, also not an SMW extension. At least not required I don't think, I thought that was optional, for extra features. I could be wrong, and can double check in a bit if needed.

Universal-Omega (approved):

PR #4843, opened 2022-08-04
Add ContactPage configuration for Meta

AgentIsai:
Creates the 'requestaccount' ContactPage form on Meta

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -140,6 +140,53 @@ switch ( $wi->dbname ) {
 		wfLoadExtension( 'GlobalWatchlist' );
 
 		break;
+	case 'metawiki':
+		$wgContactConfig['requestaccount'] = [
+			'RecipientUser' => 'Miraheze CVT',
+			'SenderName' => 'Account Creation Request Form via Meta',
+			'RequireDetails' => true,
+			'IncludeIP' => false,
+			'MustBeLoggedIn' => false,
+			'AdditionalFields' => [
+				'SelectIssue' => [
+					'label-message' => 'contactpage-requestaccount-selectissue',
+					'type' => 'radio',
+					'options-messages' => [
+						'contactpage-requestaccount-selectissue-abusefilterissue' => 'abusefilter',
+						'contactpage-requestaccount-selectissue-recaptchaissues' => 'captcha',
+						'version-other' => 'other',
+					],
+					'help-message' => 'contactpage-requestaccount-selectissue-help',
+					'required' => true,
+				],
+				'DescribeIssue' => [
+					'label-message' => 'contactpage-requestaccount-describeissue',
+					'type' => 'text',
+					'hide-if' => [
+						'!==',
+						'SelectIssue',
+						'other'
+					],
+					'help-message' => 'contactpage-requestaccount-describeissue-help',
+					'required' => true,
+				],
+				'SelectUsername' => [
+					'label-message' => 'contactpage-requestaccount-selectusername',
+					'type' => 'text',
+					'maxlength' => 50,
+					'help-message' => 'contactpage-requestaccount-selectusername-help',
+					'required' => true,
+				],
+				'OtherDetails' => [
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -139,6 +139,53 @@ function onBeforePageDisplay( OutputPage $out ) {
...
+	case 'metawiki':
+		$wgContactConfig['requestaccount'] = [
+			'RecipientUser' => 'Miraheze CVT',
+			'SenderName' => 'Account Creation Request Form via Meta',
+			'RequireDetails' => true,
+			'IncludeIP' => false,
+			'MustBeLoggedIn' => false,
+			'AdditionalFields' => [
+				'SelectIssue' => [
+					'label-message' => 'contactpage-requestaccount-selectissue',
```
Universal-Omega, line 152: This and all following messages exist?

LocalWiki.php
```diff
@@ -139,6 +139,53 @@ function onBeforePageDisplay( OutputPage $out ) {
 	case 'loginwiki':
 		wfLoadExtension( 'GlobalWatchlist' );
 
+		break;
+	case 'metawiki':
+		$wgContactConfig['requestaccount'] = [
+			'RecipientUser' => 'Miraheze CVT',
```
Universal-Omega, line 145: This user exists?

Universal-Omega (commented):
Additionally, looking at source, it appears ContactPage itself requires a captcha as well.

Default fields still exist also, https://github.com/wikimedia/mediawiki-extensions-ContactPage/blob/33f93e31afaa0375697b1229335caf684cdc63b2/includes/SpecialContact.php#L172-L198, just an FYI I thought I would mention.

LocalWiki.php
AgentIsai: It does

LocalWiki.php
AgentIsai: They will be created on Meta once this is merged

AgentIsai:
> Additionally, looking at source, it appears ContactPage itself requires a captcha as well.

It does not actually. I have tested this in a local enviornment and did so without having Captcha configured for it. While you can configure Captcha, it is not required.

> Default fields still exist also, https://github.com/wikimedia/mediawiki-extensions-ContactPage/blob/33f93e31afaa0375697b1229335caf684cdc63b2/includes/SpecialContact.php#L172-L198, just an FYI I thought I would mention.

Indeed, thanks for mentioning it

Universal-Omega:
Also, it may be overridden by https://github.com/miraheze/mw-config/blob/5eadef96a72d7687cba5ccbe593e013d49d02a9e/GlobalSettings.php#L246 since the extension merges new form type with the defaukt config. Not sure which one overrides.

AgentIsai:
GlobalSettings seems to be loaded before LocalWiki, perhaps this will override that?

Universal-Omega:
And it only loads non-default forms when using URL query (?formtype) or subpage, Special:ContactPage/formtype, something else I thought I would mention, not by default. https://github.com/wikimedia/mediawiki-extensions-ContactPage/blob/33f93e31afaa0375697b1229335caf684cdc63b2/includes/SpecialContact.php#L90

Universal-Omega:
> GlobalSettings seems to be loaded before LocalWiki, perhaps this will override that?

If you set the config fully, this is appending to the config, not setting it.

Universal-Omega:
And the default key is also **required** by the extension.

AgentIsai:
Is it? Wikimedia projects don't configure it, see [Meta-Wiki](https://meta.wikimedia.org/wiki/Special:Contact)

Universal-Omega:
> Is it? Wikimedia projects don't configure it, see [Meta-Wiki](https://meta.wikimedia.org/wiki/Special:Contact)

Yes they do. https://github.com/wikimedia/operations-mediawiki-config/blob/59219fff5298f34cf0df34828f99e0e3dbef054f/wmf-config/CommonSettings.php#L1439-L1456

PR #4874, opened 2022-08-15
Database: Fix externalLoads

paladox:


Universal-Omega:
This is incorrect. This is correct as is, as it is a key to sectionLoads if I am not mistaken. It works for me as it should using --dbcluster=parsercache for example.

paladox:
https://github.com/wikimedia/operations-mediawiki-config/blob/e3fe6c04c95717f0f914bbfa366f5f827f392b6b/wmf-config/db-labs.php#L47

https://github.com/wikimedia/operations-mediawiki-config/blob/e3fe6c04c95717f0f914bbfa366f5f827f392b6b/wmf-config/db-labs.php#L39

PR #4887, opened 2022-09-05
Prevent autocreation on internal wikis

AgentIsai:
Autocreation is not needed on internal wikis and having accounts not be created automatically would provide more clarity for users that X wiki is not meant for help, as people have been confused before on the purpose of some of these wikis if they can't access them


Reception123:
On such wikis there's really no need for accounts to be autocreated as non-authroised accounts have no business on them anyway. Arguably, a case could be made that all private wikis should be like that but that's an argument for another time.

Reception123 (approved):

PR #4898, opened 2022-09-11
Stop using config cache on MediaWiki 1.39

Universal-Omega:


RhinosF1:
Why? (Interested)

Universal-Omega:
> Why? (Interested)

Because SiteConfiguration was supposed to have some performance improvements in 1.39 that it was possible the cache could actually be slower, so I am running some tests on 1.39 to see performance.

PR #4915, opened 2022-10-05
add default for wgCreatePageUwUseVE on LS.php

ugochimobi:


Universal-Omega:
This should be part of the same PR as adding to ManageWiki for history consistency.

PR #4935, opened 2022-10-21
T9847: Set wgShortDescriptionEnableTagline to false on slymodswiki

ugochimobi:


ManageWikiSettings.php
```diff
@@ -1182,6 +1182,15 @@
...
 		'requires' => [],
 	],
+	'wgShortDescriptionEnableTagline' => [
+		'name' => 'ShortDescription Enable Tagline',
+		'from' => 'shortdescription',
+		'type' => 'check',
+		'overridedefault' => true,
+		'section' => 'parserfunctions',
+		'help' => 'Enables short descritption in site tagline',
+		'requires' => [],
```
Reception123, line 1192: This would need 'requires' => [
			'extensions' => [ as it wouldn't make sense to use this setting without the extension being enabled.

Reception123 (approved):

PR #4936, opened 2022-10-21
T9850: Setup babel cats globally

ugochimobi:
See discussion here: https://meta.miraheze.org/wiki/Community_noticeboard/Archive_31#Babel_-_change_of_configuration_(proposal)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -244,7 +244,29 @@ $wgConf->settings += [
 	'wgArticlePlaceholderRepoApiUrl' => [
 		'default' => 'https://www.wikidata.org/w/api.php',
 	],
-
+	
+	// Babel
+	'wgBabelCategoryNames' => [
+		'default' => [
+			'0' => false,
+			'1' => false,
+			'2' => false,
+			'3' => false,
+			'4' => false,
+			'5' => false,
+			'N' => false,
+		],
+		'metawiki' => [
+			'0' => 'User %code%-0',
+			'1' => 'User %code%-1',
+			'2' => 'User %code%-2',
+			'3' => 'User %code%-3',
+			'4' => 'User %code%-4',
+			'5' => 'User %code%-5',
+			'N' => 'User %code%-N',
+		],
+	],
+	
 	// BetaFeatures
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -244,7 +244,29 @@ function wfShouldEnableSwift( $dbname ) {
 	'wgArticlePlaceholderRepoApiUrl' => [
 		'default' => 'https://www.wikidata.org/w/api.php',
 	],
-
+	
+	// Babel
+	'wgBabelCategoryNames' => [
+		'default' => [
```
Reception123, line 250: This isn't the right default
https://github.com/wikimedia/mediawiki-extensions-Babel/blob/master/extension.json#L66

Reception123 (changes requested):

LocalSettings.php
ugochimobi: > This isn't the right default https://github.com/wikimedia/mediawiki-extensions-Babel/blob/master/extension.json#L66

done

LocalSettings.php
```diff
@@ -244,7 +244,33 @@ function wfShouldEnableSwift( $dbname ) {
 	'wgArticlePlaceholderRepoApiUrl' => [
 		'default' => 'https://www.wikidata.org/w/api.php',
 	],
-
+
```
Universal-Omega, line 247: spacing

Universal-Omega (changes requested):

ugochimobi:
done?

Reception123:
I thought we said since the discussion was on CN we'd apply it globally. If wikis want to opt out they can.

ugochimobi:
The CN thread didn't say anything like that, but I'd support applying it globally. Wdyt?

Reception123:
Well by definition CN is the noticeboard for all Miraheze's wikis and Meta:Community Portal is for Meta matters. Either way, it's quite uncontroversial and if the community didn't want it it could've said so in the CN post.

I recommend making it the default.

ugochimobi:
Done, I think that should be all now.

Reception123 (approved):

PR #4939, opened 2022-10-25
T9859: Setup Data namespace for arquivopkmnwiki

ugochimobi:


Universal-Omega:
This is not the configuration that task asked for. Some stuff they did not ask for, and some stuff they did is missing.

ugochimobi:
What they asked for already exists as default on LocalSettings.php (some of them) which I didn't bother to set again.

PR #4974, opened 2022-11-02
Enable secure containers

Universal-Omega:
Also make all private wikis use private uploads

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1017,12 +1017,10 @@ $wgConf->settings += [
 		'default' => true,
 	],
 	'wgCreateWikiUseSecureContainers' => [
-		'default' => false,
-		'betaheze' => true,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1007,12 +1007,10 @@
 		'default' => true,
 	],
 	'wgCreateWikiUseSecureContainers' => [
-		'default' => false,
```
paladox, line 1010: Why is this true by default? it'll mean all wikis will be created as private containers if setContainersAccess.php is run.

Universal-Omega: > Why is this true by default? it'll mean all wikis will be created as private containers if setContainersAccess.php is run.

If true, it enables them on private wikis.  (except deleted and temp which is always private). Not all wikis.

PR #4998, opened 2022-12-01
disable localClientDatabases, enable allowDataAccessInUserLanguage on gratisdatawiki

ugochimobi:


Wikibase.php
```diff
@@ -269,6 +264,8 @@
 	$wgWBRepoSettings['useKartographerGlobeCoordinateFormatter'] = true;
 	// Data-Bridge
 	$wgWBRepoSettings['dataBridgeEnabled'] = true;
+	$wgWBRepoSettings['allowDataAccessInUserLanguage'] = true;
+	$wgWBRepoSettings['entityAccessLimit'] = '500';
```
Universal-Omega, line 268: Should this be an integer?

ugochimobi: > Should this be an integer?

Yeah, it is, but I probably didn't insert it correctly? but it's an integer.

Universal-Omega: Remove the quotations, otherwise it is a string.

ugochimobi: > Remove the quotations, otherwise it is a string.

ah, done, thanks.

PR #5006, opened 2022-12-04
Move static-new to static

paladox:


paladox:
https://github.com/miraheze/mw-config/pull/5005

PR #5010, opened 2022-12-11
Add basic options for the chameleon skin

govynnus:
Makes `$egChameleonLayoutFile` and `$egChameleonEnableExternalLinkIcons` configurable.

https://github.com/ProfessionalWiki/chameleon/blob/master/docs/customization.md#layout-of-page-elements
https://github.com/ProfessionalWiki/chameleon/blob/master/docs/customization.md#enable-external-link-icons


govynnus:
Do you know if there's any way to add dictionary-like options with arbitrary key-value pairs? Or would this require additions to ManageWiki?

That kind of input would be required for the other Chameleon settings, e.g. [$egChameleonExternalStyleVariables](https://github.com/ProfessionalWiki/chameleon/blob/master/docs/customization.md#changing-existing-scss-variables). (I think, unless there's a clever way to do it without).

Universal-Omega:
> Do you know if there's any way to add dictionary-like options with arbitrary key-value pairs? Or would this require additions to ManageWiki?
> 
> That kind of input would be required for the other Chameleon settings, e.g. [$egChameleonExternalStyleVariables](https://github.com/ProfessionalWiki/chameleon/blob/master/docs/customization.md#changing-existing-scss-variables). (I think, unless there's a clever way to do it without).

Something like `list-multi-bool` (IE what `$wgAbuseFilterActions` does in ManageWiki)? That requires a set one, but not arbitrary, and that would require changes to ManageWiki.

LocalSettings.php
```diff
@@ -477,6 +477,9 @@
 		'obeymewiki' => '/srv/mediawiki/w/skins/chameleon/layouts/fixedhead.xml',
```
Universal-Omega, line 477: Wiki overrides should be removed here now. Keeping only default.

ManageWikiSettings.php
```diff
@@ -3548,6 +3548,31 @@
...
+		'overridedefault' => '/srv/mediawiki/w/skins/chameleon/layouts/standard.xml',
+		'section' => 'styling',
+		'help' => 'The layout to use for the Chameleon skin.',
+		'requires' => [],
+	],
+	'egChameleonEnableExternalLinkIcons' => [
+		'name' => 'Chameleon external link icons',
+		'from' => 'chameleon',
+		'type' => 'check',
+		'overridedefault' => 'false',
```
Universal-Omega, line 3571:
```suggestion
		'overridedefault' => false,
```

Universal-Omega (changes requested):

ManageWikiSettings.php
govynnus: woops

LocalSettings.php
govynnus: Done. Is there a way to let those wikis know or set the option for them?

Universal-Omega: > Done. Is there a way to let those wikis know or set the option for them?

We can set them once merged.

govynnus:
> Something like `list-multi-bool` (IE what `$wgAbuseFilterActions` does in ManageWiki)? That requires a set one, but not arbitrary, and that would require changes to ManageWiki.

Yeah I think `list-multi-bool` wouldn't quite work, would the project be open to / interested in those changes to ManageWiki?

Universal-Omega:
> > Something like `list-multi-bool` (IE what `$wgAbuseFilterActions` does in ManageWiki)? That requires a set one, but not arbitrary, and that would require changes to ManageWiki.
> 
> Yeah I think `list-multi-bool` wouldn't quite work, would the project be open to / interested in those changes to ManageWiki?

I think if the changes were made, we could implement it yeah.

govynnus:
> I think if the changes were made, we could implement it yeah.

Cool :)

Universal-Omega (approved):

PR #5033, opened 2022-12-19
Remove c6 from maintenance

AgentIsai:


Universal-Omega:
It is still down. This will make it instead throw random errors.

AgentIsai:
Can we not make it also use the same store as the one used by affected c3 wikis?

Universal-Omega:
* #5035

PR #5036, opened 2022-12-20
T9540: set `$wgMirahezeReportsBlockAlertKeywords`

Universal-Omega:


OwenBaines:
I thin mostly just need variations of those mostly - so underage, under age, under 13, death threats, deaththreats, death threat.

Other ones to include might be:
- child pornography
- images of children
- images of minors
- suicide
- kill me
- kill themselves
- kill themselfs
- kill themself
- murder

And any more are obviously welcome

PR #5038, opened 2022-12-21
T10180: add editguide userright for projectsekaiwiki

ugochimobi:


LocalSettings.php
```diff
@@ -3936,6 +3936,9 @@
 			'editextendedconfirmedprotected',
 			'edittemplateprotected',
 		],
+		'+projectsekaiwiki' => [
```
Universal-Omega, line 3939: They did not request a restriction level.

Universal-Omega (changes requested):

LocalSettings.php
ugochimobi: > They did not request a restriction level.

I thought this was required?

ugochimobi: removed

Universal-Omega (approved):

PR #5060, opened 2023-01-08
T10264: Add $egAutoCreatePageNamespaces and set a custom value for constantnoblewiki

redbluegreenhat:
Requested on Phabricator: https://phabricator.miraheze.org/T10264

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1,4 +1,4 @@
-<?php
+ç<?php
 
 /**
  * LocalSettings.php for Miraheze.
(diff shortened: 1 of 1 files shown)


Universal-Omega:
This is already in ManageWikiNamespaces.

Universal-Omega:
It's `$wgAutoCreatePageNamespaces` also, because we don't use the original version as it was broken. I maintain the AutoCreatePage version we use, which is the `$wg` config.

redbluegreenhat:
Then this can safely be closed. The SMW changes requested there still need to be looked at though.

PR #5061, opened 2023-01-10
re-enable global renames

Reception123:
Do not merge yet unless certain but we should re-enable global renames ASAP as there's no reason to extend the suspension indefinitely now with db141 resolved


Universal-Omega:
Superceded

PR #5069, opened 2023-01-15
Minor tweaks to Sitenotice

AgentIsai:
This PR adds a 'learn more' button to the sitenotice.

It also reverts the colors to the pre-Swift migration colors and information icon.

### DIFF
--- a/Sitenotice.php
+++ b/Sitenotice.php
@@ -31,7 +31,7 @@ $wgMajorSiteNoticeID = 81;
 				<div data-nosnippet div style="padding-top:0.3em; padding-bottom:0.1em;">
 					<div data-nosnippet style="padding-top:0.3em; padding-bottom:0.1em;">
 						<div class="floatleft"><img alt="Information" src="https://upload.wikimedia.org/wikipedia/commons/c/ca/OOjs_UI_icon_info.svg" decoding="async" width="50" height="50"></div>
-						<div style="padding-bottom: 15px; font-size: 13pt; color: white;">
+						<div style="padding-bottom: 15px; font-size: 13pt; color: black;">
 							Miraheze will be upgrading to the latest version of MediaWiki (1.39) on Saturday, 21 January 2023 from <b><u>22:00 UTC</u></b> to approximately 01:00 UTC on Sunday, 22 January 2023. During this time, you will not be able to save changes to your wiki. Please make sure to save any edits at least 5 minutes before the upgrade begins.
 						</div>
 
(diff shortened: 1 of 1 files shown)


Sitenotice.php
```diff
@@ -27,20 +27,21 @@ function wfGlobalSiteNotice( &$siteNotice, $skin ) {
 
 		$siteNotice .= <<<EOF
 			<table style="width: 100%;">
-				<tbody><tr><td style="font-size: 120%; border-left: 4px solid #ff1e00; background-color: #ff5200cf; padding: 10px 15px; color: whitesmoke;">
+				<tbody><tr><td style="font-size: 120%; border-left: 4px solid #180F67; background-color: #F7F6FF; padding: 10px 15px;">
+				<div data-nosnippet div style="padding-top:0.3em; padding-bottom:0.1em;">
```
Universal-Omega, line 31: remove

Universal-Omega (changes requested):

Sitenotice.php
AgentIsai: Done

Sitenotice.php
```diff
@@ -27,20 +27,20 @@ function wfGlobalSiteNotice( &$siteNotice, $skin ) {
 
 		$siteNotice .= <<<EOF
 			<table style="width: 100%;">
-				<tbody><tr><td style="font-size: 120%; border-left: 4px solid #ff1e00; background-color: #ff5200cf; padding: 10px 15px; color: whitesmoke;">
+				<tbody><tr><td style="font-size: 120%; border-left: 4px solid #180F67; background-color: #F7F6FF; padding: 10px 15px;">
```
Universal-Omega, line 30: This will revert it back to a colorless site notice though. The colors weren't changed because of Swift but because of my conversion with about 6 or 7 different users recieving feedback on it.
Which mostly agreed that the new colors were better to normalise it with the colorscheme we use throughout, like on the landing page.

I think for the change I also talked with you about it back the also?

Universal-Omega (changes requested):

Sitenotice.php
Universal-Omega: Didn't mean to "Request changes" here, this is just a comment we can decide on.

AgentIsai: I actually made this PR primarily because I have heard from users across the farm that the colors are a bit of an eyesore and tend to stick out. We could look into other colors but I chose the Meta pastel blue colors as blue is associated with information.

Universal-Omega: I think we should just do the learn more button for now, as it is more important at the moment, and more time sensitive, but we can look into different colors later. If thats okay with you?

AgentIsai: That's fine

Universal-Omega:
Not needed anymore. Superceded.

PR #5072, opened 2023-01-20
T10313: Add custom values to wgCargoAllowedSQLFunctions for sagan4wiki and sagan4alphawiki

redbluegreenhat:
Requested on Phabricator: https://phabricator.miraheze.org/T10313.


Universal-Omega:
This can be in LocalWiki.php due to how large it is so that the whole thing doesn't need defined again.

LocalWiki.php
```diff
@@ -292,6 +292,12 @@ function onBeforeInitialize( Title &$title, $unused, OutputPage $output, User $u
 			}
 		}
 
+		break;
+	case 'sagan4wiki':
```
Universal-Omega, line 296: 2 case statements here will combine them since it is the same thing.

Universal-Omega (changes requested):

LocalWiki.php
redbluegreenhat: ?

Universal-Omega: Like https://github.com/miraheze/mw-config/blob/1713daaa1315f4a79cb8cdd7a2b93f77234deebc/LocalWiki.php#L310-L311

redbluegreenhat: I didn't even know you could do that in PHP.

Universal-Omega (approved):

PR #5073, opened 2023-01-21
Enable StopForumSpam globally

AgentIsai:
- [ ] Enable in Global settings
- [ ] Grant right to `user` by default?

### DIFF
--- a/GlobalExtensions.php
+++ b/GlobalExtensions.php
@@ -39,6 +39,7 @@ wfLoadExtensions( [
 	'Scribunto',
 	// 'SecureLinkFixer',
 	'SpamBlacklist',
+	'StopForumSpam',
 	'TitleBlacklist',
 	'TorBlock',
 	'WikiDiscover',


Naleksuh:
Can users opt out of enabling this?

AgentIsai:
I thought about making it a default extension but it wouldn't make sense with TorBlock also being global and not default.

Naleksuh:
It is different because whether or not an IP is a Tor exit node is an objective fact, this extension is just blocking edits based on a specific abuse report by a third party. It's not the responsibility of a wiki host to prevent people from editing even if the wiki might want them to

Reception123:
Couldn't the wiki manually unblock the IPs though? Or am I misunderstanding?

Naleksuh:
@Reception123 My understanding is that this extension doesn't use Special:Block and just prevents them from doing things, sort of like TorBlock

Reception123:
We will be adding sfsblock-bypass (similar to torunblocked) and therefore allowing users to opt out from it.

PR #5092, opened 2023-01-26
Unset no-ipinfo

AgentIsai:
Group not used on Miraheze


Universal-Omega:
As far as I know this won't work. $wgGroupPermissions in it's entirety is overriden by `$wgConf` set by what is in ManageWiki cache. If the group was never added to ManageWiki it shouldn't exist to begin with, so this wouldn't do anything.

PR #5095, opened 2023-01-27
Temporarily disable approving wikis

Universal-Omega:


Universal-Omega:
This is one way, but adding a config to CreateWiki to disable only approvals is probably a better way.

Reception123:
Even if there is a better way it's not worth risking that someone approves wikis again, so merging this temporarily until a better solution is implemented.

PR #5101, opened 2023-01-31
set wgIncidentReportingInactiveServices

Reception123:


Reception123:
I've tested this and it doesn't seem to have effect

LocalSettings.php
```diff
@@ -1772,6 +1767,16 @@
 			'Varnish' => 'https://meta.miraheze.org/wiki/Tech:Varnish',
 		],
 	],
+	'wgIncidentReportingInactiveServices' => [
```
Universal-Omega, line 1770: These services should stay in the other config, and this array should just include the names, not the URLs as well (at least for now in the current state of my PR, but if wanted it can be changed to be done like this)

Universal-Omega (dismissed):

LocalSettings.php
Universal-Omega: I might change the PR to be like this though instead.

Universal-Omega, line 1770: Parsoid and Bacula should be moved to inactive, and I do not think ElasticSearch should be.

Universal-Omega (changes requested):

LocalSettings.php
Reception123: I tried that initially and they stayed in the selectable list.

Reception123:
Closing due to quite a few issues, will wait until the MirahezeMagic PR is confirmed to work/someone else does it.

PR #5102, opened 2023-02-03
T10426: Use hCaptcha

AgentIsai:
Swaps out ReCaptchaNoCaptcha for hCaptcha


Universal-Omega:
Did #5103 instead which does this to, so I could do everything else (like site key) and easier testing.

PR #5107, opened 2023-02-04
remove StopForumSpam from ManageWiki (enabled globally)

Reception123:


redbluegreenhat:
What happens when previously-available extensions are removed from ManageWiki?

Universal-Omega:
> What happens when previously-available extensions are removed from ManageWiki?

They have to be disabled with toggleExtension.php or ManageWiki breaks.

Reception123:
disabled via script.

PR #5110, opened 2023-02-05
T10440: Add some Pivot settings to ManageWiki

redbluegreenhat:
https://phabricator.miraheze.org/T10440

The Pivot skin has an uncommon approach to settings: it doesn't register any configuration variables in skin.json, instead, it accesses an unregistered $wgPivotFeatures global array, and compares it with an internal variable to see if there are any new settings (https://github.com/wikimedia/mediawiki-skins-Pivot/blob/994a682983f867964e0486eb65a79aa2c410bab9/Pivot.skin.php#L13).

The idea here is to make them available through a ManageWiki "list-multi-bool" setting. If not possible due to the way the extension handles configuration, then just add the settings requested there to that wiki's section on LocalWiki.php.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -323,4 +323,8 @@ switch ( $wi->dbname ) {
 		$wgSpecialPages['Analytics'] = DisabledSpecialPage::getCallback( 'Analytics', 'MatomoAnalytics-disabled' );
 
 		break;
+	case 'wanderingstarswiki':
+		$wgPivotFeatures['showActionsForAnon'] = false;
+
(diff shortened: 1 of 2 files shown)


redbluegreenhat:
I decided to add all the settings except "wikiName" and "wikiNameDesktop", which can be indirectly configured, the footer icons one, as I think hiding things like the CC icon is not actually a good idea and the addThis settings.

If I should add any of those (or all of them), please let me know.

Reception123 (approved):

PR #5121, opened 2023-02-12
Revert "T10440: Add some Pivot settings to ManageWiki"

Reception123:
Reverts miraheze/mw-config#5110

Didn't work per T10440


Reception123:
@Universal-Omega Does something need to be done before removing settings, similarly to extensions?

Universal-Omega:
> @Universal-Omega Does something need to be done before removing settings, similarly to extensions?

Yes.`populateWikiSettings.php` with `--sourcelist=false` and `--remove`.

PR #5129, opened 2023-02-18
T10446: Set wgAllowImageTag for dgck81lnnwiki

redbluegreenhat:
Requested on Phabricator: https://phabricator.miraheze.org/T10446.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5544,6 +5544,11 @@ $wgConf->settings += [
 		'default' => '1.34',
 		'betaheze' => false,
 	],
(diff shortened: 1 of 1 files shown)


Universal-Omega:
It was deprecated and causes warnings in our logs IIRC. In my opinion this shouldn't be set for any wiki.

redbluegreenhat:
I kinda think the same thing, tbh. Your call.

DGCK81LNN:
If enabling this will lead to warnings then I shall not insist. It seems not all external images are allowed on Miraheze anyway, due to some HTTP headers. Please close this issue.

Universal-Omega:
Closing per above.

PR #5146, opened 2023-03-03
Remove metawiki autopatrol protection level + rename

AgentIsai:
- Removes metawiki autopatrolled protection level, will be assigned via ManageWiki instead for easy controlling
- Renames autopatrolled level to editautopatrolprotected
- Renames consul and bureaucrat protection level to editconsulprotected and editbureaucratprotected, per wmfphab:T230103

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4217,7 +4217,7 @@ $wgConf->settings += [
 			'edittemplateprotected',
 		],
 		'metawiki' => [
-			'autopatrolled',
+			'editautopatrolprotected',
 		],
 		'naasgamelandwiki' => [
 			'editarchiveprotected',
@@ -4238,8 +4238,8 @@ $wgConf->settings += [
 			'edittemplate',
 		],
 		'testwiki' => [
-			'bureaucrat',
-			'consul',
+			'editbureaucratprotected',
+			'editconsulprotected',
 		],
 		'trwdeploymentwiki' => [
 			'bureaucrat',
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2670,9 +2670,6 @@ $wgConf->settings += [
 			],
 		],
 		'+metawiki' => [
-			'autopatrolled' => [
-				'autopatrolled' => true,
-			],
 			'confirmed' => [
 				'mwoauthproposeconsumer' => true,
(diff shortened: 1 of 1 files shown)


Universal-Omega:
Are these used at all, if so, I am unsure if they can just be removed or renamed, or if they are if they need to ne updated in the database somewhere. I can't remember exactly how this works.

AgentIsai:
renameRestrictions.php can be run to change the protection level of pages. On Meta at least, nothing is at autopatrolled level. On testwiki, a few pages are at consul level.

redbluegreenhat:
autopatrolled is currently used at the Global Sysop page: https://meta.miraheze.org/w/index.php?title=Special:Log&logid=835707

AgentIsai:
> autopatrolled is currently used at the Global Sysop page: https://meta.miraheze.org/w/index.php?title=Special:Log&logid=835707

was, at least

Universal-Omega:
These need matching user rights though

redbluegreenhat:
> Are these used at all, if so, I am unsure if they can just be removed or renamed, or if they are if they need to ne updated in the database somewhere. I can't remember exactly how this works.

Pretty sure that once they're no longer used on a page, all there's of them is an entry in the logs. This entry consists of static text and is not dynamically generated (see the protection log for the English Wikipedia main page, which uses the old format: https://en.wikipedia.org/w/index.php?title=Special:Log&type=protect&page=Main+Page).

I'd bet that, once no pages are using it, it's okay to remove the protection level from the configuration, no need to edit the DB. However, if you want to be sure you can always test on betaheze. This is what testing servers are for after all, to break them.

AgentIsai:
renameProtections.php was written to [handle this](https://phabricator.wikimedia.org/T237194) as WMF is also in the process of standardizing their edit protection levels for i18n purposes.

LocalSettings.php
```diff
@@ -4038,6 +4035,7 @@
 	],
 
 	// Restriction types
+	// Per wmfphab:T230103, they should ideally follow the format of editXXprotected
```
Universal-Omega, line 4038: We aren't Wikimedia, and this does not have to be a standard on Miraheze if it isn't how individual wikis want them. The comment is probably unnecessary.

PR #5151, opened 2023-03-04
add viewsuppressed to oversight

The-Voidwalker:
seems to be a bug in MW 1.39 that suppressrevision no longer allows one to view (or modify the suppressed status of) suppressed revisions.


The-Voidwalker:
Seems to be caused by something in our setup, as I cannot replicate this issue when testing.

PR #5154, opened 2023-03-07
T9853: Install Skin:Splash in ManageWiki

brandon-wm:
PR to install the Splash skin on Miraheze following the security review and approval of T9853.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -3772,6 +3772,14 @@ $wgManageWikiExtensions = [
 		'requires' => [],
 		'section' => 'skins',
 	],
+	'splash' => [
+		'name' => 'Splash',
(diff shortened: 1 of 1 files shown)


Reception123 (approved):

Reception123:
Thanks for this PR. A very minor observation but it would probably be easier if #5155 was done in the same PR/on the same branch as this.

Universal-Omega:
> Thanks for this PR. A very minor observation but it would probably be easier if #5155 was done in the same PR/on the same branch as this.

When I do them, I actually like to do extension-list *first* so that I can rebuild LC *before* deployment of the actual extension/skin, but for a skin, that doesn't have to many *urgent* i18n messages, like this one, it probably is best to do them all together.

PR #5174, opened 2023-04-03
T10688: Set $wgChatKicks = true on zhtranswiki

brandon-wm:


redbluegreenhat (changes requested):
This should go on LocalSettings.php.

PR #5175, opened 2023-04-03
T10688: Set $wgChatKicks = true on zhtranswiki

brandon-wm:


redbluegreenhat:
Agent is right, this is the default

https://github.com/wikimedia/mediawiki-extensions-MediaWikiChat/blob/4858b916a0b6f94a315dc4dbe1b86e436e604430/extension.json#L15

As this is already the default, there should be no need to do anything.

PR #5181, opened 2023-04-05
Update genshinimpactwiki settings again

thecutefinder:


LocalWiki.php
```diff
@@ -81,10 +82,19 @@ function onHtmlPageLinkRendererEnd(
...
 		}
+		
+		function onPersonalUrls( array &$personal_urls, Title $title, SkinTemplate $skin ) {
+			$personal_urls[ 'preferences' ] = [
+				'text' => 'Preferences',
+				'href' => 'https://genshinimpact.miraheze.org/wiki/Genshin_Impact_Wiki:Preferences'
+			];
+		}
+		
+		$wgAllowSiteCSSOnRestrictedPages = true;
```
redbluegreenhat, line 97: This is a huge no-no. This is not going to be enabled.

redbluegreenhat (changes requested):

LocalWiki.php
thecutefinder: Ok, I removed it, but why? It is only CSS, not JS.

redbluegreenhat: Even if it's only CSS you can still do nasty things. I'd rather be safe than sorry.

thecutefinder: [https://phabricator.wikimedia.org/T72672 I see]. Is there any way to implement some of my custom CSS on the login and create account pages without enabling CSS on restricted pages altogether?

LocalWiki.php
```diff
@@ -81,11 +82,18 @@ function onHtmlPageLinkRendererEnd(
 				return true;
 			}
 
-			$attribs['rel'] = 'nofollow';
+			$attribs[ 'rel' ] = 'nofollow';
```
redbluegreenhat, line 85: This is not needed, please remove the spaces to keep it consistent with the code style used here.

redbluegreenhat (changes requested):

LocalWiki.php
redbluegreenhat: No.

LocalWiki.php
thecutefinder: No, the MediaWiki style guidelines say to use spaces like that.

redbluegreenhat:
I'm not a huge fan of this PR, I don't think it's needed to mess with the preferences page in any way. Requesting review from @Reception123 so they can give their thoughts on this.

thecutefinder:
Correct, it is not _needed_, it is a _styling_ change. There are no policies against this.

LocalWiki.php
redbluegreenhat: This is not Wikimedia, we don't have to use their code style. For examples of spaces not being used here, see https://github.com/miraheze/mw-config/blob/e7ca82a45d7b575c65ab97dc17a2d43c90c3bf22/LocalWiki.php#L6 https://github.com/miraheze/mw-config/blob/e7ca82a45d7b575c65ab97dc17a2d43c90c3bf22/LocalWiki.php#L21 and https://github.com/miraheze/mw-config/blob/e7ca82a45d7b575c65ab97dc17a2d43c90c3bf22/LocalWiki.php#L91 for a few examples.

Style consistency is important.

LocalWiki.php
thecutefinder: Actually, what if I added a hook in LocalWiki to add CSS to the login page's head tag? That way, the CSS would have to be reviewed before it can be added to the login page.

LocalWiki.php
thecutefinder: Done

thecutefinder:
@redbluegreenhat (Replying to what you said in the other thread so that I don't make it go off-topic)

Look, I don't mean to rude, I'm sorry if I come off as defensive, but this seems like a really major problem with how Miraheze is structured.

The implication is that is immoral, so as such, you should implement a rule against modifying the preferences page.

I wasn't going to modify it, but, unfortunately, specific sections that I want to be hidden in Special:Preferences cannot be for some reason.

Overriding the preferences page is the least convenient way to get a user's email address. I don't know why you're acting like it would be such an easy way to collect PII.

JavaScript can do many dangerous things, and I assumed that the reason Miraheze has no review process for it is because they prevented it in some way. Why is it entirely possible - and Miraheze does not do anything to prevent it - to run cryptominers and log a user's email address with vanilla JS? I was very surprised that the API allowed you to fetch a user's email address, and I assumed that as Miraheze is run by "expert volunteers" as the main page puts it and has been around for so many years, there must be something that would prevent you from abusing it. 

From what I can tell, there is a CORS policy that only gives you a select number of domains that you can fetch (and therefore send data to), but the requirements to get something added there are very slim. A GitHub Pages site is on there because one wiki wanted it for Scratch blocks. Every wiki can send and receive data to `wiki-assets.sumin.wiki`.

At the very least it should not be possible to fetch a user's email address.

PR #5182, opened 2023-04-05
add new createCargoDB.php to cargo install

The-Voidwalker:
Includes configuration of $wgCargoDBname to be equal to `$wgDBname . 'cargo'`

Intended for use with https://github.com/miraheze/MirahezeMagic/pull/413

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -421,3 +421,5 @@ switch ( $wi->dbname ) {
 
 		break;
 }
+
+$wgCargoDBname = $wgDBname . 'cargo';
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -421,5 +421,3 @@ switch ( $wi->dbname ) {
 
 		break;
 }
-
-$wgCargoDBname = $wgDBname . 'cargo';
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -272,6 +272,9 @@ $wgManageWikiExtensions = [
 			],
 		],
 		'install' => [
+			'mwscript' => [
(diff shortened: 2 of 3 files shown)


ManageWikiExtensions.php
```diff
@@ -272,6 +272,9 @@
 			],
 		],
 		'install' => [
+			'mwscript' => [
+				'$IP/extensions/MirahezeMagic/maintenance/createCargoDB.php' => [],
```
redbluegreenhat, line 276: To make the style more consistent with the rest of extensions that run maintenance scripts, please use double quotes in this string.

redbluegreenhat (changes requested):

ManageWikiExtensions.php
The-Voidwalker: Woops that's not even a styling thing, pretty sure that's required for `$IP` to work.

The-Voidwalker, line 276:
```suggestion
				"$IP/extensions/MirahezeMagic/maintenance/createCargoDB.php" => [],
```

redbluegreenhat (approved):
Looks OK for me, I'll leave the actual merging to someone else as this should be merged alongside the MirahezeMagic and Puppet PRs, the later of which I don't have permission to do.

LocalWiki.php
```diff
@@ -421,3 +421,5 @@ function onBeforePageDisplay( OutputPage $out ) {
 
 		break;
 }
+
+$wgCargoDBname = $wgDBname . 'cargo';
```
redbluegreenhat, line 425: I don't have any issues with it being here, but there shouldn't be any issues with it being in LocalSettings instead. I'll leave it up to you anyway.

PR #5185, opened 2023-04-07
Don't harcode the directory

paladox:


LocalSettings.php
```diff
@@ -288,7 +288,7 @@
 
 	// Cache
 	'wgCacheDirectory' => [
-		'default' => '/srv/mediawiki/cache',
```
MacFan4000, line 291: cache doesn't exist in staging

PR #5187, opened 2023-04-07
T10707: Add $wgCleanSignatures to Meowpedia

brandon-wm:


AgentIsai:
Why isn't this added to LocalSettings.php and ManageWikiSettings.php?

thecutefinder:
@AgentIsai #5188

brandon-wm:
@AgentIsai Wasn’t aware it was wanted to be a global option. If so can change.

LocalWiki.php
```diff
@@ -419,5 +419,9 @@ function onBeforePageDisplay( OutputPage $out ) {
 	case 'worldboxwiki':
 		$wgSpecialPages['Analytics'] = DisabledSpecialPage::getCallback( 'Analytics', 'MatomoAnalytics-disabled' );
 
+		break;
+	case 'meowpediawiki':
+		$wgCleanSignatures = false;
```
redbluegreenhat, line 424: This should go on LocalSettings.php instead imo. You can also just add it to ManageWiki if you want, but that's up to you.

brandon-wm: Wouldn’t adding to LS.php set false as default for all wikis?

AgentIsai: You can set the default value to true and then set the set it to false for that wiki only but in this case, it makes more sense to just add it into ManageWiki

brandon-wm: As long as the default remains true, I have no objections. ManageWiki can be done.

redbluegreenhat: Just noting that even if you're going the ManageWiki route, the default still needs to be set at LocalSettings, else it doesn't work properly.

brandon-wm:
So.....moving to LS.php and implementing in ManageWiki?

redbluegreenhat:
yes

PR #5188, opened 2023-04-08
Add $wgCleanSignatures = false

thecutefinder:
Self-explanatory.

image

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5712,5 +5712,9 @@ if (
 	require_once __DIR__ . '/ExtensionMessageFiles.php';
 }
 
+// https://github.com/miraheze/mw-config/pull/5187#issuecomment-1500630025
+// https://www.mediawiki.org/wiki/Manual:$wgCleanSignatures
+$wgCleanSignatures = false;
+
 // Don't need a global here
 unset( $wi );
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -1978,6 +1978,17 @@ $wgManageWikiSettings = [
 		'help' => 'Whether the option to enable/disable Page Previews should be hidden on Preferences page. Please note if PopupsBetaFeature is set to true this option will be always hidden.',
 		'requires' => [],
 	],
+	// T10707
+	// https://github.com/miraheze/mw-config/pull/5187
+	'wgCleanSignatures' => [
+		'name' => 'Clean signature',
+		'from' => 'mediawiki',
+		'type' => 'check',
+		'overridedefault' => true,
+		'section' => 'preferences',
+		'help' => 'Prevents users from including templates in their user signature',
+		'requires' => [],
+	],
 
 	// Recent changes
 	'wgFeedLimit' => [


thecutefinder:
did not test it

brandon-wm:
It would be appreciated if in the future PR names and content are of a serious and appropriate nature.

thecutefinder:
fixed

thelevking:
This change, if adopted, would change a global MediaWiki default without a discussion:
https://www.mediawiki.org/wiki/Manual:$wgCleanSignatures

I do not think that this is appropriate.

thecutefinder:
Hmm, where should I open a discussison?

redbluegreenhat:
No need for such a discussion, we can just not set that globally, as at the end of the day just one wiki has requested this be changed from the default.

thecutefinder:
Miraheze users propose changes to the software all the time, there is the Community Noticeboard, Requests for Comment page, etc.. I don't know who you are but it is not correct to reply to everything by saying "this will not happen" just because you personally disagree with the implementation.

redbluegreenhat:
I only replied that way to your request to setting restricted CSS on your wiki (security and privacy concerns) and your request to override the default link to the preferences page with one to a custom page using unvetted JS that has fields for setting your email, for privacy concerns. Even so, while I would normally just close those PRs, I still requested another colleague to review that specific PR.

Yes, they do propose changes to the software, but unless there's a very good reason not to, we make these changes only in their wikis, either via LocalSettings or via adding it to ManageWiki. Otherwise we try to stay close to default MediaWiki values wherever possible.

brandon-wm:
I would recommend proceeding with the previous PR on LocalWiki.php, and keep the default as true. This can be done on a case by case basis.

thecutefinder:
> override the default link to the preferences page with one to a custom page using unvetted JS that has fields for setting your email, for privacy concerns

I hope you realize that this has always been possible to do with JS and that there is no policy against it? [Here](https://meta.miraheze.org/wiki/Requests_for_Comment) is how you propose that a customization policy be implemented that forbids overriding the link to the preferences page.

thecutefinder:
> I would recommend proceeding with the previous PR on LocalWiki.php, and keep the default as true. This can be done on a case by case basis.

Ok, you're right, but I still think it should be added to ManageWiki. I only submitted the PR because Agent Isai said it should be the default in LocalSettings.

redbluegreenhat:
> > override the default link to the preferences page with one to a custom page using unvetted JS that has fields for setting your email, for privacy concerns
> 
> I hope you realize that this has always been possible to do with JS and that there is no policy against it? 

You can also do effectively whatever with JS, hell you could use it to run cryptominers on your visitors. Does that make it okay? These kinds of rules also are kinda like [Asshole John Rules](https://en.wikipedia.org/wiki/Wikipedia:Asshole_John_rule).

>[Here](https://meta.miraheze.org/wiki/Requests_for_Comment) is how you propose that a customization policy be implemented that forbids overriding the link to the preferences page.

This is not a wiki, RFCs and the such do not matter here. What gets or not included here is effectively up to SRE.

There's a reason why, unlike in wikis, write access is extremely limited. We have effectively an unspoken moral obligation to review incoming PRs to the best of our ability for potential concerns regarding topics like security and privacy (alongside many others). While there's no legal obligation to do this on our side (and this repository is also GPL-3.0, meaning no warranty or fitness for a particular purpose), there is a moral one.

Whatever I merge will essentially become partially my responsibility, as merging it means that I should have reviewed it and given my explicit approval. Therefore, you can be sure that if I have any concerns, I will voice them, as I should.

> I am fairly sure that the MediaWiki API allows fetching a user's email address. There is no JS review process. The lack of a review process is advertised as a feature.

See the first answer. Just because you can doesn't mean you should. Touching such data is not something you should do because you can, lest you run afoul of https://meta.miraheze.org/wiki/Privacy_Policy.

redbluegreenhat:
Close in favor of https://github.com/miraheze/mw-config/pull/5187.

thecutefinder:
@redbluegreenhat I'll reply to you in the other thread, and sorry if I came off as defensive, I re-read my replies today.

PR #5193, opened 2023-04-12
disable wiki creations

Naleksuh:


MacFan4000:
As far as I’m aware there is no need to disable wiki creations.

Naleksuh:
@MacFan4000 This occured in the wiki creators channel. If you are not aware then please do not close them

Naleksuh:
@MacFan4000 This was stated in the wiki creators channel. Why are you not aware? If there's something unrelated to cloud11, let us know, otherwise if you aren't aware then why interact?

PR #5204, opened 2023-04-19
Remove IPInfo extension T10754

Amanda-Catherine:
Doesn't work on Miraheze (and apparently will need premium features to be functional) and it's generating a rather clumsy user group that no one can use. CheckUser should be sufficient for related tasks.

### DIFF
--- a/GlobalExtensions.php
+++ b/GlobalExtensions.php
@@ -23,7 +23,6 @@ wfLoadExtensions( [
 	'ImportDump',
 	'IncidentReporting',
 	'Interwiki',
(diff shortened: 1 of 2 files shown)


RhinosF1:
IPInfo will need to rolled out at some point if the IP Masking project happens

redbluegreenhat:
I mean, if it isn't working (the data we currently have is the geolite2 dbs meant to test the extension, not the ones we would run in production), there's little reason to keep it imo. @Reception123 thoughts?

PR #5207, opened 2023-04-22
Create global-renamer group on Meta

AgentIsai:
Following a discussion with Reception123 and due to technical issues with CentralAuth and accounts being attached to too many wikis loading slowly, this group is created for Stewards only to assign to their alts


redbluegreenhat (approved):
Approved for use by the Stewards only, per you. Any other use requires the usual RfC stuff.

PR #5210, opened 2023-04-25
Modify $wgAvailableRights and $wgRestrictionLevels

Reno-Rex:
I want to modify these restriction levels for the Scratchpad Wiki here on Miraheze and show them in order. I am also adding templateeditor, extendedconfirmed, moderator, and bureaucrat permissions to $wgAvailableRights. I understand that the '', 'user', 'autoconfirmed', and 'sysop' groups are already defined by default, but I want them to be in order on the wiki I'm administrating. Thank you.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4150,6 +4150,7 @@ $wgConf->settings += [
                         'autoconfirmed',
 			'templateeditor',
 			'extendedconfirmed',
+			'moderator',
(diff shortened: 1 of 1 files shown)


Reno-Rex:
I also have not protected any pages with these custom levels yet, so there will be no conflicts in that regard.

Reception123 (approved):

PR #5215, opened 2023-04-28
T10773: Add $wgUFAllowedNamespaces

brandon-wm:


brandon-wm:
Seems that unintentional changes have been added

redbluegreenhat:
This is why it is recommended to `git pull` before starting work, as otherwise things like this happens because of not being in sync with upstream.

PR #5224, opened 2023-05-04
Userrights config for nycsubwaywiki

Amanda-Catherine:
I have no intention of ever holding local elections to these user groups, even once the wiki qualifies. Therefore condensing all restricted groups into one "steward" group to minimize the number of restricted groups/rights.


Amanda-Catherine:
@AgentIsai The point of having the rights in the local steward group is to ensure that any steward actions are locally documented/logged. I don't want any steward using their global rights from Meta to execute actions on my wiki - I want them to assign local steward from Meta and then use "local rights" to do whatever they need to do. Ideally, this would apply for CU/OS too... I'd prefer for stewards to have to generate a local user rights log entry when using the tools locally since the respective logs are private, but I don't think that's possible.

as for unsetting, it's something I've done before with previous wikis, though admittedly it was before your time as a steward.

AgentIsai:
It doesn't really matter what rights they use, whether they're locally assigned or globally assigned, if it's within Stewards scope to use them for valid purposes as defined by policy. From your statement, it seems you wish to restrict what Stewards can do to only these actions as you've set forth but unfortunately, that isn't how this works. I assure you though that you won't have Stewards treading on your wiki for non-valid purposes, we have far too much to do and we certainly don't want to waste our time going about bothering you.

I am very much aware you previously unset these groups in the past but none of the Stewards from that era are with Miraheze anymore, apart from Void, so consensus within the group has changed.

Amanda-Catherine:
Certainly not trying to restrict what Stewards can do from a technical standpoint (I'll be writing a local supplement to the global policy a bit later that will deal with when they should or shouldn't use their rights in non-emergency situations). What I am trying to do is make sure that I have a record of every local action that takes place (or doesn't take place but should've taken place), even if those actions are carried out by stewards and not local users. Because as far as I'm aware there is no technical way for me to be able to see that a CheckUser was performed without seeing who/what was checked, and likewise it's impossible for me to be able to see that a particular page was suppressed without being able to see the suppressed contents, the idea is that stewards would assign local "steward" rights from Meta, then assign local CU/OS rights using the local "userrights" permission, instead of doing CU/OS directly from Meta/Interwiki and then using the tools, which leaves no local trace of them having been used.

AgentIsai:
Well with you unsetting the CheckUser and Oversight groups and condensing them into Steward, there'd still be no log of that.

Amanda-Catherine:
Hmm... so now that I've looked in a bit more detail, it does appear that access to the suppression log is a distinct permission from the ability to suppress, and that merely viewing Special:log/suppress doesn't allow you to see the actual suppressed contents. If I'm correct about that, I would think that it would be permissible for me to be able to monitor the suppression log without violating the Privacy Policy. I think the same would go for the IP Info and AbuseFilter Private Access logs.

As for checkuser, because even viewing that log would create privacy issues, there's no way for me to monitor it, so I think I'll have to rely on putting a warning at the top of Special:CheckUser reminding stewards to log their usage of the tool locally; if they don't feel like doing the local Userrights thing, I could also have a project page for public documentation of CU usage (which I believe is already done at the German Wikipedia).

AgentIsai:
Your requests to view the suppression log and private AbuseFilter logs are declined.

Amanda-Catherine:
@AgentIsai think this should be good now

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
 				'edittemplateprotected' => true,
 			],
 		],
+		'+nycsubwaywiki' => [
+			'steward' => [
+				'managewiki-restricted' => true,
```
AgentIsai, line 2865: Why assign `managewiki-restricted` when the global group already has the right? Any local ManageWiki changes would be logged locally either way.

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+				'interwiki' => true,
```
AgentIsai, line 2866: The right is in the global group so is there a need for this?

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+				'siteadmin' => true,
```
AgentIsai, line 2867: No group, local or global, has or needs `siteadmin`, as that's all handled via LocalSettings.php by SRE.

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+				'bigdelete' => true,
```
AgentIsai, line 2868: Same as above, right is redundant.

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+				'renameuser' => true,
```
AgentIsai, line 2869: Renames are done at a global scale, local renames aren't ever practiced and no group has the right in their toolset so this can be removed.

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+			],
+			'checkuser' => [
```
AgentIsai, line 2871: Is there a reason to specifically define this and not let it use the defaults?

LocalSettings.php
```diff
@@ -2860,6 +2860,23 @@
...
+				'investigate' => true,
+				'ipinfo' => true,
```
AgentIsai, line 2873: IPInfo rights are already in the global toolset.

AgentIsai (changes requested):

Amanda-Catherine:
Ideally global permissions should be reserved for emergencies only... anything that might need to be done on a non-emergency basis should be allocated to some local user group. If having additional local rights defined won't break anything, is there a problem with that?

AgentIsai:
Policy doesn't define it that way though, Stewards can use their global rights as needed. It becomes needlessly tedious to have to assign ourselves the local `steward` group every time we wish to do something. If we ever needed to do something not within the local toolset then would we have to submit a PR and wait for it to be merged before handling a task? It would be silly. Additionally, note that it is not just Stewards with global reach, you also have Global Sysops, Global Rollbackers, and Global interwiki administrators who have access to local tools using their global toolset.

Amanda-Catherine:
I consider IPInfo to be effectively a CheckUser action, so I'm going to leave that one locally defined and require stewards to assign local CU before using it, but I suppose I can remove the others.

AgentIsai:
IPInfo is just a popup which tells you the connection details of an anonymous editor (not for logged in users). You can achieve the same by plugging in that IP to a website like ipinfo.io. The only reason it's currently restricted to CVT is because it's misconfigured so there's a huge probability that local sysops would complain about it not working as intended.

Amanda-Catherine:
I've been told by upstream folks that IPInfo will be considered a sensitive right once the IP Masking project happens.

PR #5225, opened 2023-05-05
Userrights config for nycsubwaywiki

Amanda-Catherine:
Cleaner version of #5224

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -323,6 +323,11 @@ switch ( $wi->dbname ) {
 		$wgFilterLogTypes['comments'] = false;
 
 		break;
+	case 'nycsubwaywiki':
+		unset($wgGroupPermissions['interwiki-admin']);
+		unset($wgGroupPermissions['noip-info']);
+		
+		break;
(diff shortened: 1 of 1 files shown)


AgentIsai (approved):

Amanda-Catherine:
@AgentIsai This doesn't seem to have worked.. the two user groups still exist.

AgentIsai:
> This won't work. $wgConf can't be overriden like that or anywhere of ease. Just doesn't work like that from what I have experienced.

How can groups be unset then, at least for extensions?

Amanda-Catherine:
That would just remove all the permissions from the groups, but the groups themselves would still be intact. I know it's possible to completely unset extension-defined groups, as I've done it before... but maybe the devs have changed something? Does anyone subscribe/contribute to the MediaWiki-I mailing list at WMF who could ask on the list?

AgentIsai:
ManageWiki has changed an awful lot since you were last on Miraheze, Universal Omega has changed quite a few things so maybe that's why.

PR #5230, opened 2023-05-17
T10840: Set $wgDefaultLanguageVariant = zh-Hans-CN on zhtranswiki

brandon-wm:


ugochimobi:
Please Update it to zh-cn per the phab request @brandon-wm

brandon-wm:
Updated, though no need to request in two places as I can see via Phab and will take care of it when I can.

PR #5232, opened 2023-05-24
Set WebAuthn configuration to allow security key login on all *.miraheze.org

AgentIsai:
Currently, you can only login with your security key to the wiki you registered on. This change will allow your security key to work on all *.miraheze.org domains (but still not custom domains). i18n should be updated accordingly. I will test this out tomorrow.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5117,6 +5117,15 @@ $wgConf->settings += [
 		'default' => '1 hour',
 	],
 
+	// WebAuthn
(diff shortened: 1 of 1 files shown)


MacFan4000:
According to https://phabricator.wikimedia.org/T244088#5875266 this would break 2fa users who currently have a registered key.

AgentIsai:
A very, *very* small number of users use WebAuthn. I will email them about this change and will ask them to temporary use TOTP or disable 2FA while this change is done.

PR #5242, opened 2023-06-06
[1.40] Use new parser media structure globally

RhinosF1:
This is default false in 1.40


RhinosF1:
We need to make sure this has been tested on beta. I've spoken to @Reception123

paladox:
https://github.com/miraheze/mw-config/blob/cade799729b1ddced5d12351c492f127d8aa4271/GlobalSettings.php#L144

PR #5258, opened 2023-06-23
Turn on DiscussionTools enhancements in isvwiki

thelevking:
I am not sure if `'default'` is required here. If it is, I can fix the patch. I followed https://noc.wikimedia.org/conf/InitialiseSettings.php.txt on variable names and values.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1213,6 +1213,20 @@ $wgConf->settings += [
 		'default' => 'en',
 	],
 
+	// DiscussionTools
+	'wgDiscussionTools_visualenhancements' => [
+		'isvwiki' => 'available',
+	],
+	'wgDiscussionTools_visualenhancements_reply' => [
+		'isvwiki' => 'available',
+	],
+	'wgDiscussionTools_visualenhancements_pageframe' => [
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1213,6 +1213,24 @@
...
+	],
+	'wgDiscussionTools_visualenhancements_reply' => [
+		'default' => 'default',
+		'isvwiki' => 'available',
+	],
+	'wgDiscussionTools_visualenhancements_pageframe' => [
+		'default' => 'default',
+		'isvwiki' => 'available',
+	],
+	'wgDiscussionTools_visualenhancements_newsectionlink_enable' => [
```
The-Voidwalker, line 1229: Looks like this setting isn't currently available in the version of the extension we're using. We'd need to update to master branch (or REL1_40) to support it. Additionally, it appears that [the default value is true](https://github.com/wikimedia/mediawiki-extensions-DiscussionTools/blob/6ea3c7099dbec2ff7e6063b7dc5291ae517d3486/extension.json#L570-L573). WMF probably have it as false in their settings to phase in the feature.

thelevking: Removed it.

PR #5266, opened 2023-06-28
Cleanup

ugochimobi:


LocalSettings.php
```diff
@@ -3651,7 +3651,7 @@
 		'default' => true,
 	],
 
-	// T11025
+	// NearbyPages
```
The-Voidwalker, line 3654:
```suggestion
	// NearbyPages - T11025
```
Might be worth keeping the task mentioned here.

PR #5320, opened 2023-09-20
Multiple tasks: Configure LinkTargetParentClass for some wikis

redbluegreenhat:
I believe this is how you do it. Tasks are https://phabricator.miraheze.org/T11084, https://phabricator.miraheze.org/T11101, https://phabricator.miraheze.org/T11102.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2478,13 +2478,13 @@ $wgConf->settings += [
 			'_self' => [ 'sametablinks' ]
 		],
 		'scruffywiki' => [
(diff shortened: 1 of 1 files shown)


robkam:
Thanks, that looks correct except I want the links to open in a new window with target="_blank", _self opens them in the same window?

redbluegreenhat:
@robkam you're right :p, take a look again

robkam:
Thank you :-)

robkam (approved):

PR #5408, opened 2023-12-27
Add $wgDiscordName to ManageWiki (T11459)

R4356th:


OAuthority:
This can't be merged as its failed a test. Please adapt the code so it can pass the test. 

$wgSitename is unknown in this instance, you need to get the sitename from MirahezeFunctions, in the form of `$wi->sitename`.

Reception123 (approved):

PR #5432, opened 2024-01-10
T11040: Create $wmgMirahezeContactPageFooter

redbluegreenhat:
This variable is used for a hook that adds a link to the wiki's Special:Contact page to the footer.

While in ManageWiki it is set as a variable from the ContactPage extension, it is not. It's set like that so that it will only appear if ContactPage is enabled

As of writing, I'm not sure of the naming convention or where the hook should be, so I set it to wmgMirahezeContactPageFooter and added it to LocalSettings.php, respectively. I don't see any examples to go off by anyway.

### DIFF
--- a/GlobalSettings.php
+++ b/GlobalSettings.php
@@ -17,8 +17,8 @@ if ( $wmgMirahezeContactPageFooter && $wi->isExtensionActive( 'ContactPage' ) ) 
 				],
 			$skin->msg( 'contactpage-label' )->text()
 			);
-		}
-	}
+		};
+	};
 }
 
 // Extensions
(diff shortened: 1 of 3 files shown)


OAuthority:
It's probably best to wait for someone else to respond, cc @Universal-Omega, @Agent-Isai, but I don't think we generally put hooks themselves into LocalSettings.php. I think its prefered if we add them to MirahezeMagic and hook them in on load of GlobalSettings.php, see [here](https://github.com/miraheze/mw-config/blob/master/GlobalSettings.php#L3-L7)—but I would wait for consensus from UO or Agent to confirm.

redbluegreenhat:
@Universal-Omega want me to set the variable on MirahezeMagic's extension.json and move the hook to MirahezeMagic or just copypaste the hook to GlobalSettings.php?

redbluegreenhat:
done.

PR #5453, opened 2024-01-31
Create RunJobs.php

paladox:
From https://github.com/wikimedia/operations-mediawiki-config/blob/7c9f6a561b2b4b5c5db063bad83bd23e9cbac347/rpc/RunJobs.php


RhinosF1:
Make sure this file isn't accessible to the public

PR #5455, opened 2024-02-03
T10615: Add SemanticScribunto to ManageWiki

redbluegreenhat:


ManageWikiExtensions.php
```diff
@@ -3044,6 +3044,17 @@
 		],
 		'section' => 'other',
 	],
+	'semanticscribunto' => [
+		'name' => 'Semantic Scribunto',
```
Universal-Omega, line 3048: This needs to be SemanticScribunto (no space) or it wont load (needs to match exactly with 'name' in extension.json)

Universal-Omega (changes requested):

PR #5461, opened 2024-02-05
Upgrade all wikis to MediaWiki 1.41

AgentIsai:
To be applied once SQL patches are done applying

### DIFF
--- a/initialise/MirahezeFunctions.php
+++ b/initialise/MirahezeFunctions.php
@@ -58,9 +58,9 @@ class MirahezeFunctions {
 	];
 
 	public const MEDIAWIKI_VERSIONS = [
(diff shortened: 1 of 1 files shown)


initialise/MirahezeFunctions.php
```diff
@@ -58,9 +58,9 @@ class MirahezeFunctions {
 	];
 
 	public const MEDIAWIKI_VERSIONS = [
-		'alpha' => '1.42',
-		'beta' => '1.41',
-		'stable' => '1.40',
+		'alpha' => '1.43',
```
Universal-Omega, line 61: alpha is still 1.42, both beta and alpha can be 1.42 for now.

PR #5478, opened 2024-02-12
Support changing article path and using $wgMainPageIsDomainRoot

Universal-Omega:


initialise/MirahezeFunctions.php
```diff
@@ -1035,6 +1036,30 @@ public static function onManageWikiCoreAddFormFields( $ceMW, $context, $dbName,
...
+			'disabled' => !$permissionManager->userHasRight( $context->getUser(), 'managewiki-restricted' ),
+			'cssclass' => 'managewiki-infuse',
+			'section' => 'main',
+		];
+
+		$formDescriptor['mainpage-is-domain-root'] = [
+			'label-message' => 'miraheze-label-managewiki-mainpage-is-domain-root',
+			'type' => 'check',
+			'default' => $setList['wgMainPageIsDomainRoot'] ?? false,
+			'disabled' => !$permissionManager->userHasRight( $context->getUser(), 'managewiki-restricted' ),
```
songnguxyz, line 1058: If we wants it to work like WF, remove this.

initialise/MirahezeFunctions.php
```diff
@@ -1035,6 +1036,30 @@ public static function onManageWikiCoreAddFormFields( $ceMW, $context, $dbName,
...
+		$setList = $mwSettings->list();
+		$formDescriptor['article-path'] = [
+			'label-message' => 'miraheze-label-managewiki-article-path',
+			'type' => 'select',
+			'options-messages' => [
+				'miraheze-label-managewiki-article-path-wiki' => '/wiki/$1',
+				'miraheze-label-managewiki-article-path-root' => '/$1',
+			],
+			'default' => $setList['wgArticlePath'] ?? '/wiki/$1',
+			'disabled' => !$permissionManager->userHasRight( $context->getUser(), 'managewiki-restricted' ),
```
songnguxyz, line 1049: And this, see below...

songnguxyz (commented):
May I suggest to make this work just as WikiForge does (see [here](https://canary.discord.com/channels/407504499280707585/1006789349498699827/1208484292200763442)), remove these lines.

initialise/MirahezeFunctions.php
songnguxyz, line 1049: This,

initialise/MirahezeFunctions.php
songnguxyz, line 1058: and this...

initialise/MirahezeFunctions.php
Universal-Omega: This will be restricted, there are issues with cache that sometimes lead to redirect loops when changed so they will have to remain restricted for the time being until some are done to test if it will cause problems.

initialise/MirahezeFunctions.php
Universal-Omega: Same as above

PR #5498, opened 2024-03-11
Adding semantic drilldown

Disgustedorite:
Adding semantic drilldown. Semantic drilldown is already in mediawiki-repos.yaml, so I assume that no more work is required to add it other than putting it in the managewikiextensions thing.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -1701,6 +1701,17 @@ $wgManageWikiExtensions = [
 		],
 		'section' => 'specialpages',
 	],
+	'semanticdrilldown' => [
+        	'name' => 'Semantic Drilldown',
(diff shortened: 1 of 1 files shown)


ManageWikiExtensions.php
```diff
@@ -1701,6 +1701,17 @@
 		],
 		'section' => 'specialpages',
 	],
+	'semanticdrilldown' => [
+        	'name' => 'Semantic Drilldown',
```
Universal-Omega, line 1705:
```suggestion
        	'name' => 'SemanticDrilldown',
```
Needs to match name in extension.json

Universal-Omega (changes requested):

Disgustedorite:
uh, I committed the requested change and it's still saying changes are requested, what do I do here...?

PR #5504, opened 2024-03-20
DNM: grant everyone centralauth-merge

redbluegreenhat:
Possible fix to an issue where an user's global and local accounts in a wiki have been disconnected


RhinosF1:
@redbluegreenhat: which account? It's probably easier to fix one account as a one off by CLI

redbluegreenhat:
not needed anymore.

PR #5511, opened 2024-03-24
Adding function for UnlinkedWikibase

songnguxyz:
Allow the uses of Statements parser.


Universal-Omega (changes requested):
Needs added to LocalSettings.php also

Universal-Omega (approved):

PR #5512, opened 2024-03-24
Setup wgPrivilegedGroups

redbluegreenhat:
This is a feature introduced in MW 1.41

https://www.mediawiki.org/wiki/Manual:$wgPrivilegedGroups


Universal-Omega:
For the record this does absolutely nothing at all.

PR #5521, opened 2024-03-29
Add "Special:" in more languages

colleirose:


clarfonthey:
MediaWiki already emits `noindex,nofollow` meta tags for these pages; do we really need them in `robots.txt`?

colleirose:
> MediaWiki already emits `noindex,nofollow` meta tags for these pages; do we really need them in `robots.txt`?

I believe that blocking indexing of these pages can reduce server load. Aside from that, Google sometimes ignores noindex tags, but does not ignore robots.txt

PR #5527, opened 2024-04-04
Change the license image to be an svg

colleirose:
https://kinsta.com/blog/svg-vs-png/

Also the svg on MH Commons has a way smaller file size than the svg on Wikimedia Commons so that's why I'm linking to MH Commons instead


clarfonthey:
Seems pretty reasonable to me. Had no idea that this was overridden on MH Commons from the WM Commons version.

PR #5529, opened 2024-04-05
Allow using orain.org and wiki.surf also

Universal-Omega:


LocalSettings.php
```diff
@@ -4710,7 +4710,7 @@
 		'+bigforestwiki' => [
 			'editvoter',
 		],
-	        '+brandonwmwiki' => [
+			'+brandonwmwiki' => [
```
clarfonthey, line 4713: Probably should be fixed to a single indent

Universal-Omega: Thanks; fixed

RhinosF1 (changes requested):
This will need configuring on the cf side too.

RhinosF1:
@AgentIsai: can you chase cf on our open tickets?

PR #5541, opened 2024-04-18
Add configs for spacewiki

anpang54:
It turns core namespaces names lowercase. I tested this on a test wiki and it seems to work.


Universal-Omega (changes requested):
Should use single quotes, should have each entry (including the talk on a new line by itself, with only one space before the =>, and should use array format of `[]`

Universal-Omega (approved):

PR #5546, opened 2024-04-29
Set "$wgTabberNeueParseTabName" to "true" for "kanjouwiki"

spiriiii:
https://issue-tracker.miraheze.org/T12085


Universal-Omega:
This could really be done in LocalSettings.php or ideally be added to ManageWiki (if it isn't already?) tbh

PR #5556, opened 2024-05-09
Move StructuredDiscussions (Flow) to managewiki-restricted

labster:
Flow is a fairly problematic extension, and WMF is officially dropping support for it.

https://www.mediawiki.org/wiki/Structured_Discussions/Deprecation

Make the extension restricted, as we really shouldn't be moving new wikis into Flow.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -2940,7 +2940,12 @@ $wgManageWikiExtensions = [
 		'displayname' => 'StructuredDiscussions (Flow)',
 		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:StructuredDiscussions',
 		'conflicts' => false,
-		'requires' => [],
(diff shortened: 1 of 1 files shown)


Universal-Omega:
I think this should be completely grayed out (not just restricted) but only for enable not disable. I'm not sure how but that would be my preference.

labster:
Disable means losing access to content I think?  Just allowing straight disable is problematic, unless no namespaces currently have Flow enabled.

https://phabricator.wikimedia.org/T96301

I think the thing to ask is if this makes the situation better.  Does your option have an easy way to implement?  It looks to me that all of the other disabled extensions are disabled by making it restricted.

Universal-Omega:
Sorry I wasn't clear yes this is the correct way to do it as it stands but I think we should have a better way to disable extensions

Universal-Omega (approved):

PR #5557, opened 2024-05-13
T12108: Please alter these DynamicPageList3 variables on metzo.miraheze.org

brandon-wm:


Universal-Omega:
I just realized this used spaces not tabs please use tabs next time. I already fixed it for this one though.

PR #5558, opened 2024-05-13
T12085: About “Extension:TabberNeue”

brandon-wm:


Universal-Omega (changes requested):
This should be done in LocalSettings.php or ideally added to ManageWiki

PR #5559, opened 2024-05-13
T11757: set some UploadWizard config for aieseattlewiki

brandon-wm:


Universal-Omega (changes requested):
This should not be merged yet. It has some conflicting stuff with default that may need sorted.

brandon-wm:
@Universal-Omega Is there stuff that I can help fix with defaults or is it stuff that you need to sort yourself?

Universal-Omega (changes requested):
This needs a lot of tabbing fixed

PR #5560, opened 2024-05-14
T12085: About “Extension:TabberNeue”

brandon-wm:


Universal-Omega:
IMO this should actually a ManageWiki setting

PR #5562, opened 2024-05-15
Add PageLanguage and PerPageLanguage to MW extensions

brandon-wm:


clarfonthey:
Definitely looks fine for PageLanguage (WMF supported) but my only concern with PerPageLanguage is that we prefer updating to the latest MW version when it comes out, and they might lag behind in support.

ManageWikiExtensions.php
```diff
@@ -2589,9 +2589,22 @@
...
+		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:PageLanguage',
+		'conflicts' => false,
+		'requires' => [],
+		'section' => 'other',
+	],
+	'perpagelanguage' => [
+		'name' => 'PerPageLanguage',
+		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:PerPageLanguage',
+		'conflicts' => false,
+		'requires' => [],
```
Universal-Omega, line 2603:
```suggestion
		'requires' => [],
		'install' => [
			'settings' => [
				'wgPageLanguageUseDB' => true,
			],
		],
```

Universal-Omega (changes requested):

ManageWikiExtensions.php
```diff
@@ -2589,9 +2589,27 @@
 		'requires' => [],
 		'section' => 'other',
 	],
+	'pagelanguage' => [
+		'name' => 'PageLanguage',
```
Universal-Omega, line 2593:
```suggestion
		'name' => 'Page Language',
```

Universal-Omega (changes requested):
'name' must match the name in extension.json

Universal-Omega (approved):

PR #5566, opened 2024-05-19
Changes to how special pages are treated

colleirose:
See the reasoning provided in the commit


clarfonthey:
Hmm, was the reasoning in #5521 incorrect, then?

colleirose:
> Hmm, was the reasoning in #5521 incorrect, then?

Indeed, I was mistaken. I learnt something new today while researching something else :-)

PR #5580, opened 2024-05-28
Adding new footer icons (StarCitizenTools style) to lhmnwiki and sister projects

songnguxyz:
Adding new footer icons (StarCitizenTools style)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1585,11 +1585,11 @@ $wgConf->settings += [
 		'hsckwiki' => [
 			'poweredby' => [
 				'mediawiki' => [
-					"src" => "https://static.miraheze.org/lhmnwiki/9/9b/MediaWiki.svg",
-					"url" => "https://www.mediawiki.org",
-					"alt" => "Xây dựng trên MediaWiki",
-					"height" => "42",
-					"width" => "127",
+					'src' => 'https://static.miraheze.org/lhmnwiki/9/9b/MediaWiki.svg',
+					'url' => 'https://www.mediawiki.org',
+					'alt' => 'Xây dựng trên MediaWiki',
+					'height' => "42",
+					'width' => "127",
 				],
 				'songnguxyz' => [
 					'src' => 'https://static.miraheze.org/lhmnwiki/5/58/Footer.SN.xyz.svg',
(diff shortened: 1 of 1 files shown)


clarfonthey:
Is there not a setting to modify these under `ManageWiki`? It makes sense if people are using different images to match a certain style; either we should have a policy of these images remaining the same, or letting people modify them on the wikis themselves.

songnguxyz:
> Is there not a setting to modify these under `ManageWiki`? It makes sense if people are using different images to match a certain style; either we should have a policy of these images remaining the same or letting people modify them on the wikis themselves.

I'd love to if that would come true, but the support for it hasn't been implemented yet.

RhinosF1:
> > Is there not a setting to modify these under `ManageWiki`? It makes sense if people are using different images to match a certain style; either we should have a policy of these images remaining the same or letting people modify them on the wikis themselves.
> 
> I'd love to if that would come true, but the support for it hasn't been implemented yet.

Is there any reason you couldn't make it come true?

clarfonthey:
I should have also clarified, asking why this isn't a setting wasn't necessarily directed at you (songnguxyz), but mostly the folks viewing it in general. I think it would be lovely to implement if it's desired.

songnguxyz:
Accidentally closed.

songnguxyz:
Oh come on.

LocalSettings.php
```diff
@@ -1588,11 +1588,31 @@
 					'src' => 'https://static.miraheze.org/lhmnwiki/5/58/Footer.SN.xyz.svg',
 					'url' => 'https://songngu.xyz',
 					'alt' => 'Dự án được bảo quản bởi SongNgư.xyz',
+					"height" => "32",
+				        "width" => "200"
 				],
 				'miraheze' => [
-					'src' => 'https://static.miraheze.org/commonswiki/f/ff/Powered_by_Miraheze.svg',
+					'src' => 'https://static.miraheze.org/lhmnwiki/1/1c/Miraheze.svg',
```
Reception123, line 1595: is there any reason for changing the "powered by" here?

songnguxyz: I wanted to give the MH icon the personality of my wiki while still giving the hosting platform the credit it deserved. There should be no infringement at all as it still fully shows that MH have a large part of making my wiki possible.

Reception123 (approved):

PR #5597, opened 2024-06-30
T12272: Rename global-sysop to global-admin

waki285:


OAuthority:
Please amend the description of the commit detailing why you are making this change for clarity.

PR #5598, opened 2024-06-30
T12214: Avoid redirects if action=raw is set

BlankEclair:
(This PR is untested)

### DIFF
--- a/initialise/entrypoints/index.php
+++ b/initialise/entrypoints/index.php
@@ -11,8 +11,8 @@ define( 'MW_ENTRY_POINT', 'index' );
 require_once '/srv/mediawiki/config/initialise/MirahezeFunctions.php';
 require MirahezeFunctions::getMediaWiki( 'includes/WebStart.php' );
 
-// T12214: Avoid redirects if there are query parameters
-if ( $wgRequest->getValues() === [] ) {
+// T12214: Avoid redirects if action=raw is set
+if ( RequestContext::getMain()->getActionName() !== "raw" ) {
 	if ( $wgArticlePath === '/$1' && str_contains( strtoupper( $_SERVER['REQUEST_URI'] ), strtoupper( '/wiki/' ) ) ) {
 		// Redirect to the same page maintaining the path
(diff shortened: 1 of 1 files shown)


Universal-Omega (changes requested):
A few things will this.

1) We can never use wgRequest. That is deprecated and should use RequestContext instead
2) This PR kinda defeats the whole purpose of this normalizing which is aimed at normalizing all URLs to one format. For SEO and other purposes, and to stop duplicate things getting indexed which hurts SEO  and that happens with ?title=, I think the better way here would be to not redirect if we are using action=raw or something. However even that I am not sure is the best solution, but it may be the best for this particular issue.

initialise/entrypoints/index.php
```diff
@@ -11,14 +11,17 @@
...
-if ( $wgArticlePath === '/$1' && str_contains( strtoupper( $_SERVER['REQUEST_URI'] ), strtoupper( '/wiki/' ) ) ) {
-	// Redirect to the same page maintaining the path
-	header( 'Location: ' . str_replace( '/wiki/', '/', $_SERVER['REQUEST_URI'] ), true, 301 );
-	exit();
-} elseif ( $wgArticlePath === '/wiki/$1' && !str_contains( $_SERVER['REQUEST_URI'], '/wiki/' ) && !str_contains( $_SERVER['REQUEST_URI'], '/w/' ) && !( $wgMainPageIsDomainRoot && $_SERVER['REQUEST_URI'] === '/' ) ) {
-	// Redirect to the same page maintaining the path
-	header( 'Location: /wiki' . $_SERVER['REQUEST_URI'], true, 301 );
-	exit();
+// T12214: Avoid redirects if action=raw is set
+if ( RequestContext::getMain()->getActionName() !== "raw" ) {
```
Universal-Omega, line 15: A bit minor but for consistency we should use single quotes for raw

BlankEclair: Oops, it's kinda instinct for me to use double quotes since I've used other programming languages more than PHP 😅

Universal-Omega (approved):

PR #5602, opened 2024-07-09
set $wgCdnServers when cloudflare not in use

The-Voidwalker:


redbluegreenhat:
FYI there are two custom domains currently using Cloudflare, wiki.themanaworld.org and fanon.polandballwiki.com.

The-Voidwalker:
Two domains shouldn't be a problem. Worst case should just be that they'll be generating superfluous purges to varnish.

PR #5606, opened 2024-07-13
$wgRSSAllowLinkTag: Escape HTML

BlankEclair:
Here's how it looks like right now:


Reception123 (approved):

Reception123:
We should have an escaping guide somewhere on Meta as even though I didn't write this particular one, escaping in different languages always confuses me.

PR #5633, opened 2024-08-09
T12447: Add GoogleForms

AverageHelper:
If I understand right, this and https://github.com/miraheze/mediawiki-repos/pull/30 together should add the GoogleForms extension to `Special:ManageWiki/extensions`.

### DIFF
--- a/ManageWikiExtensions.php
+++ b/ManageWikiExtensions.php
@@ -1418,11 +1418,11 @@ $wgManageWikiExtensions = [
 		'section' => 'specialpages',
 	],
 	'googleforms' => [
-		'name' => 'CustomSearchProfiles',
+		'name' => 'GoogleForms',
 		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:GoogleForms',
 		'conflicts' => false,
 		'requires' => [],
-		'section' => 'other',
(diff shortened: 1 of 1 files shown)


ManageWikiExtensions.php
```diff
@@ -1417,6 +1417,13 @@
 		],
 		'section' => 'specialpages',
 	],
+	'googleforms' => [
+		'name' => 'CustomSearchProfiles',
```
BlankEclair, line 1421: Probably should be GoogleForms

ManageWikiExtensions.php
```diff
@@ -1417,6 +1417,13 @@
...
+		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:GoogleForms',
+		'conflicts' => false,
+		'requires' => [],
+		'section' => 'other',
```
BlankEclair, line 1425: Probably better goes in "parserhooks"

BlankEclair (changes requested):

ManageWikiExtensions.php
AverageHelper: Whoops 😅

BlankEclair (approved):

Reception123 (approved):

PR #5641, opened 2024-08-21
Add UploadWizard configurations for kagagawiki in LocalWiki.php

hurohukidaikon:
Hello. This is my first PR in this project, so if I had any mistake please let me know them gently :-)

- Add some licenses we needed (2 Rights Statements options).
- Add Japanese in language setting.

Our wiki is https://kagaga.jp/.

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -199,6 +199,8 @@ switch ( $wi->dbname ) {
 		break;
 	case 'kagagawiki':
 		$wgUploadWizardConfig = [
+			'campaignExpensiveStatsEnabled' => false,
+			'flickrApiKey' => $wmgUploadWizardFlickrApiKey,
 			'uwLanguages' => [
 				'ja' => '日本語',
 				'en' => 'English',
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -204,7 +204,7 @@ switch ( $wi->dbname ) {
 			'uwLanguages' => [
 				'ja' => '日本語',
 				'en' => 'English',
-  			],
+			],
 			'licenses' => [
 				'cc-by-sa-4.0' => [
 					'msg' => 'mwe-upwiz-license-cc-by-sa-4.0',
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -196,6 +196,67 @@ function onBeforePageDisplay( OutputPage $outputPage ) {
 		$wgSpecialPages['Analytics'] = DisabledSpecialPage::getCallback( 'Analytics', 'MatomoAnalytics-disabled' );
 		$wgPageImagesScores['position'] = [ 100, -100, -100, -100 ];
 
+		break;
+	case 'kagagawiki':
+		$wgUploadWizardConfig = [
+			'uwLanguages' => [
```
Universal-Omega, line 202: Should also probably include the default value we have set here also:
```suggestion
			'campaignExpensiveStatsEnabled' => false,
			'flickrApiKey' => $wmgUploadWizardFlickrApiKey,
			'uwLanguages' => [
```

Universal-Omega (changes requested):

LocalWiki.php
hurohukidaikon: Thank you for your reply.

> `'campaignExpensiveStatsEnabled' => false,`

I don't know what is the value for.
Is this for decrease computer resources?

https://github.com/wikimedia/mediawiki-extensions-UploadWizard/blob/master/UploadWizard.config.php#L109C2-L110C42

> `'flickrApiKey' => $wmgUploadWizardFlickrApiKey,`

We'er not using Flickr, still we should set the API key?

Universal-Omega: > Thank you for your reply.
> 
> > `'campaignExpensiveStatsEnabled' => false,`
> 
> I don't know what is the value for. Is this for decrease computer resources?

Yep, this should always be disabled no matter what
> https://github.com/wikimedia/mediawiki-extensions-UploadWizard/blob/master/UploadWizard.config.php#L109C2-L110C42
> 
> > `'flickrApiKey' => $wmgUploadWizardFlickrApiKey,`
> 
> We'er not using Flickr, still we should set the API key?

It may cause warnings on our side if this isn't set while unlikely it will cause issues it is possible so please do set it anyway.

hurohukidaikon: Okay, I added both.

Universal-Omega (approved):

PR #5649, opened 2024-09-02
T12526: Add redundent langs to DisabledLanguages on meta

pixDeVl:


Universal-Omega:
What is the reason this is a draft? Is it blocked on something?

pixDeVl:
The task was stalled but I can’t recall what right now lmao. Maybe deleting the rest of the translations? Let me see

clarfonthey:
I believe that this only disables adding new translations, and you could in theory merge this and delete the translations later. Would have to verify in the code, though.

Either way, since the translate extension actually writes translations to pages instead of just caching them, I'm 90% sure it should be fine to merge first and delete later.

winstonsung:
@AgentIsai :

There are still remaining CNBanner pages not migrated.

Screenshot

winstonsung:
Remaining migration done, could re-apply:

* <https://meta.miraheze.org/wiki/Special:ActiveLanguages/zh>
* <https://meta.miraheze.org/wiki/Special:ActiveLanguages/zh-cn>
* <https://meta.miraheze.org/wiki/Special:ActiveLanguages/zh-tw>

PR #5657, opened 2024-09-10
T12565: Fully disable transcodes for hommwiki

BlankEclair:
$wgEnableTranscode should, in theory, disable transcoding. However, there is (probably) a bug upstream that makes it not fully honor that setting. For now, we'll just do a workaround to make the extension chill out, and perhaps revert this commit once the extension is fixed.

Miraheze issue tracker task: https://issue-tracker.miraheze.org/T12565
Relevant upstream task: https://phabricator.wikimedia.org/T374433

### DIFF
--- a/LocalWiki.php
+++ b/LocalWiki.php
@@ -192,6 +192,11 @@ switch ( $wi->dbname ) {
 		];
 
 		break;
+	case 'hommwiki':
+		$wgEnabledAudioTranscodeSet = [];
+		$wgEnabledTranscodeSet = [];
+
+		break;
 	case 'houkai2ndwiki':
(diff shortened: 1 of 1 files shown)


LocalWiki.php
```diff
@@ -191,6 +191,11 @@ function onBeforePageDisplay( OutputPage $outputPage ) {
 			'url' => 'https://gpcommons.miraheze.org/w/api.php'
 		];
 
+		break;
+	case 'hommwiki':
+		$wgEnabledAudioTranscodeSet = [];
+		$wgEnabledTranscodeSet = [];
```
Universal-Omega, line 197: This doesn't seem to be necessary? It seems just $wgEnabledAudioTranscodeSet along side $wgEnableTranscode (which is available in ManageWiki)

BlankEclair: I did it just in case, but I suppose I can remove it.

PR #5678, opened 2024-09-24
Support new event/jobrunner for beta

paladox:
We want to be able to use it for beta in addition to production.


RhinosF1:
@paladox: status?

PR #5686, opened 2024-09-25
set wgUploadNavigationUrl for Abyssal Wiki

silentgamer568:
I hope I did this right.


Universal-Omega:
This is available in Special:ManageWiki/settings

PR #5711, opened 2024-10-11
Create public/private wiki list

AgentIsai:
Useful for wiki backups

### DIFF
--- a/initialise/MirahezeFunctions.php
+++ b/initialise/MirahezeFunctions.php
@@ -963,6 +963,7 @@ class MirahezeFunctions {
 				 'wiki_deleted',
 				 'wiki_closed',
 				 'wiki_inactive',
+				 'wiki_private',
 			] )
 			->caller( __METHOD__ )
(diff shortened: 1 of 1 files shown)


Universal-Omega:
What is the reason here? More DB lists = worse performance so if there isn't an absolute necessity they shouldn't be done.

AgentIsai:
> What is the reason here? More DB lists = worse performance so if there isn't an absolute necessity they shouldn't be done.

Really? Little things like this should be documented. Otherwise, I thought it was because no one had thought about it.

Universal-Omega:
Hmm maybe I fixed the worse performance with more lists now lol

Universal-Omega:
They just don't update unless databases.php until is updated.

AgentIsai:
Should we see if any performance issues occur?

Universal-Omega:
Yeah we can try this. I actually do believe I fixed it with my recent changes.

AgentIsai:
Very well.

PR #5719, opened 2024-10-19
Add db172

redbluegreenhat:


Universal-Omega:
This should be c1 — on beta with two seperate setups entirely.

redbluegreenhat:
> This should be c1 — on beta with two seperate setups entirely. It reduces the chance of accidentally using it in production if something goes wrong.

Like define two separate `$wgLBFactoryConf` depending on the hostname?

Universal-Omega:
> > This should be c1 — on beta with two seperate setups entirely. It reduces the chance of accidentally using it in production if something goes wrong.
> 
> Like define two separate `$wgLBFactoryConf` depending on the hostname?

This is what I was thinking, as that way the clusters only exist in the correct realm and not exist at all otherwise. Maybe even a seperate Databases-beta.php or something.

Universal-Omega:
I can't review this in depth right now but looks okay on a glance. Feel free to do it if you feel like it is correct now.

PR #5722, opened 2024-10-31
Open interwiki links in new tab

robkam:
Get interwiki links to open in a new tab for the Popular Music Studies Wiki.


robkam:
The interwiki links already open in a new tab, instead of the same tab?

PR #5739, opened 2024-11-18
Switch virtual database domain settings

AgentIsai:


Universal-Omega:
This is going to break things

Universal-Omega:
Like CentralAuth since it doesn't have a virtual domain support until 1.43

PR #5763, opened 2024-12-08
T12963: Set allowUnlimitedCategories to true for gui7814sgtafanonwiki

BlankEclair:
Closes T12963


BlankEclair:
https://issue-tracker.miraheze.org/T12963#260556

PR #5768, opened 2024-12-11
Remove 2880 as a candidate MultimediaViewer thumbnail size for Strinova Wiki

lihaohong6:
Address T12977. Not sure if the issue will be completely fixed, but this config seems to be the culprit.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4333,6 +4333,25 @@ $wgConf->settings += [
 	'wgMediaViewerEnableByDefault' => [
 		'default' => true,
 	],
+	'wgMediaViewerThumbnailBucketSizes' => [
+		'default' => [
+			320,
+			800,
+			1024,
+			1280,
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -4333,6 +4333,25 @@
...
 	],
+	'wgMediaViewerThumbnailBucketSizes' => [
+		'default' => [
+			320,
+			800,
+			1024,
+			1280,
+			1920,
+			2560,
+			2880
```
BlankEclair, line 4344: Slight nitpick: We usually have trailing commas at the end of (multiline) arrays :p

lihaohong6: This has been addressed, though your update on Phorge seems to suggest that the root cause is some other configuration problem, so this change may not be necessary.

Universal-Omega (approved):

PR #5780, opened 2025-01-12
Set smwgConfigFileDir to /srv/mediawiki/

paladox:
Right now it locates in /srv/mediawiki/\<version\>/extensions/SemanticMediaWiki/.


Universal-Omega:
I believe I enabled a config that makes this file not even used, so doesn't really matter where it is.

PR #5811, opened 2025-02-09
Adding RobloxURI to cafewiki allowed protocols - T13215

enews:
Adding RobloxURI to wgUrlProtocols as per T13215


enews:
Added Urbanshade and UTG/UFG as their request

PR #5831, opened 2025-02-19
Un-restrict CirrusSearch in ManageWiki

Reception123:
Leaving this open for a few days to make sure there's no objections.

With resetWikiCaches.php being added, there is no more need for tech intervention when enabling CirrusSearch.

@Agent-Isai mentioned that os162 is only at 11% storage which means there aren't any concerns anymore about more users enabling CS as a result.

Because of this I see no need to keep it restricted.


Universal-Omega (changes requested):
This can not be merged yet for 3 reasons:
* It needs to handle restoring native search functionality when extension is disabled.
* We first need to launch another OpenSearch server for MediaWiki.
* We need to consider if we should just make CirrusSearch global or a default extension etc...

Fluffy728 (changes requested):
It’s still broken I think…

Reception123:
The community supported keeping it in ManageWiki (not as default) for now in https://meta.miraheze.org/wiki/Community_portal/Archive_53#Request_for_Feedback:_CirrusSearch

Regarding disabling, I think it would be quite rare but it would probably be easier to just restrict disabling only (if possible). Or even just a help message warning users not to restrict.

Reception123:
After discussing with CA, we will likely be making CS a global extension instead.

PR #5841, opened 2025-03-02
Set default `$wgKartographerSrcsetScales` to empty array

tjhorner:
Since the default tile server is `tile.openstreetmap.org` and that server only supports 1x tiles, it makes sense for the default `$wgKartographerSrcsetScales` to be an empty array which, according to the Kartographer docs, will implicitly include `1`. There shouldn't be overrides for individual wikis unless they use a different tile server which supports tile scaling.

> Doesn't need to start with 1, this happens automatically. Note that most tile servers don't support this at all and thus require setting this to an empty array.

— [Kartographer docs](https://www.mediawiki.org/wiki/Extension:Kartographer#Configuration)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2776,25 +2776,7 @@ $wgConf->settings += [
 		'default' => 'https://tile.openstreetmap.org',
 	],
 	'wgKartographerSrcsetScales' => [
-		'default' => [
-			1.3,
-			1.5,
-			2,
-			2.6,
-			3,
-		],
-		'bluepageswiki' => [
-			1,
-		],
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2776,25 +2776,7 @@
...
-		'hkrailwiki' => [
-			1,
-		],
-		'isvwiki' => [
-			1,
-		],
-		'leborkwiki' => [
-			1,
-		],
+		'default' => [ ],
```
BlankEclair, line 2779:
```suggestion
		'default' => [],
```

BlankEclair (approved):

PR #5844, opened 2025-03-03
T13275: Add Data.JsonConfig to ftlmultiversewiki

BlankEclair:
Fixes T13275


BlankEclair:
Not self-merging for the next few hours because I won't be available for then in case something breaks

PR #5860, opened 2025-03-11
Adding entries for Extension:EditSimilar to ManageWikiSettings.php

Aeywoo:
Adding Extension:EditSimilar from [T13286](https://issue-tracker.miraheze.org/T13286).


ManageWikiSettings.php
```diff
@@ -709,6 +709,42 @@
 	],
 
 	// Editing
+	'$wgEditSimilarMaxResultsPool' => [
+		'name' => 'EditSimilar Max Results Pool',
+		'from' => 'editsimilar',
+		'type' => 'integer',
+		'overridedefault' => true,
```
BlankEclair, line 716: > overridedefault: a string/array override default when no existing value exist.

This is meant to be what the default value is

Aeywoo: So, I don't need to define that?

BlankEclair: I believe you have to define it, but as the value you set as the default in LocalSettings.php

ManageWikiSettings.php
```diff
@@ -709,6 +709,48 @@
...
+		'overridedefault' => 1,
+		'section' => 'editing',
+		'help' => 'This number specifies per which number of edits will show the results\' message, this means every time results are found a message is shown.',
+		'requires' => [],
+	],
+	'wgEditSimilarAlwaysShowThanks' => [
+		'name' => 'EditSimilar Always Show Thanks',
+		'from' => 'editsimilar',
+		'type' => 'check',
+		'overridedefault' => true,
```
BlankEclair, line 749:
```suggestion
		'overridedefault' => false,
```

PR #5878, opened 2025-03-20
Don't set $wgCentralAuthCookieDomain

Universal-Omega:


GlobalSettings.php
```diff
@@ -276,11 +276,9 @@
 // Dynamic cookie settings dependant on $wgServer
```
BlankEclair, line 276:
```suggestion
// Dynamic cookie settings dependent on $wgServer
```

BlankEclair (commented):
Why though?

Universal-Omega:
> Why though?

Because I don't get logged out when visiting new custom domains, I do when visiting new miraheze.org subdomains. So wanted to see what happens removing this which is how custom domains are configured right now.

BlankEclair (approved):
Fair enough ^_^

(Technically, there's still the typo I reported, but I'm not gonna harp on that)

Fluffy728 (approved):

PR #5894, opened 2025-04-03
Updating Cloudflare's IPs.

Aeywoo:
Updated the Cloudflare IPs from Cloudflare's IP page in $wgCdnServersNoPurge.


OAuthority:
LGTM. Please can you keep the IPs in the same order as listed on the Cloudflare IP page so that it is easier for us to check the PR in future against the list on Cloudflare's website. I have amended this PR to reflect this :).

Aeywoo:
It is in format to the Cloudflare IP Files.

OAuthority:
I'm slightly confused -- looking again at this, there seems to be no difference between the original and your change, other than you moved the IPs around?

PR #5895, opened 2025-04-04
Re-enabling Skin:Evelution

Aeywoo:
[Skin:Evelution](https://www.mediawiki.org/wiki/Skin:Evelution) is now compatible with MW REL1_42 and above, see the appropriate GitHub repository: https://github.com/AWikia/SkinEvelution

Recent changes to the extension have rendered the configuration option `$wgEvelutionServerMode` deprecated, it will need to be removed from all wiki's configuration settings before this PR is merged.


Universal-Omega:
IMO, this skin should never be reenabled. It should be removed instead. It's to horrible performance wise. I'll let @BlankEclair decide though if wants to review it again or not.

PR #5932, opened 2025-04-30
T13612: Add settings for NewUserMessage

SomeMWDev:
Adds all three settings ([docs](https://www.mediawiki.org/wiki/Extension:NewUserMessage#Configuration)) of the extension to ManageWiki which resolves https://issue-tracker.miraheze.org/T13612 (`$wgNewUserMessageOnAutoCreate` was requested, I added the other two ones as well since they weren't in ManageWiki either).

`$wgNewUserMessageOnAutoCreate` will have to be set to true for `nmfwikiwiki` via ManageWiki when this is merged since this was specified in LocalSettings before this change.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4350,9 +4350,14 @@ $wgConf->settings += [
 	],
 
 	// NewUserMessage configs
+	'wgNewUserSuppressRC' => [
+		'default' => false,
+	],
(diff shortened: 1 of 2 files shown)


Aeywoo (approved):
Looks good to me, but we need to make sure the nmfwiki has the $wgNewUserMessageOnAutoCreate setting set to true before merging this.

PR #5938, opened 2025-05-03
Enable followRedirects for ExternalData

BlankEclair:


BlankEclair:
@RhinosF1 gave the +2 in Mattermost (I sent it there since I wasn't 100% sure if everyone else would count the pre-existing SSRF as a security issue). He'll send his two cents here once he manages to escape from the enrapturing B.E.D.

RhinosF1 (commented):
To summarise my thoughts given on Mattermost, we have no limits on where External Data can link to so allowing followRedirects doesn't change the threat model. It makes some threats slightly harder to detect.

hence, this is a low risk PR.

PR #5985, opened 2025-06-14
T13842: Enable $wgTemplateStylesExtenderEnableUnscopingSupport

asko1:
Resolves [T13842](https://issue-tracker.miraheze.org/T13842).

$wgTemplateStylesExtenderEnableUnscopingSupport enables support for unscoping TemplateStyles CSS from the default `mw-parser-output` class. Unscoping, by default, would require the `editinterface` permission, so this would not cause a security risk for the wiki.
[Extension:TemplateStylesExtender#Configuration](https://www.mediawiki.org/wiki/Extension:TemplateStylesExtender#Configuration)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5817,6 +5817,12 @@ $wgConf->settings += [
 		],
 	],
 
+	// TemplateStylesExtender
+	'wgTemplateStylesExtenderEnableUnscopingSupport' => [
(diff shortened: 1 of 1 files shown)


SomeMWDev:
Closing as we cannot add that setting right now per https://issue-tracker.miraheze.org/T13842#277336, sorry

PR #5991, opened 2025-06-18
Add $wgExtractsRemoveClasses to ManageWikiSettings

pskyechology:
Also updates default values from upstream.
Resolves [T13859](https://issue-tracker.miraheze.org/T13859).

❗For maintainers❗
Two overrides are removed and have to be set in ManageWiki.

`'+gratispaideiawiki' => [
			'.metadata',
			'span.coordinates',
			'span.geo-multi-punct',
			'span.geo-nondefault',
			'#coordinates',
		],
`

`'softcellwiki' => [
			'table',
			'script',
			'input',
			'style',
			'ul.gallery',
			'.mw-editsection',
			'sup.reference',
			'ol.references',
			'.error',
			'.nomobile',
			'.noprint',
			'.noexcerpt',
			'.sortkey',
			'#toc',
		],`
`

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -458,12 +458,8 @@ $wgManageWikiSettings = [
 			'.mw-empty-elt',
 		],
 		'section' => 'other',
-		'help' => 'Elements which will be excluded from extraction.',
-		'requires' => [
-			'extensions' => [
-				'textextracts',
-			],
-		],
+		'help' => 'Selectors for elements which will be excluded from extraction.',
+		'requires' => [],
 	],
 	'wgRSSAllowLinkTag' => [
 		'name' => 'Allow links in RSS feeds',


ManageWikiSettings.php
```diff
@@ -433,6 +433,38 @@
...
+			'.noprint',
+			'.noexcerpt',
+			'.sortkey',
+			'.mw-empty-elt',
+		],
+		'section' => 'other',
+		'help' => 'Elements which will be excluded from extraction.',
+		'requires' => [
+			'extensions' => [
+				'textextracts',
```
SomeMWDev, line 464: This doesn't need to be specified here, since 'from' is set to 'textextracts' already.
See also: https://github.com/miraheze/mw-config/blob/d60adfe9a96375ae7bbf4527c1e030f8b5360b20/ManageWikiSettings.php#L46

ManageWikiSettings.php
```diff
@@ -433,6 +433,38 @@
...
+			'ol.references',
+			'.error',
+			'.nomobile',
+			'.noprint',
+			'.noexcerpt',
+			'.sortkey',
+			'.mw-empty-elt',
+		],
+		'section' => 'other',
+		'help' => 'Elements which will be excluded from extraction.',
```
SomeMWDev, line 461: Since this can also contain classes and IDs, maybe "Selectors for elements" instead of "Elements"?

ManageWikiSettings.php
asko1: Missed that, thanks! Used the setting below as an example, which makes the same mistake. Could do with a cleanup. Fixed by 9f405381611a2c8526dc5b8c9a9f9645365dd183

PR #5997, opened 2025-06-23
cgwiki: Protection level fix

songnguxyz:
Rename protection level and define order for cgwiki

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5230,7 +5230,11 @@ $wgConf->settings += [
 			'editconsulprotected',
 		],
 		'+cgwiki' => [
-			'editcgprotected',
+			'',
+			'user',
+			'autoconfirmed',
+			'cg',
+			'sysop',
 		],
 		'+cmgwiki' => [
 			'bureaucrat',
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3452,11 +3452,6 @@ $wgConf->settings += [
 				'read' => true,
 			],
 		],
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -5230,7 +5230,11 @@
 			'editconsulprotected',
 		],
 		'+cgwiki' => [
-			'editcgprotected',
+			'',
+			'user',
+			'autoconfirmed',
+			'cg',
```
SomeMWDev, line 5236: This should probably remain `editcgprotected`, otherwise you'll have to rename the right as well

Universal-Omega: Please make sure it's removed from any groups in ManageWiki first then. Also just to note *eventually* ManageWiki will support restriction levels and ones like these will not be supported. They will have to follow a certain naming convention for security. Likely edit*

Universal-Omega: yes because the protection level is rights thats what you enter in it. Any right that exists can be a protection level so to add custom protection levels you have to create new rights.

songnguxyz:
I would like to ask for updates on this, and are there anything else I have to fix?

SomeMWDev:
> I would like to ask for updates on this, and are there anything else I have to fix?

Per https://github.com/miraheze/mw-config/pull/5997/files#r2162951103, it would probably be better if the protection level followed the naming convention like others (so `editcgprotected`), otherwise it might have to be changed in the future

PR #6000, opened 2025-06-25
Update LocalSettings.php

Xaloria:


LocalSettings.php
```diff
@@ -5252,7 +5252,7 @@
 		'+cricketnepalwiki' => [
 			'editbureaucratprotected',
 			'editstaffprotected',
-			'edittemplateprotected',
+                        'editfounderprotected',
```
asko1, line 5255: https://github.com/miraheze/mw-config/blob/e970a824e6f42797748ab1f2bf4b08efe719bd0f/LocalSettings.php#L5390
You also have to change the associated group right here.
Make sure the old right `edittemplateprotected` one is removed from any groups in ManageWiki.

Additionally, I suggest you make a ticket on [Phorge](https://meta.miraheze.org/wiki/Phorge), in case there are any doubts about whether you are a bureaucrat for this wiki.

PR #6037, opened 2025-07-27
T14052: Add multi-level interwiki prefix for utgwiki

pskyechology:


pskyechology:
Tested at https://skye.mirabeta.org/wiki/Special:Interwiki and https://skye.mirabeta.org/wiki/Utginterwiki
Working as intended.

SomeMWDev (changes requested):
'tl' is the language code for Tagalog. This means that the interwiki prefix will be used as an interlanguage prefix:
image

The link will therefore appear in the language dropdown instead of the dropdown:
image

pskyechology:
darn, should've known that's too easy to be correct! figured it was stackd tomfoolery causing it to default to tagalog

SomeMWDev (approved):

PR #6112, opened 2025-09-17
T14285: add NS_MAIN to wgMFNamespacesWithoutCollapsibleSections for twistwoodtaleswiki

c-oreills:
As a potential workaround for the deprecated wgMFCollapseSectionsByDefault setting

PR #6113, opened 2025-09-17
T14285: Add wgMFNamespacesWithoutCollapsibleSections to ManageWikiNamespaces.php

c-oreills:
As a potential workaround for the deprecated wgMFCollapseSectionsByDefault setting

Supersedes https://github.com/miraheze/mw-config/pull/6112, suggested by Skye [here](https://issue-tracker.miraheze.org/T14285#286268)

### DIFF
--- a/ManageWikiNamespaces.php
+++ b/ManageWikiNamespaces.php
@@ -387,4 +387,22 @@ $wgManageWikiNamespacesAdditional = [
 		'help' => '',
 		'requires' => [],
 	],
+	'wgMFNamespacesWithoutCollapsibleSections' => [
+		'name' => 'Disable collapsible sections in this namespace?',
+		'from' => 'mobilefrontend',
+		'type' => 'check',
+		'main' => true,
+		'talk' => true,
+		'excluded' => [],
+		// See https://github.com/wikimedia/mediawiki-extensions-MobileFrontend?tab=readme-ov-file#wgmfnamespaceswithoutcollapsiblesections
+		'overridedefault' => [
+			NS_FILE => true,
+			NS_CATEGORY => true,
+			NS_SPECIAL => true,
+			NS_MEDIA => true,
+			'default' => false,
+		],
+		'help' => 'If enabled, collapsible sections will be disabled in this namespace on mobile devices.',
+		'requires' => [],
+	],
 ];


ManageWikiNamespaces.php
```diff
@@ -387,4 +387,22 @@
...
 	],
+	'wgMFNamespacesWithoutCollapsibleSections' => [
+		'name' => 'Disable collapsible sections in this namespace?',
+		'from' => 'mobilefrontend',
+		'type' => 'check',
+		'main' => true,
+		'talk' => true,
+		'excluded' => [],
+		// See https://github.com/wikimedia/mediawiki-extensions-MobileFrontend?tab=readme-ov-file#wgmfnamespaceswithoutcollapsiblesections
+		'overridedefault' => [
```
c-oreills, line 398: I'm unsure if (a) this is the right format or (b) it's good practice to override the default like this rather than just allow the `mobilefrontend` extension to set the default. Advice very much appreciated!

pskyechology (approved):
Will merge with #6112

LocalSettings.php
```diff
@@ -4389,6 +4389,15 @@
 	'wgDefaultMobileSkin' => [
 		'default' => 'minerva',
 	],
+	'wgMFNamespacesWithoutCollapsibleSections' => [
+		// See https://github.com/wikimedia/mediawiki-extensions-MobileFrontend?tab=readme-ov-file#wgmfnamespaceswithoutcollapsiblesections
+		'default' => [
```
c-oreills, line 4394: @pskyechology I've cherry picked in the commit from #6112 and then removed the override [as you requested](https://github.com/miraheze/mw-config/pull/6112/files#r2356921423).

Is it fine to have replicated the defaults in two different places? Should I include a comment to point developers to the other place in case they get updated?

pskyechology: It is the intended way and is documented in [CONTRIBUTING.md](https://github.com/miraheze/mw-config/blob/main/.github/CONTRIBUTING.md)

c-oreills:
@pskyechology thanks for you help! 😊

c-oreills:
It seems like this hasn't had the intended effect. I enabled it for the main namespace on my wiki and am still getting sections collapsing on mobile for e.g. [the main page](https://www.twistwoodtales.wiki/wiki/Main_Page).

I've reviewed [the code in question](https://github.com/wikimedia/mediawiki-extensions-MobileFrontend/blob/b7c59f5fe3c56390066b599236eaa19908cd7cf5/includes/ExtMobileFrontend.php#L108) and can't see anything that would obviously make it fail. They do disable section collapsing for parsoid too, so maybe there's just a different route that collapses sections now.

My usual next port of call would be to add debug logging, but obviously I can't do that here. Any suggestions @pskyechology ? I could also try to trace how the old disable collapsing setting worked before it was ripped out.

pskyechology:
@c-oreills You're most likely facing caching issues, as the main namespace does not have collapsible sections on my end. Try in incognito?

PR #6126, opened 2025-09-27
T14340: Set custom logo for dathereckoning.miraheze.org

Justman100:


pskyechology (commented):
We already support setting logos via ManageWiki. https://dathereckoning.miraheze.org/wiki/Special:ManageWiki/settings

PR #6323, opened 2026-02-20
T15003: Set 'vector-theme' preference to 'os' on lazerpigeonwiki

YTFGolf:
Resolves T15003 (the actionable parts at least)


pskyechology (approved):
Thank you!

PR #6358, opened 2026-04-02
T15102: Add another condition in referrer policy for private wiki which is to check whether map extension is active

Conscience93:
https://issue-tracker.miraheze.org/T15102

### DIFF
--- a/GlobalSettings.php
+++ b/GlobalSettings.php
@@ -932,8 +932,8 @@ $wgCaptchaRegexes[] = '/<a +href/i';
 // 12 MB
 $wgAPIMaxResultSize = 12582912;
 
-// Referrer Policy. Check whether the wiki is private and have maps extension not active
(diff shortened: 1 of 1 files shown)


GlobalSettings.php
```diff
@@ -932,7 +932,7 @@
 // 12 MB
 $wgAPIMaxResultSize = 12582912;
 
-$wgReferrerPolicy = $cwPrivate ?
+$wgReferrerPolicy = ( $cwPrivate && !$wi->isExtensionActive( 'maps' ) ) ?
```
lihaohong6: LGTM. Maybe include a comment in the line before noting the task (T15102) so that people know why this extra check is added.

Though the task number can be found via git blame, when code gets moved around, the initial commit introducing the change could get very hard to find.

lihaohong6 (changes requested):

PR #6364, opened 2026-04-07
Add reasons why settings are `managewiki-restricted`

YTFGolf:
Per discussion on Discord, this PR adds help descriptions to `managewiki-restricted` additional settings, similar to those that already exist for some restricted extensions. Extensions that do not have help info: AutoCreateCategoryPages, CirrusSearch, External Data, Wikimedia Incubator. Core ManageWiki settings and image dump generation also do not have any help info either but that can probably be ignored.

Not sure about formatting. The descriptions I wrote are educated guesses based on what I could find in the git blame and in MediaWiki documentation. See fc5854e for the commit hashes of five of these settings being restricted.


ManageWikiSettings.php
```diff
@@ -2940,7 +2940,7 @@
 		'maxint' => 360,
 		'overridedefault' => 7,
 		'section' => 'recentchanges',
-		'help' => 'Default amount of days to show at Special:RecentChanges. Users can override this at Special:Preferences.',
+		'help' => 'Default amount of days to show at Special:RecentChanges. Users can override this at Special:Preferences. Added in 773ac87.',
```
pixDeVl: Is there a reason to put a commit hash on the help text for end users? It doesn't really explain why it's restricted, and will be pretty confusing to most crats.

ManageWikiSettings.php
```diff
@@ -3004,7 +3004,7 @@
 		'type' => 'text',
 		'overridedefault' => 'HAWelcome',
 		'section' => 'restricted',
-		'help' => 'This configuration variable contains the name of the user that should leave the welcome message. If this user is flagged as a bot, the edits will be marked as a bot edit. Note: this should not be set to the same name of an existing user.',
+		'help' => 'This configuration variable contains the name of the user that should leave the welcome message. If this user is flagged as a bot, the edits will be marked as a bot edit. Note: this should not be set to the same name of an existing user, as doing so will lock that user out of the account. Restricted due to that issue.',
```
pixDeVl: To follow the format of previous messages, it would probably make more sense to have a sentence in the format of "Restricted: this setting may cause users who's usernames are entered to become locked out of their account" or the such

ManageWikiSettings.php
YTFGolf: That was because I wasn't sure exactly why, I imagine performance reasons but I'll let Random explain that one.

ManageWikiSettings.php
```diff
@@ -2924,7 +2924,7 @@
 		'maxint' => 62208000,
 		'overridedefault' => 180 * 24 * 3600,
 		'section' => 'recentchanges',
-		'help' => 'Items in the recentchanges table are periodically purged; entries older than this many seconds will be deleted.',
+		'help' => 'Items in the recentchanges table are periodically purged; entries older than this many seconds will be deleted. Restricted as if configured incorrectly, it will evict entries from RecentChanges and cause wikis to close earlier than they should.',
```
Universal-Omega: This is actually no longer true. This could *potentially* be unrestricted or restricted for a *different* reason. I changed CreateWiki awhile ago to rely on logging and revision tables for closing wikis, not recentchanges at all anymore.

YTFGolf: So would it be better for this PR to leave this setting alone then?

pskyechology (approved):

PR #6368, opened 2026-04-12
Add Translations and Translations_talk namespaces to ManageWikiExtensions

KK2-5:
Ref T15219


ManageWikiExtensions.php
```diff
@@ -1818,6 +1818,28 @@
 				'translate_tms' => 'extensions/Translate/sql/mysql/translate_tm.sql',
 				'translate_translatable_bundles' => 'extensions/Translate/sql/mysql/translate_translatable_bundles.sql',
 			],
+			'namespaces' => [
+				'Translations' => [
+					'id' => 1198,
+					'searchable' => 0,
+					'subpages' => 0,
```
lihaohong6: This namespace does in fact allow subpages. Same goes for the corresponding talk namespace. See https://github.com/wikimedia/mediawiki-extensions-Translate/blob/756fd526f5a672f02d291c88deb3524c4e73140c/src/HookHandler.php#L233.

ManageWikiExtensions.php
```diff
@@ -1818,6 +1818,28 @@
...
+					'protection' => 'translate',
+					'content' => 0,
+					'aliases' => [],
+					'contentmodel' => 'wikitext',
+					'additional' => [],
```
lihaohong6: If you put something like
```
'additional' => [
    'wgNamespaceRobotPolicies' => 'noindex',
],
```
here. Whenever the Translate extension is enabled, the newly created namespace will start with `noindex` by default, which aligns with extension defaults in https://github.com/wikimedia/mediawiki-extensions-Translate/blob/756fd526f5a672f02d291c88deb3524c4e73140c/src/HookHandler.php#L411

lihaohong6 (changes requested):
Thanks for the PR. Mostly good except for 2 minor issues.

ManageWikiExtensions.php
KK2-5: Hmm, interesting… I tested on a local instance and it does seem to be true, but this doesn't seem to apply on Miraheze (hence the Phorge request). Was it because this entry was missing here?

lihaohong6: For settings that are under ManageWiki's control, it will override anything set by the extension itself.

So Extension:Translate would add entries to `$wgNamespaceRobotPolicies`, but since this is under ManageWikiPermissions.php, ManageWiki will override `$wgNamespaceRobotPolicies` with its own settings. And that's why we also need to explicitly add namespaces and default user permissions to ManageWiki when installing an extension.

lihaohong6 (approved):

PR #6374, opened 2026-04-15
Toggle for replacing CodeEditor with CodeMirror

KockaAdmiralac:
Wikimedia are looking to [replace CodeEditor with CodeMirror](https://phabricator.wikimedia.org/T419332), so I thought a toggle for this might be useful. CodeMirror also seemingly has improved support for CSS linting, and doesn't yell at me when I save a file with CSS variables. There is also some autocompletion support for Scribunto.

I took [this particular block](https://github.com/wikimedia/operations-mediawiki-config/blob/a25f8b8/wmf-config/CommonSettings.php#L3020-L3049) from Wikimedia's config and added AbuseFilter ([pending change](https://gerrit.wikimedia.org/r/c/operations/mediawiki-config/+/1268652)) and DataMaps ([pending change](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/DataMaps/+/1271952)) to it. AbuseFilter support is not going to get backported to 1.45, but it's there just in case.

### DIFF
--- a/GlobalSettings.php
+++ b/GlobalSettings.php
@@ -171,6 +171,41 @@ if ( $wi->isExtensionActive( 'VisualEditor' ) ) {
 
 if ( $wi->isExtensionActive( 'CodeMirror' ) ) {
 	$wgDefaultUserOptions['usecodemirror'] = (int)$wmgCodeMirrorEnableDefault;
+	if ( $wmgCodeMirrorReplaceCodeEditor ) {
+		// Taken from WMF's configuration.
+		$wgCodeMirrorEnabledModes['javascript'] = true;
+		$wgCodeMirrorEnabledModes['json'] = true;
+		$wgCodeMirrorEnabledModes['css'] = true;
+		$wgCodeMirrorEnabledModes['lua'] = true;
+		$wgCodeMirrorEnabledModes['vue'] = true;
+		// CodeEditor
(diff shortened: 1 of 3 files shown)


The-Voidwalker:
Not sure this is needed since we don't enable either extension by default. Users could just toggle on or off whichever one they want to use. Does somewhat raise the point that the two extensions should probably be flagged as conflicting with each other though.

KockaAdmiralac:
> Users could just toggle on or off whichever one they want to use.

No, they can't. Firstly because CodeEditor cannot handle wikitext highlighting, so users currently enabling CodeMirror do so because of its wikitext highlighting, and may not want to switch to CodeMirror for code highlighting as well. And secondly because not all extensions implement compatibility with CodeMirror either, so turning CodeEditor off would disable syntax highlighting in AbuseFilter (until the 1.46 upgrade, but there are probably other examples).

The way I implemented this toggle is the way that WMF does it, and they enable both extensions in production, as far as I'm aware.

The-Voidwalker:
Ah, I didn't quite get that, having assumed they were feature complete with each other. I also hadn't appreciated that CodeMirrorEnabledModes defaults to mediawiki only currently. Making this toggle necessary if you are only using CodeMirror to get all its features. I think. Might be worth having the toggle auto-enabled for wikis with only CodeMirror enabled?

PR #6382, opened 2026-04-22
T15278: New Vector-2022 Default Preferences for realitycompromised.miraheze.org

IPutMyPCInTheFridge:
Changes to default width, pinned navigation, font size, and theme. See https://issue-tracker.miraheze.org/T15278.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5339,6 +5339,13 @@ $wgConf->settings += [
 			'usenewrc' => 0,
 			'thumbsize' => 3,
 		],
+		'+realitycompromisedwiki' => [
+			'vector-limited-width' => '0',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -5339,6 +5339,13 @@
 			'usenewrc' => 0,
 			'thumbsize' => 3,
 		],
+		'+realitycompromisedwiki' => [
+			'vector-limited-width' => '0',
```
YTFGolf: This should be a number rather than a string, ditto for the following 3 lines

IPutMyPCInTheFridge: Should be fixed now, copied vector-theme as a base and forgot to actually change em.

YTFGolf (approved):

PR #6392, opened 2026-04-30
Modifying read access on govnpwiki

himvat-dev:
Removing read access for everyone (*) user group.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3694,6 +3694,10 @@ $wgConf->settings += [
 				'read' => true,
 			],
 		],
+	    '+govnpwiki' => [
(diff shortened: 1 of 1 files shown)


YTFGolf:
Why is this necessary over just making the wiki private?

himvat-dev:
Yes, the wiki is already private, but the main page can be still accessed or seen so I don't want that as well so restricting the read rights completely.

RhinosF1:
This isn't needed. Setting a wiki to private will remove the 'read' right from the '*' group automatically. It doesn't affect the fact that the main page can be seen which is done through the read whitelist.

PR #6398, opened 2026-05-02
Add wgNoFollowDomainExceptions to ManageWiki

tbodt:
See [T15325](https://issue-tracker.miraheze.org/T15325)


ManageWikiSettings.php
```diff
@@ -1536,6 +1536,16 @@
 		'help' => 'If enabled, external links in wikitext will be given the <code>rel="nofollow"</code> attribute.',
 		'requires' => [],
 	],
+	'wgNoFollowDomainExceptions' => [
+		'name' => 'NoFollow Domain Exceptions',
+		'from' => 'mediawiki',
+		'global' => true,
+		'type' => 'texts',
+		'overridedefault' => [],
```
lihaohong6, line 1544: Maybe this should default to whitelisting miraheze.org and wikitide.org domains?

tbodt: Not a bad idea but I'm not the one to make the call

tbodt: Created https://issue-tracker.miraheze.org/T15330 to attempt to separate this decision into a separate task.

PR #6404, opened 2026-05-06
Set global usage for govnp wikis

himvat-dev:
Set global usage for _govnpcommons_ and _govnpmicro_ for shared media


pskyechology:
Per https://meta.miraheze.org/wiki/Steward_requests/(Un)deletions#GovNP_Commons_Wiki

PR #6405, opened 2026-05-08
T15267: Add wgSitemapNamespaces for sitemap generation

Conscience93:
Resolves T15267 (https://issue-tracker.miraheze.org/T15267)


ManageWikiNamespaces.php
```diff
@@ -152,6 +152,37 @@
 		'help' => 'Overrides <code>$wgDefaultRobotPolicy</code> for this namespace.',
 		'requires' => [],
 	],
+	'wgSitemapNamespaces' => [
+		'name' => 'Allow which main-namespace content for sitemap generation only?',
+		'from' => 'mediawiki',
+		'type' => 'list',
```
The-Voidwalker: Should use the check type instead. I don't think the options you've created are considered valid for the config option.

The-Voidwalker (changes requested):

ManageWikiNamespaces.php
Conscience93: Meaning it's either false or all namespaces?

YTFGolf: Meaning it should work similar to `$wgTemplateStylesNamespaces`, if that helps.

Conscience93: ah got it

YTFGolf: Wait no I'm completely spreading misinformation, I meant `$wgTemplateSandboxEditNamespaces` which has `'type' => 'check'`. Sorry for the confusion.

ManageWikiNamespaces.php
```diff
@@ -155,17 +155,16 @@
...
-			NS_USER => true,
-			NS_PROJECT => true,
-			NS_HELP => true,
-			NS_TEMPLATE => true,
-			NS_MODULE => true,
+			0 => true,
+			2 => true,
+			4 => true,
+			10 => true,
+			12 => true,
```
YTFGolf: I think this should be
```php
		'overridedefault' => [
			NS_MAIN => true,
			NS_USER => true,
			NS_PROJECT => true,
			NS_TEMPLATE => true,
			NS_HELP => true,
			NS_MODULE => true,
			'default' => false,
		],
```
based on what PetraMagna said on the task

Conscience93: oh they are different? Oops lemme change it

PR #6408, opened 2026-05-13
T15373: Set wgSVGNativeRendering to false on vishwamitranarchiveswiki

Conscience93:
Resolves T15373 (https://issue-tracker.miraheze.org/T15373)


pskyechology (changes requested):
This could be added to ManageWiki instead.

Conscience93:
> This could be added to ManageWiki instead.

ManageWiki means put in ManageWikiSetting.php to give option for user to disable svg rendering?

RhinosF1:
> > This could be added to ManageWiki instead.
> 
> ManageWiki means put in ManageWikiSetting.php to give option for user to disable svg rendering?

Yes

PR #6411, opened 2026-05-16
T15383: allow disabling restricted extensions

lihaohong6:
Applied to extensions where there are no known adverse side effects.

Depends on https://github.com/miraheze/ManageWiki/pull/789


ManageWikiExtensions.php
```diff
@@ -311,7 +311,9 @@
 		'conflicts' => 'semanticmediawiki',
 		'requires' => [
 			'permissions' => [
-				'managewiki-restricted',
+				'enable' => [
+					'managewiki-restricted',
```
Universal-Omega: This one has potential issues with disabling also it should be kept fully restricted I think.

PR #6413, opened 2026-05-18
T15397: Add wgBucketMaxDataPerPage + configure for sagan4alphawiki

YTFGolf:
Resolves T15397

Not sure if 10,000,000 is too high, the wiki currently . Also wondering if I should put underscores in these numbers so you can actually tell they're different, only reason I didn't was because there wasn't any precedent in the file.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -419,6 +419,10 @@ $wgConf->settings += [
 	'wgBucketDBuser' => [
 		'default' => 'bucketuser',
 	],
+	'wgBucketMaxDataPerPage' => [
+		'default' => 1000000,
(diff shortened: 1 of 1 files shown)


Disgustedorite:
> Not sure if 10,000,000 is too high

As noted in T15397 the limit is arbitrary and the performance is not expected to be too bad, it mostly affects how quickly the page can be saved tmu. I'm successfully writing thousands of rows on some pages already and it's not affecting performance much especially compared to semantic mediawiki which will write this much data and more without complaint lol. We're gonna take steps to prevent that much data from being written in future ecosystem pages but we need a good buffer for futureproofing and due to the length of the page title having a significant impact on how many rows can be inserted.

PR #6426, opened 2026-05-27
add fallback handler for GeoJson content model while maps is disabled

The-Voidwalker:


The-Voidwalker:
Handled better by #6428

PR #6428, opened 2026-05-27
add FallbackContentHandler for content models added by non-enabled extensions

The-Voidwalker:
fixes T14017


The-Voidwalker:
~~Not ready for merge. When testing on beta, disabling and enabling an extension that adds a content model (TemplateStyles), the content model (sanitized-css) _remains unavailable_ for reasons I am finding impossible to explain.~~

fixed

PR #6441, opened 2026-06-16
T15536: disable EventStreamConfig logging

pskyechology:
Unnecessary pollution in Graylog; presently of no use to us.

Resolves T15536


pskyechology (commented):
did not blow up Mirabeta

PR #6447, opened 2026-06-23
T15556: Remove wikis that no longer exist from mw-config

kai8484:


LocalSettings.php
The-Voidwalker, line 3740: Remove these lines. The first four are a part of the block you removed in your commit. The rest are for another deleted wiki that needs to be removed.

The-Voidwalker (changes requested):

PR #6448, opened 2026-06-24
add script for removing config for deleted wikis

The-Voidwalker:


YTFGolf:
I can't speak for the code itself but I did manually check all the wikis removed in this commit using [this method](https://meta.miraheze.org/wiki/Special:ApiSandbox#action=query&format=json&list=wikiconfig&formatversion=2&wcfprop=sitename&wcfwikis=brolandiawiki%7Cwikitrashwiki%7Cwikilexiconwiki%7Cmoviepediawiki%7Cthegreatwarwiki%7Cosmaniawiki%7Cbanjokazooiewiki%7Cfwtdwiki%7Clandarwiki%7Cssbuniversewiki%7Cstickmancomicwiki) (I did this multiple times so that link won't have all the wikis on it) and found no false positives.

The-Voidwalker:
I think the code could do with
- better error handling for curl and/or API problems;
- in testing, processLocalWiki seems to check the same wiki multiple times and I'm not sure how that happens without an infinite loop, but it works for now (might be related to below);
- better handling for [stacked case statements](https://github.com/miraheze/mw-config/blob/07ac19b73fc2becc89dc1ec7689c4bbb55235f5a/LocalWiki.php#L74-L86) (if softcellwiki is deleted, it will remove the config block even though it shouldn't).

There's also [this todo](https://github.com/miraheze/mw-config/pull/6448/changes#diff-8a0d75743636f440b5cdf7b4f856394bb43378c3dabee6fa974444afa45bbee4R75) but it isn't actually relevant for any of the code that gets processed by this script.

The-Voidwalker:
I'm also considering whether or not we want something like a github actions workflow or other utility to run the script on a (semi) regular basis.

RhinosF1:
> I'm also considering whether or not we want something like a github actions workflow or other utility to run the script on a (semi) regular basis.

It's a maintenance script so why not use a timer on mwtask like the rest of our regular jobs. I don't see why this needs something new.

cleanup/CleanupMissingWikis.php
```diff
@@ -0,0 +1,316 @@
+<?php
+/*
+ * This script is intended to function as a standalone utility. It is therefore
```
RhinosF1, line 3: Why?

RhinosF1 (commented):
Why are we putting a script in config unaware of mediawiki?

It feels like an antipattern

PR #6461, opened 2026-07-12
Add 1.46 configuration to LocalSettings.php

YTFGolf:
This at the moment doesn't include `wgRestrictedGroups` or `$wgScribuntoEngineConf['luasandbox']['shareInvocationEnv']`. The first of these could probably be added here, but possibly would require more configuration. The second I think could be added as a wmg variable.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -1693,6 +1693,9 @@ $wgConf->settings += [
 		'default' => false,
 		'ysmwikiwiki' => true,
 	],
+	'wgNamespacesWithoutAutoSummaries' => [
+		'1.46' => [],
+	],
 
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -1693,6 +1693,9 @@
 		'default' => false,
 		'ysmwikiwiki' => true,
 	],
+	'wgNamespacesWithoutAutoSummaries' => [
+		'1.46' => [],
```
Universal-Omega: Might as well just use default for these, should hurt if they are added on 1.45 and reduce chance cache issue happens and they don't appear on 1.46 once upgraded either.

PR #6468, opened 2026-07-19
Add extendedconfirmed protection level and user group for otshestwiki

dmehus:
Per https://issue-tracker.miraheze.org/T15685

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4086,7 +4086,7 @@ $wgConf->settings += [
 		],
 		'+otshestwiki' => [
 			'extendedconfirmed' => [
-				'editextendeconfirmedprotected' => true,
+				'editextendedconfirmedprotected' => true,
 			],
 		],
 		'+phightingwiki' => [
@@ -5790,6 +5790,7 @@ $wgConf->settings += [
 		],
 		'+otshestwiki' => [
 			'editextendedconfirmedprotected',
+			'sysop',
 		],
 		'+phightingwiki' => [
 			'edittrusteduserprotected',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -4084,6 +4084,11 @@
 				'edit-admin-pages' => true,
 			],
 		],
+		'+otshestwiki' => [
+			'extendedconfirmed' => [
+				'editextendeconfirmedprotected' => true,
+			],
+		],
```
YTFGolf, line 4094: This isn't really necessary as it can be done through managewiki

dmehus: Yeah, I know. It said it was an optional step, so decided to do it for them anyway.

YTFGolf:
The right also needs to be added to `$wgAvailableRights`

LocalSettings.php
```diff
@@ -4084,6 +4084,11 @@
 				'edit-admin-pages' => true,
 			],
 		],
+		'+otshestwiki' => [
+			'extendedconfirmed' => [
+				'editextendeconfirmedprotected' => true,
```
YTFGolf: typo, also probably might as well give it to sysop too

dmehus: Good point. Might as well.

LocalSettings.php
```diff
@@ -5783,6 +5788,10 @@
 		'+nomifactoryceuwiki' => [
 			'editmostlyprotectedprotected',
 		],
+		'+otshestwiki' => [
+			'editextendedconfirmedprotected',
+			'sysop',
```
YTFGolf: sysop protection already exists as `editprotected`, by "give it to sysops" I meant for wgManageWikiPermissionsAdditionalRights

dmehus: Okay, so was I not needing to do anything then, should I just revert adding `sysop` to that group?

YTFGolf: yeah. and again, as I say, it is completely optional to do the managewikipermissionsadditionalrights

PR #6471, opened 2026-07-25
set wgCreateWikiClosedReasons T15628

Reception123:
once miraheze/CreateWiki#828 is merged


dmehus (commented):
This looks good to me. Only two I'd suggest might be one for Trust and Safety reasons, not specifically limited to the Terms of Use (could be Privacy Policy or other legal reason), and the other one would be per a request to stewards on [[m:Stewards requests|Steward requests]] that wasn't the result of a community discussion

Reception123:
> This looks good to me. Only two I'd suggest might be one for Trust and Safety reasons, not specifically limited to the Terms of Use (could be Privacy Policy or other legal reason), and the other one would be per a request to stewards on [[m:Stewards requests|Steward requests]] that wasn't the result of a community discussion

Thanks for the suggestion, probably best to indeed rename to Trust and Safety action rather than mentioning the ToU. As for a request for Stewards, what situation would you have in mind where Stewards would close a wiki that's not the result of a community discussion or for a Content Policy violation?

dmehus:
> > This looks good to me. Only two I'd suggest might be one for Trust and Safety reasons, not specifically limited to the Terms of Use (could be Privacy Policy or other legal reason), and the other one would be per a request to stewards on [[m:Stewards requests|Steward requests]] that wasn't the result of a community discussion
> 
> Thanks for the suggestion, probably best to indeed rename to Trust and Safety action rather than mentioning the ToU. As for a request for Stewards, what situation would you have in mind where Stewards would close a wiki that's not the result of a community discussion or for a Content Policy violation?

Sounds good re: simply renaming ToU as Trust and Safety. No need for two different reasons for them, I think you're right there.

For the second reason, it would mainly be a wiki structured as a 'community wiki' with no active community to speak of. You could _call it_ per "community discussion," but that might be stretching it a bit if the only participant is the discussion proponent and there's active 'community' to speak of.

dmehus:
Just realized there's no reason for a Technology Team-closure/deletion as per Terms of Use section 9. Perhaps it _may_ still be useful to have a ToU reason?

Reception123:
Thanks for the suggestions/feedback! As we discussed privately as well, I do think that the local community reason might be a bit too niche for now but we could consider adding it later if it really does come up a lot. As for the others, I think to make things less confusing we could have a Technology Team action and TS action, and reconsider later if we have to.

PR #6472, opened 2026-07-26
T15724: Change footer icons for zvhwiki

TheVoltaireian:
https://issue-tracker.miraheze.org/T15724

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2582,7 +2582,6 @@ $wgConf->settings += [
 				],
 			],
 		],
-	],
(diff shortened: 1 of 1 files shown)


pixDeVl:
Checked over the config format and contents, minus some small confusion with the keys @pskyechology explained on DC it all looks consistent with what hasn't burst into flames, so, **+1** from me.

LocalSettings.php
YTFGolf: this should be after the zvhwiki stuff

PR #6474, opened 2026-07-28
add 4 domains to the CSP

Reception123:
approved in T13196 T15622 T15519 T15516


ContentSecurityPolicy.php
```diff
@@ -226,6 +229,7 @@
 		'video.fastly.steamstatic.com',
 		'shared.fastly.steamstatic.com',
 		'*.instatus.com'
+		'tally.so'
```
YTFGolf, line 232: Wouldn't it be better to just format the CSP with trailing commas (like we do in the rest of mw-config) so we don't get syntax errors like this?

dmehus: Seems reasonable to me.

Reception123: I agree, it can be a little annoying

PR #6483, opened 2026-08-03
Update citizen config

YTFGolf:
https://issue-tracker.miraheze.org/T15760

Updating Citizen config to 3.19.

Removed config:
```
wgCitizenEnableCommandPalette
wgCitizenMaxSearchResults
wgCitizenSearchDescriptionSource
wgCitizenSearchGateway
```

Variables not (yet) in this PR:
```
wgCitizenCompat
wgCitizenEnablePreferences
wgCitizenPreview
```


YTFGolf:
Ready for review, but gonna ask Petrus + Alistair about the config settings I haven't done

YTFGolf:
not entirely sure what the manifest or share options do but they're there ig

alistair3149:
Manifest means [web application manifest](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest). Citizen generates them automatically so that the wiki is installable as a progressive web app. `wgCitizenManifestOptions` is used to customize some properties of the manifest.

Share is the [custom share panel](https://mwcitizen.skin/features/share). By default, Citizen uses the native share when possible, but sometimes folks want their own share panel, and some browser (e.g. Firefox) does not support web share API.

lihaohong6:
It seems that `$wgCitizenThemeColor` has deviated from our default. We should probably change it. Upstream has `#0d0e12` but ours is `#131a21`.

https://github.com/StarCitizenTools/mediawiki-skins-Citizen/blob/516013b1b3dae0087f9f27a845d3df0e2e70f201/skin.json#L1016

For this reason I wonder if it makes more sense if we don't add `$wgCitizenManifestOptions` to LocalSettings.php. If we do, we will be liable for maintaining it and matching upstream. Since it's not configurable on ManageWiki, we can just let skin.json populate its default value, which shouldn't make a difference.

ManageWikiSettings.php
```diff
@@ -4450,6 +4428,58 @@
...
+			'Top' => 'top',
+			'Bottom' => 'bottom',
+		],
+		'overridedefault' => 'bottom',
+		'section' => 'styling',
+		'help' => 'Position of the header on the mobile and tablet layout (below the desktop breakpoint).',
+		'requires' => [],
+	],
+	'wgCitizenTableOfContentsCollapseAtCount' => [
+		'name' => 'Table of Contents Collapse at Count',
```
lihaohong6: The name is not prefixed by Citizen unlike the other ones.

YTFGolf: wait how does that work then? I thought all of them would be prefixed like `wgCitizen`.

YTFGolf: Oh right in the description yeah

PR #6486, opened 2026-08-07
Add wgIWDPrefixes to ManageWiki/settings

dmehus:
Per discussion with @Reception123 on Discord. Tagging him on this as it will need his review as I will most likely need to make changes to this or otherwise require additional commits...

### DIFF
--- a/ManageWikiSettings.php
+++ b/ManageWikiSettings.php
@@ -1536,6 +1536,17 @@ $wgManageWikiSettings = [
 		'help' => 'Set a default target for external links.',
 		'requires' => [],
 	],
+	''wgIWDprefixes' => [
+    	'name' => 'Interwiki Dispatcher',
+		'from' => 'mediawiki',
+		'global' => true,
+		'type' => 'check'
+    	'section' => 'links',
+    	'overridedefault' => true,
+    	'section' => 'links',
+    	'help' => 'If enabled, you will be able to control multi-level interwiki prefixes on your [[Special:Interwiki]] table', // i18n key or descriptive comment
+		'requires' => [],
+	],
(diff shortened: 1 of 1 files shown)


Universal-Omega:
This variable is not supported by ManageWiki and it needs rarely changed I think keeping it just in LocalSettings would be the better option then doing a hack and splitting it into numerous variables in ManageWiki tbh.

dmehus:
> This variable is not supported by ManageWiki and it needs rarely changed I think keeping it just in LocalSettings would be the better option then doing a hack and splitting it into numerous variables in ManageWiki tbh.

Sounds goods, thanks, CA! I will close this PR then. Should we just override it on wikis in LocalSettings.php then?

Universal-Omega:
> > This variable is not supported by ManageWiki and it needs rarely changed I think keeping it just in LocalSettings would be the better option then doing a hack and splitting it into numerous variables in ManageWiki tbh.
> 
> Sounds goods, thanks, CA! I will close this PR then. Should we just override it on wikis in LocalSettings.php then?

Yes but wiki overrides need a +dbname, like +metawiki, so we don't have to repeat the default each time. The + will cause it to append to the default.

dmehus:
> > > This variable is not supported by ManageWiki and it needs rarely changed I think keeping it just in LocalSettings would be the better option then doing a hack and splitting it into numerous variables in ManageWiki tbh.
> > 
> > 
> > Sounds goods, thanks, CA! I will close this PR then. Should we just override it on wikis in LocalSettings.php then?
> 
> Yes but wiki overrides need a +dbname, like +metawiki, so we don't have to repeat the default each time. The + will cause it to append to the default.

Thanks! Noted.

PR #6490, opened 2026-08-12
T15770: add footer icon for capoeirawiki

KK2-5:
Resolves [T15770](https://issue-tracker.miraheze.org/T15770)

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2794,7 +2794,7 @@ $wgConf->settings += [
 				],
 			],
 		],
-		'+capoeirawiki' => [
+		'capoeirawiki' => [
 			'iwf' => [
 				'iwf' => [
 					'src' => 'https://static.wikitide.net/capoeirawiki/3/38/Iwf_footer_icon.png',
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -2794,6 +2794,15 @@ $wgConf->settings += [
 				],
 			],
 		],
+		'+capoeirawiki' => [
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -2794,6 +2794,15 @@
 				],
 			],
 		],
+		'+capoeirawiki' => [
```
YTFGolf: The + is already inside +wgFooterIcons so the + on this line is not necessary

KK2-5: I thought so as well, but the + prevents overwriting the Miraheze logo. PixDeV helped me figure this one out yesterday [via Discord](https://wm-bot.wmcloud.org/browser/index.php?start=08%2F12%2F2026&end=08%2F12%2F2026&display=%23miraheze) (and seems to be the current deployment on meta.mirabeta)

YTFGolf: yes, + means that you're adding configuration rather than replacing it, but if you scroll up 900 lines you'll see that the configuration you're editing is called `+wgFooterIcons`. Therefore the + here is unnecessary.

YTFGolf (approved):

LocalSettings.php
YTFGolf: wait nevermind I was completely wrong about this whoops

KK2-5: Oh don't worry, it was my mistake for not requesting a second pair of eyes when I was confident about the outcome. It was a good learning opportunity either way. Thanks for taking the time!

PR #6491, opened 2026-08-14
Remove skipcaptcha from ManageWiki Permissions blacklist for the user…

dmehus:
… group

@Universal-Omega and @Reception123 Emiliers, Raidarr and I were having a chat in #miraheze on IRC and Discord. I did not realize we had since blacklisted the `skipcaptcha` user right from the `user` group. It was always just blacklisted from the `*` group, which makes sense. Could we please consider removing this from the `user` group? I think it's a bit too strict.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -4453,7 +4453,6 @@ $wgConf->settings += [
 			'user' => [
 				'autoconfirmed',
 				'noratelimit',
-				'skipcaptcha',
 				'managewiki-core',
 				'managewiki-extensions',
 				'managewiki-namespaces',


RhinosF1 (changes requested):
no, no, no and if this wasn't clear, no.

All users shouldn't have skipcaptcha unless we want a lot of spambots.

pskyechology:
Please open a Phorge ticket and consult with Stewards before making such PRs. From a tech standpoint, this is a bad idea, since unknowing crats could make a relatively important bot/CVT countermechanism completely obsolete.
Any issues with captchas should be resolved instead of circumventing them entirely.

dmehus:
I disagree with these assessments. For one thing, the problem of spambots are best thwarted at _account creation_, and between @Agent-Isai and @Reception123, they have been regularly importing the global blocks from Wikimedia. This alone has largely solved the spambot issues. We also have Cloudflare-level verifications. In this way, in my view, allowing `skipcaptcha` to be used in the `*` user group would've been a completely bad idea for the reasons discussed above. For the `user` group, I disagree. Note this would not change the default configuration. We have also deployed StopForumSpam globally, and I have not proposed unrestricting the `sfs-bypass` user right for this reason.

It's the practice of Miraheze to allow as much local control over wikis as possible. I would also note that Miraheze previously restricted the `interwiki` user right against _any_ local delegation, which was the reason for the global Interwiki Administrators group. We then allowed local elections for local Interwiki Administrators, but allowed only Stewards to grant the user group. This was further devolved to allow local Bureaucrats to manage the user right and group completely. This seemed like a reasonable change, but, granted, I have only discussed it with one Steward and should've discussed it more broadly.

RhinosF1:
@dmehus: Our tools on Cloudflare are not designed to prevent spambots in anyway. Doing so would likely impact a significant number of legitimate users and automated tools.

We generally would not and I'm not sure we've ever deliberately used tech side controls to mitigate against community managed issues.

Cloudflare is there to prevent scraping etc. which would be significantly detrimental to the performance of the whole farm or things like vulnerability checks and DOS attacks.

PR #6495, opened 2026-08-15
T15830: Disable new ParserMigration defaults

pixDeVl:
As of MediaWiki 1.46, ParserMigration defaults to enabling Parsoid on articles and discussion pages, which is effectively a blocker for upgrading the farm. This explictly adds false defaults to the changed settings, to at least throw this issue to 1.47.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5311,6 +5311,12 @@ $wgConf->settings += [
 	],
 
 	// ParserMigration
+	'ParserMigrationEnableParsoidArticlePages' => [
+		'default' => false,
+	]
+	'ParserMigrationEnableParsoidDiscussionTools' => [
+		'default' => false,
+	],
 	'wgParserMigrationEnableQueryString' => [
 		'default' => true,
 	],
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5311,10 +5311,16 @@ $wgConf->settings += [
 	],
 
 	// ParserMigration
+	'wgParserMigrationEnableParsoidMobileArticlePages' => [
+		'default' => false,
+	],
+	'wgParserMigrationEnableParsoidMobileTalkPages' => [
+		'default' => false,
+	],
 	'wgParserMigrationEnableParsoidArticlePages' => [
 		'default' => false,
 	],
-	'wgParserMigrationEnableParsoidDiscussionTools' => [
+	'wgParserMigrationEnableParsoidTalkPages' => [
 		'default' => false,
 	],
 	'wgParserMigrationEnableQueryString' => [
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -5311,10 +5311,10 @@ $wgConf->settings += [
 	],
 
 	// ParserMigration
-	'ParserMigrationEnableParsoidArticlePages' => [
+	'wgParserMigrationEnableParsoidArticlePages' => [
 		'default' => false,
-	]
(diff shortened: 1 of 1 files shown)


RhinosF1 (commented):
Should we add this to ManageWiki in case someone wants to try it out?

pixDeVl:
> Should we add this to ManageWiki in case someone wants to try it out?

Hmmm, I wouldn't turn it on on my wiki but I don't see why not, could maybe be a useful part for NextTide too. We should have a good help text on the setting so people understand what it means, but I don't know what to write, so I'll just leave it blank/very basic and add that real quick.

pixDeVl:
Per chatting with Rhinos on DC, going to not add the setting in this PR, will be it's own subtask

SomeMWDev:
> Should we add this to ManageWiki in case someone wants to try it out?

Based on an analysis I did a while ago for Weird Gloop, on 1.45, Parsoid is about 2.2x slower for content pages than the legacy parser. According to the CTT, this has improved in 1.46 and 1.47. I think we should restrict the setting and only enable it on smaller wikis that don't have a lot of traffic for now.

LocalSettings.php
```diff
@@ -5311,6 +5311,12 @@
 	],
 
 	// ParserMigration
+	'wgParserMigrationEnableParsoidArticlePages' => [
+		'default' => false,
+	],
+	'wgParserMigrationEnableParsoidDiscussionTools' => [
```
SomeMWDev: Looks like this setting is being renamed in 1.46: https://github.com/wikimedia/mediawiki-extensions-ParserMigration/commit/b8ed877d9c711d70aa1ea0596fc8ab472ae31c10

Since the default of the old setting is `false` anyway on 1.45, it probably makes sense to use the new name here already

SomeMWDev (changes requested):
Maybe we should disable https://github.com/wikimedia/mediawiki-extensions-ParserMigration/blob/a86122b1f935e96f439ee64fb14f30f58756b92d/extension.json#L37-L44 as well?

SomeMWDev:
> Maybe we should disable https://github.com/wikimedia/mediawiki-extensions-ParserMigration/blob/a86122b1f935e96f439ee64fb14f30f58756b92d/extension.json#L37-L44 as well?

(Should've probably already noted this in the task, sorry about that)

pixDeVl:
> Maybe we should disable https://github.com/wikimedia/mediawiki-extensions-ParserMigration/blob/a86122b1f935e96f439ee64fb14f30f58756b92d/extension.json#L37-L44 as well?

If these are also new defaults compared to 1.45, then yes they should be.

LocalSettings.php
pixDeVl: > Looks like this setting is being renamed in 1.46: [wikimedia/mediawiki-extensions-ParserMigration@b8ed877](https://github.com/wikimedia/mediawiki-extensions-ParserMigration/commit/b8ed877d9c711d70aa1ea0596fc8ab472ae31c10)
> 
> Since the default of the old setting is `false` anyway on 1.45, it probably makes sense to use the new name here already

So, the following will all be explicitly false.
ParserMigrationEnableParsoidMobileArticlePages
ParserMigrationEnableParsoidTalkPages
ParserMigrationEnableParsoidArticlePages 
ParserMigrationEnableParsoidMobileTalkPages

SomeMWDev (approved):

PR #6499, opened 2026-08-18
T15865: Restrict Page Forms per performance issues

pixDeVl:


ManageWikiExtensions.php
```diff
@@ -1572,8 +1572,13 @@
 	'pageforms' => [
 		'name' => 'PageForms',
 		'linkPage' => 'https://www.mediawiki.org/wiki/Special:MyLanguage/Extension:Page_Forms',
+		'help' => 'Not enabled on new wikis, per preformance issues.',
```
YTFGolf: ...

YTFGolf: also even with fixing the typo I think "per performance" is a bit too much purring (once again, pix), imo "due to performance issues" is nicer to read

Universal-Omega: "Not being enabled on new wikis due to performance issues." Reads a little better to me. Up to you though.

PR #6503, opened 2026-08-21
Add `bucketwiki` as an import source if Bucket enabled

YTFGolf:
See https://issue-tracker.miraheze.org/T15340

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3440,6 +3440,9 @@ $wgConf->settings += [
 		'+zhdelwiki' => [
 			'zhwikipedia',
 		],
+		'+ext-Bucket' => [
+			'bucketwiki',
+		],
 	],
 
 	'wgExportMaxHistory' => [
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3441,7 +3441,9 @@ $wgConf->settings += [
 			'zhwikipedia',
 		],
 		'+ext-Bucket' => [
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3440,6 +3440,9 @@
 		'+zhdelwiki' => [
 			'zhwikipedia',
 		],
+		'+ext-Bucket' => [
+			'bucketwiki',
```
SomeMWDev: Does this interwiki prefix exist on all wikis using bucket?

YTFGolf: no

YTFGolf: my comment on the task was "add `bucketwiki` ... to the global interwiki table" and then I think this would work

SomeMWDev: This probably won't work then per https://www.mediawiki.org/wiki/Manual:$wgImportSources
> $wgImportSources can be a simple flat array that contains valid interwiki prefixes.

SomeMWDev: > my comment on the task was "add `bucketwiki` ... to the global interwiki table" and then I think this would work

Ah okay. I think we could also use the `mh` interwiki though per https://www.mediawiki.org/wiki/Manual:$wgImportSources#Syntax, which would be easier

YTFGolf: Updated, although will add that pix added bucketwiki to global interwiki table

PR #6513, opened 2026-08-28
Optimize MirahezeFunctions

SomeMWDev:
There is more stuff that can be optimized, but let's start with this and see if it has any effect on the wall time of MirahezeFunctions in the aggregations.

https://issue-tracker.miraheze.org/T15909


SomeMWDev:
Tested this on mirabeta, nothing obvious broke, but it's also difficult to see the effects on beta

Universal-Omega (approved):
Looks good to me. I have some other ideas on how to optimize more things in MirahezeFunctions also but this is good for the worst parts of it. A lot was actually left over from when you could access beta from prod so that things could be proxied from test151 but I removed that awhile ago so definitely isn't needed like that anymore.

PR #6515, opened 2026-08-29
WikiOasis Added to IWD list

NotAracham:
Per community request, creating a multi-level prefix to support linking to WikiOasis wikis in the localsettings IWD definition


NotAracham:
Closing pull request, excessive changes left over from previous testing that should not be merged. :(

PR #6516, opened 2026-08-29
Update LocalSettings.php to include WO in IWD

NotAracham:
Updating the interwiki dispatcher list to include the multi-level domain for WikiOasis

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3316,6 +3316,13 @@ $wgConf->settings += [
 				'urlInt' => 'https://$2.wiki.gg/$3/wiki/$1',
 				'baseTransOnly' => true,
 			],
+			'wikioasis' => [
+				/** WikiOasis */
+				'interwiki' => 'wo',
+				'url' => 'https://$2.wikioasis.org/wiki/$1',
+				'urlInt' => 'https://$2.wikioasis.org/$3/wiki/$1',
+				'baseTransOnly' => true,
+			],
 		],
 		'+utgwiki' => [
 			'translate' => [
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -3318,9 +3318,8 @@ $wgConf->settings += [
 			],
 			'wikioasis' => [
 				/** WikiOasis */
-				'interwiki' => 'wo',
+				'interwiki' => 'woasis',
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -3316,6 +3316,13 @@
 				'urlInt' => 'https://$2.wiki.gg/$3/wiki/$1',
 				'baseTransOnly' => true,
 			],
+			'wikioasis' => [
+				/** WikiOasis */
+				'interwiki' => 'wo',
+				'url' => 'https://$2.wikioasis.org/wiki/$1',
+				'urlInt' => 'https://$2.wikioasis.org/$3/wiki/$1',
```
Universal-Omega: Does WO support this?

LocalSettings.php
```diff
@@ -3316,6 +3316,13 @@
 				'urlInt' => 'https://$2.wiki.gg/$3/wiki/$1',
 				'baseTransOnly' => true,
 			],
+			'wikioasis' => [
+				/** WikiOasis */
+				'interwiki' => 'wo',
```
Universal-Omega: The global interwiki doesn't seem to exist. IIRC, doesn't matter what it is is pointing to (like just meta.wikioasis.org) but it needs to exist.

LocalSettings.php
NotAracham: Removed urlInt as I do not have confirmation of this

LocalSettings.php
NotAracham: Added global interwiki and updated to woasis, as wo is a reserved interlanguage code

LocalSettings.php
Zippybonzo: Afaict we don't support it and tbh I am not sure what the benefits of it are lol

Universal-Omega (approved):
Not able to deploy right now but can in a bit if no one else does first.

LocalSettings.php
NotAracham: This is just another way to handle multi-language versions of pages, e.g. meta.wikioasis.org/fr/wiki/bread for the french language version of the page bread.  

100% just me moving too fast and making assumptions, thanks for the catch, CA.

PR #6517, opened 2026-08-30
add bilibili player to xyywiki csp

Honoka55:
Resolves [T15800](https://issue-tracker.miraheze.org/T15800).

Per

> In T14069#318220, Emiliers wrote:
> So does this mean any wikis who want to request per-wiki CSP allowance for bilibili can now make a separate Phorge post requesting it?

> In T14069#318221, NotAracham wrote:
> Correct


Universal-Omega:
Requested a review from @lihaohong6 to see if they want to merge this for now, or if #6520 first, if that is almost ready to go or not. If it isn't we could do this for now, and then move to that later.

lihaohong6:
@Universal-Omega Since adding bilibili as a service will allow users to enable it through ManageWiki unless we filter it out explicitly, I think it'll be fine to just list a bunch of wikis here that request bilibili on Phorge. The difference in workload is pretty small either way.

lihaohong6 (approved):

Universal-Omega:
Thanks for merging this then and sounds good to me.

PR #6519, opened 2026-08-30
MirahezeFunctions: Always write wiki URL to dblist

SomeMWDev:
The old code only does this if cw_wikis.wiki_url is not empty, which is only the case for 875 out of almost 30k wikis (presumably ones using custom domains).

This should improve the performance of ::getServers() a lot once the dblists are regenerated.

Part of T15909

### DIFF
--- a/initialise/MirahezeFunctions.php
+++ b/initialise/MirahezeFunctions.php
@@ -267,7 +267,7 @@ class MirahezeFunctions {
 				$servers[$db] = $data['u'];
 				continue;
 			}
-			$server = self::getServerForWiki( $realm, $db );
+			$server = self::getServerForWiki( $realm, $db, $data['d'] ?? null );
 			if ( $server !== null ) {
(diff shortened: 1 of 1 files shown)


initialise/MirahezeFunctions.php
```diff
@@ -968,6 +977,7 @@ private static function generateDatabaseLists( string $globalDatabase ): array {
 				'v' => $wikiVersion,
 			];
 
+			$url = $url ?: self::getServerForWiki( $realm, $db );
```
Universal-Omega: Can't this  use `$url ??= self::getServerForWiki( $realm, $db );` It should never be an empty string, if a custom domain is not set it would be null so that should work I think?

SomeMWDev: It can, I wrongly assumed it could be an empty string when writing this, but I confirmed that it returns null

Universal-Omega (approved):

PR #6520, opened 2026-08-31
T15205: move less frequently used domains in the CSP to ManageWiki

lihaohong6:
Backward-compatible patch will adds a ManageWiki setting to selectively add services to the CSP.

Each service corresponds to one or more CSP directives. It's either all or nothing. For finer granularity some services are split into two, such as `wikimedia-base` and `wikimedia-full`.

### DIFF
--- a/ContentSecurityPolicy.php
+++ b/ContentSecurityPolicy.php
@@ -5,12 +5,14 @@ $wgCSPHeader = [
 	'useNonces' => false,
 ];
 
-// Default CSP for all wikis
-$wgMirahezeMagicCSPHeaderDefault = [
+/**
+ * Domains that are essential to the operation of the farm and are therefore
+ * always present in the CSP of every wiki. Cannot be removed via ManageWiki.
+ */
+$wgMirahezeMagicCSPHeaderEssential = [
 	'default-src' => [
 		"'self'",
 		'*.miraheze.org',
-		'*.mirabeta.org',
 		'*.wikitide.org',
 		'*.wikitide.net',
 	],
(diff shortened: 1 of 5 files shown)


Universal-Omega:
This should not be merged without my approval first once I weigh some risks of it. I would prefer things be very explicit and listed in config rather than ManageWiki to avoid a potential security vulnerability and make it easier to fully ensure certain domains are removed if we remove them from the CSP. Overall the idea is good, but it carries risk I would like to consider more before this is merged.

UPDATE, just to clarify a bit, this is a great idea and design for it, I just want to ensure it is done properly and needs a proper security review first.

ContentSecurityPolicy.php
```diff
@@ -92,168 +46,754 @@
...
+];
+
+/**
+ * Services that an extension cannot work without.
+ */
+$wgMirahezeMagicCSPHeaderExtensionServices = [
+	'EmbedSpotify' => [
+		'spotify',
+	],
+	// We don't know which sites a wiki wants to embed videos from, so enable all of them?
```
RhinosF1, line 730: I would rather we enable the minimum when the extension is turned on and let people decide

ContentSecurityPolicy.php
```diff
@@ -92,168 +46,754 @@
...
+				'upload.wikimedia.org',
+			],
+			'connect-src' => [
+				'*.wikimedia.org',
+				'*.wikipedia.org',
+				'www.mediawiki.org',
+			],
+		],
+	],
+	'wikimedia-full' => [
```
RhinosF1, line 644: thing to test: enabling both wikimedia-base and wikimedia-full doesn't duplicate what's in the CSP

ContentSecurityPolicy.php
```diff
@@ -92,168 +46,754 @@
...
+		'google',
+	],
+	'Kartographer' => [
+		'openstreetmap',
+	],
+	'Maps' => [
+		'openstreetmap',
+		// Google Maps is blocked by the current CSP
+	],
+	// For image embeds. connect-src not required for the extension.
```
RhinosF1, line 757: this comment should be repeated where roblox is defined

ContentSecurityPolicy.php
```diff
@@ -92,168 +46,754 @@
...
+	'reddit' => [
+		'label' => 'Reddit (image embeds)',
+		'sources' => [
+			'img-src' => [
+				'*.redd.it',
+				'*.redditmedia.com',
+			],
+		],
+	],
+	'roblox' => [
```
RhinosF1, line 490: see other comment, this includes connect-src so should it be split?

ContentSecurityPolicy.php
```diff
@@ -5,12 +5,14 @@
 	'useNonces' => false,
 ];
 
-// Default CSP for all wikis
-$wgMirahezeMagicCSPHeaderDefault = [
+/**
+ * Domains that are essential to the operation of the farm and are therefore
+ * always present in the CSP of every wiki. Cannot be removed via ManageWiki.
+ */
+$wgMirahezeMagicCSPHeaderEssential = [
```
RhinosF1, line 12: lgtm

RhinosF1 (approved):
Generally happy and like the concept, see a few comments below though that could be addressed.

RhinosF1:
Assigned CA per his comment

ContentSecurityPolicy.php
lihaohong6: This is currently deduplicated through https://github.com/miraheze/MirahezeMagic/blob/5697d921503ad8044344190e7601789a97880072/includes/HookHandlers/ContentSecurityPolicy.php#L29 and once MH Magic is changed the new version would have a similar line.

ContentSecurityPolicy.php
lihaohong6: Figuring out the right granularity is hard since connect-src is more risky than img-src, but at the same time having too many services will add bloat to ManageWiki options. There was a split for Wikimedia wikis but I'm not sure if the same should be done for Roblox.

Maybe RobloxAPI should enable a hidden service not on ManageWiki that covers solely the image embedding part. It sort of loses the brevity advantage of mapping extensions to services (as opposed to spelling out all the directives again) but might be fine if it's done for only one or two extensions.

Universal-Omega (approved):
After considering this more, and thinking about potential ways it could be abused, I have decided to approve this concept as well.

PR #6525, opened 2026-09-07
Configuration changes for doorswiki per T15989

Aeywoo:
Adding custom cache expiries for the Official DOORS Wiki (doorswiki) per T15989.


RhinosF1:
@SomeMWDev should these just be set farm wide?

pskyechology (changes requested):
This could be in LocalSettings instead https://github.com/miraheze/mw-config/blob/1e7a9e2f4c80edd1e658eecc54655aa0c3dcaa67/LocalSettings.php#L6281-L6292

Aeywoo:
I've moved the changes to LocalSettings.php, is it done correctly?

LocalSettings.php
```diff
@@ -6289,6 +6289,13 @@
 			'userId' => 86400,
 			'userInfo' => 86400,
 		],
+		'+doorswiki' => [
+			'ext-RobloxAPI' => [
```
pskyechology, line 6293: Extension-based values (ext-RobloxAPI) aren't supported for wiki overrides. It's fine to not have it: I don't think you're going to disable RobloxAPI, and most configurations are present on all wikis despite not having their respective extensions installed.

pskyechology (changes requested):

SomeMWDev:
@Aeywoo the task description states:
> Hello, we're running into issues with the [RobloxAPI](https://www.mediawiki.org/wiki/Extension:RobloxAPI) extension because we're making a lot of API requests on one page ([Badges](https://doorsgame.wiki/wiki/Badges)) and we would like to be able to cache those results for a few minutes in-between API requests

What exactly are the issues you're running into? How are these issues resolved by decreasing the cache expiry for `gameData` and `groupRoles` from 30 minutes to 2 minutes?

Aeywoo:
> What exactly are the issues you're running into? How are these issues resolved by decreasing the cache expiry for `gameData` and `groupRoles` from 30 minutes to 2 minutes?

As far as I'm aware from reading `extension.json`, the `gameData` and `groupRoles` do not have any cache expiry values set, so they're refreshed constantly when the page is purge or edited. And since we are using a page that makes about 100+ calls to the API, it's better to not have these values refreshed every time the page is purged or edited.

SomeMWDev:
> > What exactly are the issues you're running into? How are these issues resolved by decreasing the cache expiry for `gameData` and `groupRoles` from 30 minutes to 2 minutes?
> 
> As far as I'm aware from reading `extension.json`, the `gameData` and `groupRoles` do not have any cache expiry values set, so they're refreshed constantly when the page is purge or edited. And since we are using a page that makes about 100+ calls to the API, it's better to not have these values refreshed every time the page is purged or edited.

Data sources that do not have an entry in the array use the value for `*`, which is `1800` (30 mins) by default on Miraheze, and will be `6000` (100 mins) after this PR. `120` is way too low.

Aeywoo:
Changes aren't needed anymore, realised it was an issue of using too many expensive parser tags.

PR #6535, opened 2026-09-15
test bilibili mobile csp on mirabeta

Honoka55:
Based on observations in [T14069#320004](https://issue-tracker.miraheze.org/T14069#320004)


Universal-Omega:
This does not seem needed anymore. I approved the CSP change already.

PR #6543, opened 2026-09-16
T15917: Add option to enable show categories in Minerva

Conscience93:
Resolve T15917 (https://issue-tracker.miraheze.org/T15917)


Universal-Omega:
I do not believe this will work like this, as the config expects an array but the check type will just do a true/false. Additionally, this one already seems to have per wiki overrides. Can just add necessary wikis in the LocalSettings.php declaration for now I think.

Conscience93:
Ah alright then

